PR #51, opened 2015-09-25
Added support for Monto

charleshofer:
Hey guys. Sorry this is so late in the week. I've added some jars, a few Java files, and a new Silver library for Monto. If you do `import lib:monto` in a Silver file, you can use the Monto features.


charleshofer:
Cleaned up support for Monto to do loops in Java rather than Silver
Added a tutorial for using Monto with Silver

PR #53, opened 2016-02-15
Work on list patterns

krame505:
I have added patterns of the form
case foo of
  [bar, bar1] -> baz
end
as an alternative to the current syntax
case foo of
  bar :: bar1 :: [] -> baz
end

This was done within the pattern matching extension, with the addition of a new production.  In addition, the current production for [] patterns was made abstract to avoid a conflict.

### DIFF
--- a/grammars/silver/extension/patternmatching/PatternTypes.sv
+++ b/grammars/silver/extension/patternmatching/PatternTypes.sv
@@ -165,15 +165,24 @@ concrete production listPattern
 top::Pattern ::= '[' ps::PatternList ']'
 {
   top.pp = s"[${ps.pp}]";
-  forwards to makeListPattern(ps.patternList, top.location);
+  forwards to ps.asListPattern;
 }
 
-function makeListPattern
-Pattern ::= ps::[Decorated Pattern] loc::Location
+synthesized attribute asListPattern::Pattern occurs on PatternList;
+
+aspect production patternList_one
+top::PatternList ::= p::Pattern
 {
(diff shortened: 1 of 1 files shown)


grammars/silver/extension/patternmatching/PatternTypes.sv
```diff
@@ -160,3 +160,20 @@ top::Pattern ::= hp::Pattern '::' tp::Pattern
...
 
+-- List literal patterns
+concrete production listPattern
+top::Pattern ::= '[' ps::PatternList ']'
+{
+  top.pp = s"[${ps.pp}]";
+  forwards to makeListPattern(ps.patternList, top.location);
+}
+
+function makeListPattern
```
tedinski: This looks like a good change, but I think I'd prefer this function to be a new attribute (on PatternList) instead. Especially since I think it'd be less wrong about the location annotation, right?

krame505: I can do that.

Lucas Kramer

krame505: Updated.

tedinski:
Looks good. Assuming you've tested it, go ahead and merge?

krame505:
OK, merged.  For simple things like this, do we want to delete the feature branch after merging?

tedinski:
Yeah, I can't think of a reason why we wouldn't. It'd just be a mess if we left them around.

PR #56, opened 2016-02-18
Added propagate for functor attributes, try 2

krame505:
Previous pull request had something weird happen.  

I have added a production statement of the form
propagate functor foo, bar;
which generates synthesized equations for foo and bar to copy the tree using the same attributes. For children on which they do not occur, the child itself is used in the new tree.


tedinski:
lgtm, as a first step.

ericvanwyk:
Ted - what does "Igtm" mean in your comment?

Do you want to pull this in or not?

tedinski:
Sorry, "looks good to me."

ericvanwyk:
Ok, I'm merging.

PR #58, opened 2016-02-26
Bringing flow analysis in line with thesis: part 1: phantom graphs

tedinski:
We need to enforce `ft(syn) >= ft(fwd)` for *extension* synthesized attribute occurrences.

Previously, we used a hack that didn't work perfectly. This is now removed.

This change means we now construct phantom production graphs for each nonterminal, which is exactly what is described in my thesis.


ericvanwyk:
Glad to see this completed.  Go ahead and merge.

ericvanwyk:
I mean, lgtm...

PR #59, opened 2016-02-28
More work on functor attributes

krame505:
Misc updates / fixes:
* Changed syntax as discussed
* Including the extension in the IDE spec
* Annotations are now copied from the current tree - before they were just missing
* Decorated nonterminals on the production RHS are now used directly, instead of copying by attribute.


krame505:
1. That doesn't appear to be the case!  Without stripping the full name, I get errors "full:path:anno is not appropriate for <whatever>" and "missing annotation anno for <whatever>

krame505:
2. I will fix that.

krame505:
*3. That appears to be the case, I can change it.

tedinski:
Re: 1. Huh. I'm thinking that's a bug then, though it's possible I'm forgetting something. I'll open an issue.

tedinski:
Oh, and looks good, feel free to merge.

PR #62, opened 2016-02-28
Do not strip carriage returns. Warn about forgetting them in terminals.

### LINKED ISSUE #24: Fix newline CRLF issues
```
Currently, Silver does a preprocess pass over its input to turn CRLF into just 
LF.  That is, we've basically chosen to work around stupid incompatibilities in 
text file formats by mandating the unix style.  I think something like this is 
important, because otherwise a perfectly functioning parser on unix will break 
on windows. (Worse: will break on unix, given an input from windows.)

However, a preprocessor pass before parsing might be somewhat silly.  A 
possible alternative is this: change the interpretation of '\n' in regular 
expressions to '(\n|\r\n)', thus allowing for either format whenever a 
"newline" appears in a regex.

I'm not sure this is the best way to handle this yet, though.  Perhaps we could 
leave \n and \r alone, and introduce '[:newline:]'. Or something. What do other 
parsers do?
```

Original issue reported on code.google.com by `tux...@gmail.com` on 9 Aug 2012 at 4:59

### LINKED ISSUE #49: Carriage returns cause issues with error highlighting
If there are carriage returns in a file, the index of where the error occurs gets messed up, and the underline gets shifted.

tedinski:
Fixes #24 and also fixes #49 

We discussed making this change long ago, as removing carriage returns now causes more problems than it solves (see #49.)

In order to help the transition along, I added a nice little hack of a warning to discover potentially broken terminal regexes. Then I fixed a few it discovered in Silver. Hah!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@ JARS-BAK
 .\#*
 
 # Runtime and testing junk
-test/**/test.output
+test/**/*.test.output
 test/**/test.jar
 test/**/build.xml
 test/silver.testing.bin.jar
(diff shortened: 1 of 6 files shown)


ericvanwyk:
Ted, can you say what the hack is?  Are we doing what is suggested in #24 (that is interpret \n as (\n|\r\n) ?  Thanks.

tedinski:
The warning just spots regular expressions that contain \n but not \r.

I call it a hack just because that is a pretty crude condition, but it appears to work perfectly so far!

PR #66, opened 2016-04-07
Adding lambda expressions

krame505:
This modification adds lambda expressions, of the form
```lambda x::Integer y::String -> toString(x) ++ y```
There are still several items which I am not sure are correct (marked with TODOs):
* Lazy translation
* Lambda param value names - need to use genInt()?
* Type checking
* Flow analysis
* Translation types
* Syntax
* More test cases...

### CODE REVIEW

File: grammars/silver/definition/core/AGDcl.sv [hunk 1]
```diff
@@ -42,6 +42,14 @@ top::AGDcl ::=
   top.errors := [];
 }
 
+abstract production errorEAGDcl
```
tedinski: Why EAGLDcl? Typo? Should just be errorAGDcl, no?

grammars/silver/definition/core/ProductionBody.sv
```diff
@@ -105,6 +105,17 @@ top::ProductionStmt ::= h::ProductionStmt t::ProductionStmt
   top.errors := h.errors ++ t.errors;
 }
 
+aspect production errorProductionStmt
```
tedinski: aspect and abstract are backwards from here to the typechecking module. This one should be abstract.

grammars/silver/modification/lambda_fn/Lambda.sv
```diff
@@ -0,0 +1,71 @@
+import silver:definition:flow:ast only ExprVertexInfo, FlowVertex;
+
+--- Concrete Syntax for lets
+--------------------------------------------------------------------------------
+
+terminal Lambda_kwd 'lambda' lexer classes {KEYWORD,RESERVED};
```
tedinski: Q: Do we want to stick with haskell style syntax? That is, '\' instead of 'lambda'? I don't know, but I thought I'd ask if this was deliberate choice.

grammars/silver/modification/lambda_fn/Lambda.sv
```diff
@@ -0,0 +1,71 @@
...
+terminal Arrow_t '->' precedence = 0, lexer classes {SPECOP};
+
+concrete production lambda_c
+top::Expr ::= 'lambda' params::ProductionRHS '->' e::Expr
+{
+  top.pp = "lambda " ++ params.pp ++ " -> " ++ e.pp;
+
+  forwards to lambdap(params, e, location=top.location);
```
tedinski: I don't recall why the 'let' grammar uses a concrete/abstract divide, but I'm not sure if we're getting any advantage from doing so here. Is merging these together okay?

Despite concrete/abstract probably being a good thing, the Silver compiler is generally built on just doing things on concrete trees. I'd stick to doing it that way, for consistency.

grammars/silver/modification/lambda_fn/Lambda.sv
krame505: I don't know.  I just picked that for now, it is easy to change.

Lucas Kramer

grammars/silver/modification/lambda_fn/Lambda.sv
```diff
@@ -0,0 +1,71 @@
...
+abstract production lambdap
+top::Expr ::= params::ProductionRHS e::Expr
+{
+  top.pp = "lambda " ++ params.pp ++ " -> " ++ e.pp;
+  
+  top.errors := params.errors ++ e.errors;
+  
+  top.typerep = functionTypeExp(e.typerep, map((.typerep), params.inputElements), []);
+  
+  -- TODO?
```
tedinski: Both these TODO questions look good to me.

grammars/silver/modification/lambda_fn/Lambda.sv
krame505: I think part of the reasoning is that these are both things that could find
a bit of common use with other extensions, so it would be a bit simpler to
not need to pass the terminal symbols explicitly.

Lucas Kramer

grammars/silver/modification/lambda_fn/Lambda.sv
krame505: OK I will remove them.

Lucas Kramer

grammars/silver/modification/lambda_fn/Lambda.sv
```diff
@@ -0,0 +1,71 @@
...
+concrete production lambda_c
+top::Expr ::= 'lambda' params::ProductionRHS '->' e::Expr
+{
+  top.pp = "lambda " ++ params.pp ++ " -> " ++ e.pp;
+
+  forwards to lambdap(params, e, location=top.location);
+}
+
+abstract production lambdap
+top::Expr ::= params::ProductionRHS e::Expr
```
tedinski: Oooh, hmm. I actually don't know if I'm okay with re-using ProductionRHS here.

...okay, I went and looked through things a bit, and it seems like a perfectly ok choice for now. However, I'd add maybe a comment pointing it out and justifying it, because I thought it was unexpected!

grammars/silver/modification/lambda_fn/java/Lambda.sv
```diff
@@ -0,0 +1,72 @@
...
+  local finTy :: TypeExp = finalType(top);
+  
+  top.translation = 
+s"""(new common.NodeFactory<${e.typerep.transType}>() {
+  public final ${e.typerep.transType} invoke(final Object[] args, final Object[] namedArgs) {
+    ${params.lambdaTranslation}
+    return ${e.translation};
+  }
+})""";
+  top.lazyTranslation = top.translation; -- TODO: is this correct?
```
tedinski: Good question. After looking at things a bit, I'm pretty sure, yes. Values seem to skip wrapping in lazy.

krame505:
Hmm, I tried to clean up the commit history, but it is still kind of a mess.  All the files are correct as-is, though.

tedinski:
This is looking good to me. Eric?

I would like to place a vote for '\' instead of 'lambda' though. I think the examples in the test cases look kinda... funny. However, I am used to Haskell here.

ericvanwyk:
Ted - I'll defer to you on the technical things and it seems you're happy
at this point.

Regarding "lambda" vs "\".  Silver's syntax is heavily influenced by
Haskell, though some parts are different.  How about we take some
inspirations from OCaml and use "fun" instead?

Just kidding!  I'd vote for "\" as well.

-Eric

PR #67, opened 2016-04-10
Correct handling of lazy value parameters

krame505:
This fixes a bug with Thunks passes in as objects being casted to their final types directly, instead of being unpacked if needed.


tedinski:
This looks good. Only request: add a comment explaining why we do 'final Object' instead of transType. After all, it tripped us up once. :)

krame505:
Done.

PR #68, opened 2016-04-10
Feature/lambda

tedinski:
Fixed the bug related to thunk 'context' being captured via this instead of directly (since it gets mutated to null inside Thunk.)


krame505:
Hi Ted,
Thanks for working on this!  I am getting a large number of compilation errors when compiling the test cases, even after rebuilding with --clean.  Are you sure you committed everything?

tedinski:
self-compile? That won't work for changes like this.

Start with fresh jars (re fetch-jars) and deep-rebuild.

krame505:
Ah, yes, I forgot I had to deep-rebuild when the runtime changes.

krame505:
OK, looks good to me!

krame505:
Oops, meant to merge not close

PR #69, opened 2016-04-12
Adding monads

krame505:
This extension adds monads to Silver.  Since we don't have typeclasses (yet), the do expression requires the bind and return functions to be provided explicitly.  The do expression allows for sequences of plain monadic expressions, bind statements, let assignment statements, conditional statements, and return.  These are fairly straightforward, except for conditional statements.  These can have a single statement for a body, or a bracketed sequence of statements.  The else clause is optional, if there is no else provided then it defaults as ```return unit();```.  

There are provided bind and return functions for State, IO, lists, and Maybe.  

For example:
```
local result::IOMonad<Integer> = do (bindIO, returnIO) {
  printM("Hello, world!\n");
  printM("Choose a number: ");
  txt::String <- readLineStdinM;
  num::Integer = toInt(txt);
  if num == 42 then {
    printM("Yes");
    return num;
  }
  else
    return num * 10;
};
```

Issues to resolve/consider:
* It appears that we can't write functions that ignore their parameter type, due to the way Silver handles type variables, which forces all expression do body statements to have their monadic base type be Unit.  How do we fix this?  
* Should actions such as readLineStdinM that take no arguments just be globals?  
* Unit type; is this done right?
* core:monad isn't being imported automatically, is this an issue?
* if/then/else syntax
* More tests are needed...


grammars/core/Unit.sv
```diff
@@ -0,0 +1,7 @@
+grammar core;
+
+nonterminal UnitT;
```
tedinski: Why is this UnitT and not Unit?

grammars/core/monad/IO.sv
```diff
@@ -0,0 +1,220 @@
+grammar core:monad;
+
+type IOMonad<a> = State<IO a>;
+
+abstract production bindIO
```
tedinski: I think both of these should be functions, not productions.

grammars/core/monad/State.sv
```diff
@@ -0,0 +1,50 @@
...
+
+abstract production returnState
+top::State<s a> ::= x::a
+{
+  top.stateOut = top.stateIn;
+  top.stateVal = x;
+}
+
+abstract production getState
+top::State<s s> ::=
```
tedinski: Ahaha! Awesome. Casual use of GADTs. :D

grammars/silver/extension/monad/ConcreteSyntax.sv
```diff
@@ -0,0 +1,126 @@
+
+concrete production do_c
+top::Expr ::= 'do' '(' bindFn::QName ',' returnFn::QName ')' '{' body::DoBodyStmts '}'
+{
+  top.pp = s"do (${bindFn.pp}, ${returnFn.pp}) {${body.pp}}";
+
+  -- TODO: quit reporting errors about undefined bind and lookup functions repeatedly
```
tedinski: This is probably a good place to use errorExpr that you recently added.

In addition to looking up the names, you could check their types, too, before deciding to forward to `body.transform`.

But this is a good first draft, of course.

grammars/silver/extension/monad/ConcreteSyntax.sv
```diff
@@ -0,0 +1,126 @@
...
+  top.errors <- bindFn.lookupValue.errors ++ returnFn.lookupValue.errors;
+  --top.errors <- body.errors; 
+  
+  body.bindFn = bindFn;
+  body.returnFn = returnFn;
+  
+  forwards to body.transform;
+}
+
+terminal DoId_t /[a-zA-Z][A-Za-z0-9\_]*/ lexer classes {IDENTIFIER};
```
tedinski: Unused?

grammars/silver/extension/monad/Name.sv
```diff
@@ -0,0 +1,34 @@
+-- Exact copy of Name, but productions have higher precedence.
+-- This is needed to avoid a parse conflict with '::' as both the type and cons operator
```
tedinski: So, this should be accomplishable with an operator precedence on '<-' without MName at all. I think.

In fact, I'm not sure how this solves the problem, except by forbidding `x :: ...` from starting an expression when we're inside a do block. I think this is a wrong approach to disambiguation here.

test/silver_features/Monad.sv
```diff
@@ -0,0 +1,11 @@
+import core:monad;
```
tedinski: Definitely need more here. :)

At the very least, for each kind of monad you put in the stdlib. This is just lists.

BTW, you can write tests on lists.

`equalityTest(blah, [1,2,3], [Integer], silver_tests);`

Might be better than folding them here.

grammars/core/monad/IO.sv
krame505: Eric, what are your thoughts on bindIO and returnIO being functions instead
of productions?

Lucas Kramer

krame505:
After discussing with Eric, actions such as readLineStdinM will remain as no-argument functions, although they could be global values.  This is to improve consistency with other actions that must be productions such as get(), and also since Silver functions aren't curried, the correspondence with Haskell is weaker anyways.

grammars/core/monad/State.sv
```diff
@@ -0,0 +1,50 @@
...
+synthesized attribute stateOut<s>::s;
+synthesized attribute stateVal<a>::a;
+
+abstract production bindState
+top::State<s b> ::= st::State<s a> fn::(State<s b> ::= a)
+{
+  st.stateIn = top.stateIn;
+  forward.stateIn = st.stateOut;
+  
+  forwards to fn(st.stateVal);
```
tedinski: Although the use of "forwards to" here saves typing, I think this qualifies as an abuse of forwarding. I think you should explicitly have a local that you thread into and out of (and copy val out of.) Actually, unless I'm mistaken, it only adds one line.

tedinski:
Another thought I had: I think it might be worthwhile to have a specialized type for IO.

That main thing is that we could write `top.stateVal = unsafeTrace(the_value, top.stateOut);` inside bind. I think this might actually be a solution where we successfully ensure IO has occurred, even when evaluation is being driven by the value and not the IO token.

krame505:
Hmm... is that really an issue if we are driving the IO by demanding a
value?  Using unsafeTrace looks a bit weird if someone is reading the
code.  Eric, do you have an opinion on this?

Lucas Kramer

krame505:
OK, so after discussing with Eric, I guess using unsafeTrace could actually an effect, as it would give better control of the ordering with respect to long computations.  For now I am doing this within the bindIO forwarding production.  Having a separate nonterminal for IO would complicate things, as that would require having 'getIO' and 'setIO' productions exactly the same as State, or else make every IO action a production, which would be less than optimal.  
Ted, do you have any opposition to this?  Otherwise we can talk more on this Thurs, but I really do want to get this finalized and merged by the end of the week.

krame505:
OK, does this look good to merge?

krame505:
Merging.

PR #70, opened 2016-04-14
Feature/docgen

charleshofer:
Merging feature branch so that we can write documentation in develop, and simply checkout develop rather than feature/docgen in the website generation scripts.


krame505:
Any reason not to merge this now?

ericvanwyk:
Not that I can think of.  Go ahead if things are working.

-Eric

PR #73, opened 2016-05-11
add vim highlighting support

TravisCarlson:
keywords, etc. copied from gedit and google-code-prettyify

### DIFF
--- a/support/vim/install.sh
+++ b/support/vim/install.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
-mkdir -p ~/.vim/syntax/ ~/.vim/ftdetect \
-	&& cp -f sv.vim ~/.vim/syntax/ \
-	&& echo "au BufRead,BufNewFile *.sv setfiletype sv" > ~/.vim/ftdetect/sv.vim
+
+# stop on error
+set -eu
+
+mkdir -p ~/.vim/syntax/ ~/.vim/ftdetect
+cp -f sv.vim ~/.vim/syntax/
+echo "au BufRead,BufNewFile *.sv setfiletype sv" > ~/.vim/ftdetect/sv.vim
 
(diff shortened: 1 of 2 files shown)


support/vim/install.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
tedinski: Instead of chaining things up with &&, start the script off with 'set -eu'.  e will stop the script on any error. u will error on undefined variables (not helpful here, but this is general advice for all scripts.)

then everything else can just be ordinary lines

support/vim/sv.vim
```diff
@@ -0,0 +1,56 @@
...
+
+syn keyword flow case of let in end decorate with prefix else forward if new then toString toInt toFloat length
+
+syn keyword ide temp_imp_ide_dcl temp_imp_ide_font font color italic bold
+
+syn keyword kw builder postbuilder exporter folder property string required string wizard new file stub generator product name version option resource
+
+syn keyword boolean false true
+
+syn match comment "\(--.*\)\|\({-\([^-]\|-\+[^-}]\)*-}\)$"
```
tedinski: I think Proper Vim Style is to kick off all regexes with \v to avoid problems with different magic modes.

tedinski:
Also, I had these in mine, which might be helpful or adaptable:

{{{
syntax match svlangComment "\v--.*$" contains=svlangCommentKeyword
syntax region svlangComment start=/\v\{\-/ end=/\v\-\}/ fold contains=svlangCommentKeyword
highlight link svlangComment Comment

syntax sync fromstart

set foldmethod=syntax

set autoindent
set expandtab
set shiftwidth=2
set softtabstop=2

}}}

As I mentioned, I hadn't gotten this to 100% work to my liking before I just stopped using vim that much. But it might be helpful?

TravisCarlson:
Thanks for the suggestions. They should all be addressed now. Let me know if I missed anything.

Regarding the `if exists("b:current_syntax")` being at the top of the file, I checked /usr/share/vim/vim73/syntax/c.vim and it does the same.

tedinski:
aha, found what was bugging me

the c example ends with
```
let b:current_syntax = "c"
```

so it's like include guards. we should probably have that with "sv" or "svlang" or something at the bottom of the file.

After that, looks good to me to merge.

PR #74, opened 2016-05-18
Feature/docgen

ericvanwyk:
It has been a while since docgen was merged into develop.


ericvanwyk:
I'm going ahead and doing the merge into develop as nothing seems to conflict.

PR #75, opened 2016-05-19
Use java 7 APIs

tedinski:
This changes:

* system - pass through stdin/out/err
* copyFile - use simpler routine to do the copy
* readFile - use nice helper to fully read a file
* deleteTree - correctly detect and handle symlinks
* getStr - don't dump lines when stdin if overfull


TravisCarlson:
Thanks, Ted. I've confirmed this solves the problem discussed on the ableC pull request of system() not printing its command's stderr.

https://github.com/melt-umn/ableC/pull/13

PR #77, opened 2016-06-09
Feature/transparentprefixes

krame505:
Adding support for copper's transparent prefixes.  This replaces the slightly hacking solution from eff4ecfa886e1606b3d79bd9ac58d525d95b49c9 which simply renamed existing silver terminals, excluding the option of writing a marking terminal without a prefix.  

The only major item that could be improved is to auto-generate the needed disambiguation functions instead of needing to write them for every conflicting group.  This would require figuring out which terminals are conflicting in the first place, which would be nice to get from Copper, but that is currently ruled out by the way the build process is structured.  We are thinking of modifying this, so maybe we will return to this concept someday.


ericvanwyk:
This looks good to me.  Ted - any reason not to merge?

-Eric

grammars/silver/modification/copper/ParserDcl.sv
```diff
@@ -1,24 +1,45 @@
...
-import silver:extension:easyterminal; -- only Terminal_t, EasyTerminalRef;
+import silver:driver:util only computeDependencies;
 
 terminal Parser_kwd 'parser' lexer classes {KEYWORD}; -- not RESERVED?
 
 -- TODO: You know, maybe parser specs should get moved over here as well.
 
+-- TODO: parserDcl now just gets the AGDcls needed to be declared for prefixes then forwards to
+-- decls using that list with parserDclBase handling what parserDcl did
+-- This means we need to include the current grammar by default
```
tedinski: This comment is now obsolete, right?

krame505: Uh, yes.  I'll pull that out...

Lucas Kramer

tedinski:
It looks good other than I don't know what's going on with prefix_separator?

A weird, ad-hoc, special DclInfo seems like a bad solution, but maybe that's because I just don't know what's going on.

krame505:
Eric wanted it to be something special that gets declared once in the host
language, then gets looked up when defining prefixes in the parser spec.
It needs to go in the interface files because of this, so I don't know how
else to do it.

Lucas Kramer

grammars/silver/modification/copper/env_parser/Parser.sv
```diff
@@ -30,4 +30,10 @@ top::IDclInfo ::= 'parse_attr' '(' l::ILocation ',' fn::IName ',' t::ITypeRep ')
   top.defs = [parserAttrDef(top.grammarName, l.alocation, fn.aname, t.typerep)];
 }
 
+concrete production aDclInfoPrefixSeparator
+top::IDclInfo ::= 'parse_attr' '(' l::ILocation ',' s::IString ')'
```
tedinski: Quick bug: 'parse_attr'

grammars/silver/modification/copper/Prefix.sv
```diff
@@ -0,0 +1,192 @@
...
+  top.terminalPrefixes = [];
+  top.liftedAGDcls = prefixSeparatorAGDcl($1, $2, $3, $4, location=top.location);
+}
+
+concrete production prefixSeparatorAGDcl
+top::AGDcl ::= 'prefix' 'separator' s::String_t ';'
+{
+  top.pp = s"prefix separator ${s.lexeme};";
+  top.errors := 
+    case getValueDcl("_prefix_seperator", top.env) of
```
tedinski: should ask if it's null instead. This code will be fine with 3 declarations! :)

grammars/silver/modification/copper/env_parser/Parser.sv
krame505: Huh, not sure how that is acting like it is working.

Lucas Kramer

tedinski:
Okay, annoying question then: can we do this in the CstAst instead? So it doesn't touch the Silver environment?

The only problem I can forsee if I understand what's going on is that we'd need to mark terminal declarations as being prefixes so they know to add the prefix to their regex, but that seems okay to me.

krame505:
I guess we could, but that seems like a lot of extra work?  I don't really
see an issue with what we are doing now...

Lucas Kramer

grammars/silver/modification/copper/Prefix.sv
krame505: You mean line 187, right?  Since all defs in silver are recursive, when that is the only separator defined, it still shows up in the env, so one occurrence is expected there.

krame505:
Just realized this still isn't merged...
It works as is, any issue with just merging this and then moving the separator stuff to CstAst if we decide to do so later on?

grammars/silver/modification/copper/DclInfo.sv
```diff
@@ -111,3 +111,14 @@ top::DclInfo ::= sg::String sl::Location fn::String ty::TypeExp
   top.defLHSDispatcher = parserAttributeDefLHS(_, location=_);
 }
 
+abstract production prefixSeparatorDcl
```
tedinski: Okay. Let's add a comment here:

```
-- TODO: This sort of thing probably ought to be done on the CstAst and not be a part of the Silver environment pretending to be a value.
```

And then merge it.

krame505:
OK, merging if there aren't any other objections.

tedinski:
Hi Lucas,

I think these changes caused a new error in the `test/copper_features` test in Silver. Could you take a look at it and change the test appropriately? I think the test is now wrong, but I don't really know what you guys decided to change the behavior to.

The test failing right now looks like 

```
parser shouldSucceed :: Root {
  copper_features:mdatests:host;
  copper_features:mdatests:ext prefix 'c' with '1:';
  copper_features:mdatests:ext2 prefix 'c' with '2:';
}
```

and gives:

```
Errors for : copper_features:mdatests :
Test.sv:17:38: error: Found ambiguous possibilities for 'c'
	copper_features:mdatests:ext2:C (Ext.sv:7)
	copper_features:mdatests:ext:C (Ext.sv:5)
Test.sv:18:39: error: Found ambiguous possibilities for 'c'
	copper_features:mdatests:ext2:C (Ext.sv:7)
	copper_features:mdatests:ext:C (Ext.sv:5)
```

and there are some equality tests following this that probably don't match the current behavior.

krame505:
OK. This stems from the way we handle transparent prefixes now requiring a
disambiguation function for the case when no prefix is provided.

Hi Lucas,

I think these changes caused a new error in the test/copper_features test
in Silver. Could you take a look at it and change the test appropriately? I
think the test is now wrong, but I don't really know what you guys decided
to change the behavior to.

The test failing right now looks like

parser shouldSucceed :: Root {
  copper_features:mdatests:host;
  copper_features:mdatests:ext prefix 'c' with '1:';
  copper_features:mdatests:ext2 prefix 'c' with '2:';
}

and gives:

Errors for : copper_features:mdatests :
Test.sv:17:38: error: Found ambiguous possibilities for 'c'
    copper_features:mdatests:ext2:C (Ext.sv:7)
    copper_features:mdatests:ext:C (Ext.sv:5)
Test.sv:18:39: error: Found ambiguous possibilities for 'c'
    copper_features:mdatests:ext2:C (Ext.sv:7)
    copper_features:mdatests:ext:C (Ext.sv:5)

and there are some equality tests following this that probably don't match
the current behavior.

—
You are receiving this because you modified the open/close state.
<https://github.com/melt-umn/silver/pull/77#issuecomment-232791974>, or mute
the thread
<https://github.com/notifications/unsubscribe/AIE1ilBvHHuOWrt6QWkCjzimlDu-J1EWks5qVqRQgaJpZM4IyKcQ>
.

krame505:
Ah, the error you were seeing is actually because of the change in how we generate the env inside parser specs.  What I mentioned also needed to be fixed.  Anyways, fixed in #87.  

Also, do we potentially want to change the handling of the env inside the parser spec to add a new scope for the env seen by each item containing its defs?  This would make ambiguous references like the errors you mentioned go away - and not needing to write the full name basically every time (since you are usually using prefixes when two are the same) would be nice.

tedinski:
That'd probably be nice.

Anyway, #87 looks good to me.

ericvanwyk:
This is an issue, #85, labeled as an 'enhancement'.

PR #82, opened 2016-06-30
Improving scripts

tedinski:
I want to make more fixes, but I did these awhile ago, but apparently haven't merged them yet.

These change improve style: use of `set -eu` and `printf`.
They also fix some errors in a few scripts.
Finally, I realized we only need `readlink` not `readlink -f` so this should fix annoying problems with OS X.

Lastly, the `update` script creates symlinks to scratch space for you on CS machines, if you don't have them already. So that'd be nice to have now that coldpress was reinstalled.


ericvanwyk:
Have you tried running the silver script on coldpress now that it is at 16.04?  It seem to sill need the -f flag on readlink.

Also, with the set -eu declaration at the top of deep-rebuild, I get an error (again on coldpress) if I don't specify at least one argument.  The -e causes an error since $1 is not defined.  I wrapped the 'new core' if-then with 
  if [ $# -gt 0 ]; then
and then I don't get the error and deep-rebuild runs to completion as expected.

Thoughts?  I can fix the deep-rebuild one but I'm not sure about the silver script.

tedinski:
I'll fix the `deep-rebuild` issue asap.

However... there's no reason for silver should need the `-f` on readlink. Mine doesn't. What's going on with yours?

```
% ls -l ~/bin/silver
lrwxrwxrwx 1 tedinski grad 55 Jul 12 18:15 /home/grad04/tedinski/bin/silver -> /project/melt4/People/ted/git/silver/support/bin/silver*
```

Do you get something very different? Did you use the `install-silver-bin` script to put it there?

ericvanwyk:
I have the silver/support/bin directory in my $PATH.  I didn't run the
install-silver-bin script.

Without the -f, this line in support/bin/silver:
   ME=$(readlink "${BASH_SOURCE}")
is the empty string, thus the "! -f "$SILVER" test fails and I'm told the
jars couldn't be found.

tedinski:
Is there some reason you did it that way? (Other than ages ago I may not have written the install script yet?)

ericvanwyk:
No reason - probably historical as you suggest.  But it seems a reasonable
thing to do, no?

tedinski:
Well, it breaks the assumption the silver script is now making.

I'm trying to eliminate the need for `-f` since that's what complicates silver scripts running on macs, with the need for `greadlink` and such.

ericvanwyk:
Makes sense to me.  I've created the link in my ~/bin directory.

But the install-silver-bin script still has the Mac OS X distinction.  Will
this be simplified as well?

tedinski:
Oh yeah. I haven't finished working on this. Basically started it before the defense and now haven't returned to it.

Issue #8, opened 2015-08-20
IOToken


GoogleCodeExporter:
```
Provide an explanation of what the project should entail:

Right now the "IO token" passed around in order to "thread the state of the 
world" and make IO possible in Silver is just 'null'. Of type 'Object'.

The runtime already has a type called IOToken, and I believe it has a 
'singleton' public static final field.

Change the type to IOToken instead of Object, and pass that object around 
instead of null.

Provide a concrete example of a problem this project will fix:

Technically there is a huge bug with buggy code. Silver conflates 'null' with 
"has not been evaluated yet", so if buggy code demands an IO token by two 
different paths, it could cause an IO action to be repeated. Of course code 
shouldn't do that but, well..

Also, this is mostly a type-safety & code cleanliness sort of thing.

Provide a short guide on how to start this project:

silver:defintion:type:io
silver:translation:java:type:io
Check out IOToken in runtime/java/src/common/

Provide a rough expectation of what the set of patches completing this
project might look like:

One lump that changes over.

Probable challenges in completing this project:

Well, it involves working with the runtime a little bit, although that work is 
already sort of done.

Additional Information:

Making it work might be tricky because you have to spot all the places where 
null is used for the IOToken. Here:

grep -R "null[^\(]" grammars/

```

Original issue reported on code.google.com by `tux...@gmail.com` on 10 Jan 2012 at 1:25

TravisCarlson:
I've started looking at this and could use some guidance on whether I'm on the right track or not. Lucas and I plan to talk in person tomorrow. I'm writing this mostly to help facilitate that discussion.

> Change the type to IOToken instead of Object, and pass that object around 
instead of null.

I guess this involves making lots of s/Object/IOToken/ and s/null/IOToken.singleton/ edits in runtime/java/src/common/Util.java such as:

```diff
-       public static Object print(String sb) {
+       public static IOToken print(String sb) {
                // TODO: should we avoid demanding stringcatter objects?
                System.out.print(sb);
-               return null;
+               return IOToken.singleton;
        }
```

The only function given me compilation problems is io().
```diff
-       public static Object io(Object i, Object o) {
+       public static IOToken io(Object i, IOToken o) {
                return null;
+             return IOToken.singleton;
        }
```

Replacing null with IOToken.singleton results in the following NullPointerException.

```
One more full cycle ...
Compiling Grammar: silver:composed:Default
	[/project/melt4/People/travis/silver/grammars/silver/composed/Default/Main.sv]

[...]

Issue #78, opened 2016-06-09
IO monad repeats actions


krame505:
I'm not sure how nobody managed to notice this yet, but the IO monad appears to be repeating actions.  For example, calling runIO on the following
```
do (bindIO, returnIO) {
  printM("A\n");
  printM("B\n");
};
```
causes
```
A
A
B
```
to be printed.  Removing the ```unsafeTrace``` call for the equation for stateOut in bindIO prevents this from occuring.

tedinski:
Ohohohoho!

This is issue #8 popping up and roaring.

tedinski:
Fixed with #91.

krame505:
Glad to see this is finally fixed!  Has anyone checked that the monad
driver for ableC now runs correctly (without printing errors, etc. more
than once?)

Lucas Kramer

TravisCarlson:
Hi Lucas,

Yes, I've been verifying this against ableC and it no longer prints in duplicate. Though if you could also verify and let me know if you still notice any problem, that would be great.

Thanks,
Travis

tedinski:
Ah, good call. I closed it assuming this would fix it. We should ofc verify.

PR #91, opened 2016-08-18
Fix/iotoken

TravisCarlson:
Fix issues #8 and #78.

This will not build cleanly. See [this comment](https://github.com/melt-umn/silver/issues/8#issuecomment-240524558).


tedinski:
okay, I just did a publish-jars, so let's merge.

tedinski:
Looks like this broke the gitbot script, so I've pushed a quick fix for that as well. Seems to be fine now.

Issue #81, opened 2016-06-23
Restructuring IDE features


tedinski:
I wanted to write down these ideas, so they're available to look at in a single location.

This is the general plan for improving the IDE stuff:
1. Refactor things so that there's no reason why the silver compiler needs to _generate_ anything eclipse-specific.
2. Re-arrange everything so that it doesn't.

My memory is hazy enough that I can't give a full guide to what involved to accomplish this, but I think step 1 is obvious: we should eliminate the duplicate, parallel parsers generated for the IDE, and instead move all the necessary information directly onto the normally generated parsers.

The main thing to look at is `grammars/silver/modification/impide/cstast/` as well as the code that uses it, one level up from that. We have a whole thing here dedicated to creating a modified parsers that does IDE-specific stuff, mostly so that we can then extract a list of tokens to do highlighting.

So let's modify the normal parsers to also build a list of tokens. Problem number 1: we use `TerminalRecord` to store lexeme and location and... that's it. We don't store the terminal type. No associated list of lexer classes.

After thinking about it, I think the simplest, most future-proof method of fixing this to change Silver to have every terminal be a subclass of `TerminalRecord` (in fact, maybe we just rename this `Terminal` while we're at it.) And then add a few methods onto it to get name and an array of lexer classes, for the benefit of IDE code using it.

tedinski:
As a concrete example, here's what I imagine a generated terminal class might look like:

```
class TFoo extends common.Terminal {
...
@Override
String getName() { return "some:grammar:Foo"; }

private static String[] lexerclasses = null;
@Override
String[] getLexerClasses() {
  // Avoid doing more work at class-load time, in case we don't need this
  if(lexerclasses == null) {
    lexerclasses = new String[] {"some:class", "z:another:class"};
  }
  return lexerclasses;
}
}
```

so on Terminal(Record) we'd add:

```
/**
 * Obtains the type of this terminal.
 * @returns The full name of this terminal type (e.g. "silver:definition:core:Id_t")
 */
public abstract String getName();

/**
 * Obtains the lexer classes this terminal belongs to.
 * @returns The classes, as an array of interned strings in sorted order.
 */
public abstract String[] getLexerClasses();
```

(String literals are always interned, and this allows us to compare two interned strings with == instead of .equals. This will be useful for implementing the coloring code.)

TravisCarlson:
[...]

PR #93, opened 2016-09-01
make parts of the driver more reusable

tedinski:
Hi @TravisCarlson 

I need to get back to doing other things. I got into a fight with the driver, and I think I've managed to come away with an improvement that should allow the IDE driver bits to more easily re-use what we have in the normal driver here.

I haven't actually re-written the IDE bits though. I'm pretty sure I didn't break them though, so I've got that going for me.

With this change we should be able to call `cmdLineRunInitial` (uhhh... so that was sort of a temporary name I never fixed. Maybe rename that to something better. I dunno what though! OH! And when we're calling it from Java code, there should be a comment on the function saying that. So we later don't think we can freely rename it and assume everything's fine.) and then implement the rest on the Java side of things (to implement the Java equivalent of `ideAnalyze` re: #81)

(Also, not sure if it's clear, but I think we should be able to do a better job of implementing the equivalent of `ideGenerate` using state. That is, save the `Decorated Compilation` from the previous analyze and then use it to call the postOps somehow. This might be obfuscated by the present design of these hooks in the silver-eclipse runtime. So maybe I'll have to look at that again. But right now "analyze" does a whole build to look for errors, and then "generate" does a full build again assuming no errors to actually do generation. This is pointlessly inefficient, it's just that we never did anything with state passing to fix it.)

If you think this looks sane, go ahead and merge it (and delete branch.) Thanks! (N.B. I'm requesting this into develop, not your branch so you'd want to merge it there too ofc.)

### DIFF
--- a/grammars/silver/driver/BuildProcess.sv
+++ b/grammars/silver/driver/BuildProcess.sv
@@ -9,6 +9,8 @@ imports silver:util:cmdargs;
 
 exports silver:driver:util;
 
+import silver:langutil only message;
+
 type SVParser = (ParseResult<Root> ::= String String);
 type SVIParser = (ParseResult<IRoot> ::= String String);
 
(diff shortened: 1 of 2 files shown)


TravisCarlson:
I think it looks good. My first refactoring of `ideAnalyze` and `ideGenerate` is below. I'm still not comfortable converting even simple Silver code to Java. I mentioned on #81 that even how to simply get the `foldableRanges` attribute from the root in Java isn't entirely clear to me.

I understand and agree about state passing to `ideGenerate` and will try to get around to it.

```
function ideAnalyze
IOVal<[IdeMessage]> ::= args::[String]  svParser::SVParser  sviParser::SVIParser ioin::IO
{
  local unit :: IOErrorable<Decorated Compilation> =
    cmdLineRunInitial(args, svParser, sviParser, ioin);

  return case unit.iovalue of
  | left(re) -> ioval(unit.io, [makeSysIdeMessage(ideMsgLvError, re.message)])
  | right(comp) -> ioval(unit.io, getAllBindingErrors(comp.allGrammars))
  end;
}

function ideGenerate
IOVal<[IdeMessage]> ::= args::[String]  svParser::SVParser  sviParser::SVIParser  ioin::IO
{
  local unit :: IOErrorable<Decorated Compilation> =
    cmdLineRunInitial(args, svParser, sviParser, ioin);

  return case unit.iovalue of
  | left(re) -> ioval(unit.io, [makeSysIdeMessage(ideMsgLvError, re.message)])
  -- this is only called if the previous does NOT fail. So there should be no additional errors.
  | right(comp) -> ioval(runAll(sortUnits(comp.postOps), unit.io).io, [])
  end;
}
```

PR #102, opened 2017-01-26
Fix/scripts

krame505:
Fix minor syntax error in deep-rebuild script.  Merge this after #101 since I somehow accidentally made this fix from the wrong branch.

### DIFF
--- a/deep-rebuild
+++ b/deep-rebuild
@@ -22,7 +22,7 @@ ant
 mv silver.composed.Default.jar ../jars/
 
 # Modifications may need to be made to how core translates, to accomodate translation/runtime changes
-if [ $# -gt 0 && "$1" == "--newcore" ]; then
+if [ $# -gt 0 ] && [ "$1" == "--newcore" ]; then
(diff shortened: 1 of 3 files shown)


tedinski:
1. Script bit lgtm.
2. We already talked about the problem with the dependency on langutil from core. That's a definite no-go.
3. If we're introducing a `builtinLoc` should we have a `bogusLoc` as well? What about a `generatedLoc` construct? I've resisted trying to stuff this stuff into core previously just because I think there's a good design question to solve here. But maybe it's not such a big deal, we can deal with it later? We can always refactor code, I guess. :)
4. MAAAAJJJOR possible issue: there may be Java code in the runtime that assumes an `NLocation` is always a `Ploc`. I don't know. But this should be investigated. Perhaps this would be a good reason to make `builtinLoc` a function that just yields a `loc`? Is there a reason to prefer it as a separate production?

krame505:
Merging, since location stuff already got merged by Eric in the other pull request

PR #106, opened 2017-01-27
Change Location and Message to add proper support for built-in locations

krame505:
Ted's (relevant) comments from before:
3. If we're introducing a builtinLoc should we have a bogusLoc as well? What about a generatedLoc construct? I've resisted trying to stuff this stuff into core previously just because I think there's a good design question to solve here. But maybe it's not such a big deal, we can deal with it later? We can always refactor code, I guess. :)
4. MAAAAJJJOR possible issue: there may be Java code in the runtime that assumes an NLocation is always a Ploc. I don't know. But this should be investigated. Perhaps this would be a good reason to make builtinLoc a function that just yields a loc? Is there a reason to prefer it as a separate production?

### DIFF
--- a/grammars/core/Location.sv
+++ b/grammars/core/Location.sv
@@ -1,13 +1,11 @@
 grammar core;
 
-import silver:langutil only unparse;
-
 annotation location :: Location;
 
 {--
  - Data structure storing location information on tree nodes from a parse.
  -}
-nonterminal Location with unparse, filename, line, column, endLine, endColumn, index, endIndex;
+nonterminal Location with filename, line, column, endLine, endColumn, index, endIndex;
 
 synthesized attribute filename :: String;
 synthesized attribute line :: Integer;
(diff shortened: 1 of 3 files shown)


krame505:
I just grep'ed through the runtime, and issue 4 doesn't seem to be a problem.

krame505:
Changes as suggested by Eric, added bogusLoc as suggested by Ted.  Ted, does this look good to you?

tedinski:
Sure, looks okay.

On Jan 26, 2017 6:51 PM, "Lucas Kramer" <notifications@github.com> wrote:

Changes as suggested by Eric, added bogusLoc as suggested by Ted. Ted, does
this look good to you?

—
You are receiving this because you are subscribed to this thread.
<https://github.com/melt-umn/silver/pull/106#issuecomment-275560436>, or mute
the thread

krame505:
I am going to merge this, unless there are any other objections.

PR #107, opened 2017-01-27
ensure we don't avoid recompiles when asking for copper dumps

### LINKED ISSUE #100: Copper should always rebuild when run with --copperdump
Currently if the parser is seen as up-to-date no copper dump is generated.  This may actually be a copper bug.

tedinski:
This modernizes the functions, uses string interpolation, and fixes the issue with `avoidRecompile` flags when dumping.

Fixes #100


tedinski:
Ping @krame505 

Does this look good to you?

krame505:
Seems fine to me, but I'm not really familiar with that part of Silver

PR #108, opened 2017-01-31
Sets sv to have higher precedence than systemverilog on neovim.

remexre:
This fixes an/the issue on neovim where you need to `set filetype=sv` every time you open a file in a way that doesn't involve invasive `init.vim` changes.


tedinski:
Hey @TravisCarlson can you review this?

TravisCarlson:
A minor issue I see is that it appends to rather than overwriting filetype.vim. I wouldn't want the install script causing my configuration file to grow each time it is run. Also, a future mistake in the install script might mean that lines would need to be removed manually rather than simply re-running a fixed install script.

Hi @remexre , I think you copied a problem from the vim install.sh in develop before the latest changes were merged. Please try replacing `setfiletype sv` with `set filetype=sv` together with the previously used ftdetect method.

https://neovim.io/doc/user/filetype.html

remexre:
@TravisCarlson That works too, yeah. I'll change the PR to do that instead.

tedinski:
I think the file should be copied to its destination, rather than created with an echo/redirect.

Oh, and I think the `VIMHOME+dummytext` bit is buggy. I suggest using something like

    SVJVM_FLAGS=${SVJVM_FLAGS:-"-Xmx1800M -Xss6M"}

(this example from support/bin/silver) to set an environment variable if it's not set already.

remexre:
@tedinski better?

(The merge conflict is because of `bugfix/vimhl`, I believe.)

remexre:
@TravisCarlson @tedinski Is this mergeable in its current state?

TravisCarlson:
I think it looks OK. @tedinski ?

tedinski:
sure

Issue #109, opened 2017-01-31
Vim highlighting w/ block comments


remexre:
```
{-- has the bug
-}
{- no bug
-}
```

There's an issue with `{--` as an... ermmm... comment starter? Is that what that's called? With that thing.

![2017-01-31-01 19 13](https://cloud.githubusercontent.com/assets/4196789/22455629/72225b24-e753-11e6-9da9-aa1bbca27dd9.png)
![2017-01-31-01 19 16](https://cloud.githubusercontent.com/assets/4196789/22455630/7380ad04-e753-11e6-8533-bbad07c3c2b6.png)

TravisCarlson:
Can you confirm that the change just pushed to the [bugfix/vimhl](https://github.com/melt-umn/silver/tree/bugfix/vimhl) branch fixes this and doesn't introduce any other issues?

ericvanwyk:
Nathan, please write a bit more in your description - specifically if this is bug in Silver or in syntax highlighting in some editor.  This is a bit too terse.

remexre:
@ericvanwyk Sorry for the terseness; it's vim.

@TravisCarlson Looking over a couple files in ableC, this issue looks fixed with no regressions in the bugfix/vimhl branch.

tedinski:
Looks like this was fixed, but the pull request didn't automatically close this issue.

PR #110, opened 2017-01-31
fix/vimhl: Issue #109, vim block comments

TravisCarlson:
Fixes Issue #109 regarding syntax highlighting of block comments in the vim editor.


tedinski:
lgtm

PR #111, opened 2017-01-31
Adds toString() for Booleans

remexre:
It does what it sounds like.

```
toString(true) == "true"
toString(false) == "false"
```


tedinski:
This pull request seems to also have some vim stuff in it. Possibly branched from the wrong base branch?

My only immediate objection: I'd like to see the test cases added to the test suite. Should just be the two lines.

remexre:
Oh, whoops. Yeah, I'll remove those and add the test cases.

remexre:
Okay, I also added tests for the rest of the `toString()`-able types as well.

tedinski:
Hah. I didn't know the float would output as 0.0.

Okay, looks good then.

ericvanwyk:
Looks good - thanks Nathan.

PR #112, opened 2017-02-07
Adds is(Left|Right)(Or)? functions.

remexre:


ericvanwyk:
Should toLeft and toRight raise some kind of explicit error when the pattern matching - as it is now - fails?   What kind of message do we get for incorrect usage when it is done this way as opposed to catching the mistaken use and generating an error message that explicitly mentions an incorrect application of toLeft/toRight?

remexre:
I might *cough cough* not know how to throw an exception in Silver... (I'll
look into that tonight)

ericvanwyk:
There is an 'error' construct for this.  The point would be that this might
yield a more helpful error message than the default one.  Perhaps try it
with both and see which is best.

tedinski:
I have a few questions.

1. What new use of `Either` is motivating these additions? (I ask because I've considered following Rust in replacing `Either` with a `Result` type instead, that has the "result type or error type" sort of semantics that seem to get universally used with `Either`. Curious if this new use is still in that mold.)
2. What code wants to use these functions, are you sure they're what we actually want? Adding functions to the standard library is a careful design problem. I've resisted adding functions like these in part because of the previous parenthetical, but also because I'm a bit against introducing more functions like `head` and `tail` that just crash if you accidentally use them wrong. (I also note there don't seem to be functions like this in Haskell: https://hackage.haskell.org/package/base-4.9.1.0/docs/Data-Either.html which doesn't mean we shouldn't have them, but again makes me curious what code will use these.)
3. Glancing at the nomenclature we have for `Maybe` I think these should be `fromLeft` not `toLeft`? (e.g. `fromJust`)

remexre:
I'm using toLeft and toRight in dcv2 for errors when interpreting; in my experience with Rust, Result is a better name 95% of the time, but there's also [a (third-party) Either](https://docs.rs/either/1.0.2/either/enum.Either.html) for the other 5% of cases.

`fromLeft` is probably a better name, I'll change that.

In terms of error handling, should `fromLeft`/`fromRight` behave like Rust's [`ok()`](https://doc.rust-lang.org/std/result/enum.Result.html?search=#method.ok)/[`err()`](https://doc.rust-lang.org/std/result/enum.Result.html?search=#method.err) and return a Maybe instead of dying?

tedinski:
> In terms of error handling, should fromLeft/fromRight behave like Rust's ok()/err() and return a Maybe instead of dying?

Probably not, it doesn't really help, people would just be using `fromJust` or something then. Can we get away with only introducing the two-argument versions? (And maybe calling then `fromLeft` instead of `fromLeftOr`?) That'd be fine with me, if it works...

remexre:
In terms of how I use it e.g. in [dcv2's `add` abstract production](https://github.com/melt-umn/dcv2/blob/develop/edu.umn.cs.melt.dcv2/abstractsyntax/AbstractSyntax.sv#L24), would I want to make its `value` attribute an `Either`/`Result`, and have that store errors encountered in interpretation, while the "other" `errors` attribute stores pre-interpretation errors (e.g. if condition not a boolean)?

tedinski:
Hmm, fair question. Would you object to `fromLeft(val, error("Evaluating a tree with errors"))`?

If so, then I suppose the one argument versions are fine, but maybe we should give them obnoxious names? `fromLeftUnsafe`?

remexre:
That's actually a much better solution; I'm still not 100% used to having
laziness w.r.t. errors, so I didn't think of that. I'll change the PR
accordingly.

tedinski:
ok looks good

PR #119, opened 2017-03-09
Updates .gitignore to ignore all jars.

### LINKED ISSUE #118: jars not inside jars/ should also be gitignored
It's pretty easy to build inside grammars and accidentally commit that jar.

remexre:
Fixes #118.


tedinski:
lgtm

PR #124, opened 2017-03-09
Adds charsToString and stringToChars.

remexre:
Does the simple thing and converts the string to/from a `char[]`, which is converted to/from a `[Integer]`. This allows manipulating nonprintable characters, as well as Unicode characters that we might not want to include in our source code.

### DIFF
--- a/grammars/core/String.sv
+++ b/grammars/core/String.sv
@@ -289,3 +289,28 @@ Boolean ::= s1::String s2::String
   return s1 <= s2;
 }
 
+{--
+ - Converts a list of code points to a string. Note that due to Java's use of
+ - UCS-2, code points greater than 0xFFFF (i.e. and characters outside the Basic
+ - Multilingual Plane) aren't supported.
+ -}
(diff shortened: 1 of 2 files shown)


tedinski:
Two issues:

1. We should have a few new tests.

2. Because we're going to change runtime and core at the same time, Jenkins is going to hiccup, just like your local build did. Are you in the `melt` group? If so, then when you merge, you'll have to run `publish-jars` manually to solve the chicken and egg problem.

remexre:
I'll add the tests; I'm probably going to close this PR though in favor of merging this via the feature/monto PR that'll probably be added within the next week (or early next) -- that way, there's one jar update instead of two. Other than that, is this good?

tedinski:
Should be.

remexre:
https://github.com/melt-umn/silver/commit/98da59a0ef935af5f6cbb826c8a8e8461d1d6c14 @tedinski do these tests look good?

PR #126, opened 2017-03-21
Adds a parseError attribute to ParseResult.

remexre:
Without this, we need to pattern match to get at the ParseError. (also see #127)


tedinski:
lgtm

PR #131, opened 2017-04-03
Adds build.xml files to .gitignore

### LINKED ISSUE #120: Some build.xml files should be gitignored.
e.g. `tutorials/hello/silver-compile` creates a `build.xml` that relies on system-dependent paths. Non-generated `build.xml`s shouldn't be gitignored, though, and I'm not sure on the best approach to ignore the right ones. I can think of three approaches:

 - Ignore all generated `build.xml`s manually in the root `.gitignore`
 - Ignore all `build.xml`s in the root `.gitignore`, then unignore (with `!`) the ones that aren't generated
 - Ignore each generated `build.xml` in a `.gitignore` that's in the directory it gets generated into

remexre:
Closes #120.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,10 @@ generated/doc
 generated/ide
 generated/src
 
+# build.xml files
+build.xml
+!runtime/java/build.xml
+
 # Editor created backup files
 *.swp
 *~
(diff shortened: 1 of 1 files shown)


tedinski:
lgtm

PR #132, opened 2017-04-04
Adds build-everything script.

### LINKED ISSUE #130: deep-rebuild doesn't rebuild all grammars
`./deep-rebuild` only rebuilds grammars `silver:composed:Default` depends on -- should this be changed?

remexre:
Closes #130.

This also deletes two grammars:

`lib:langproc:tests` doesn't build.
`silver:composed:monto` hasn't worked for a while, and won't in the near future.


tedinski:
As long as you're removing `langproc:tests` with this, I endorse removing `langproc` entirely I think.

You'll need to remove the import of it from that one build everything grammar in the documentation extension.

remexre:
Removed `lib:langproc`, everything else builds properly.

remexre:
@tedinski This OK to merge?

tedinski:
Sure.

PR #134, opened 2017-04-06
Feature/monto+highlight

remexre:
Includes a lot of @TravisCarlson's work on the [`feature/ide`](https://github.com/melt-umn/silver/tree/feature/ide) branch, and some FFI bindings from `getTokens` to Silver. This is accessed via the `parseTerminals` attribute on `ParseResult`.

Additional functions are added to `lib:monto:helpers` to allow the language designer to easily add highlighting support -- adding the callback is [only one line](https://github.com/melt-umn/dcv2/blob/28effcf89d9e891f0d294392409bcd0c00f4dfaf/edu.umn.cs.melt.dcv2/monto/Main.sv#L12).

Unfortunately, the code to convert between lexer classes and styles is [significantly longer](https://github.com/melt-umn/dcv2/blob/28effcf89d9e891f0d294392409bcd0c00f4dfaf/edu.umn.cs.melt.dcv2/monto/Highlighting.sv).

The highlighting protocol is implemented [as documented](https://monto-editor.github.io/ir/#syntax-highlighting), however.


grammars/silver/modification/impide/cstast/Syntax.sv
```diff
@@ -58,9 +58,9 @@ top::SyntaxDcl ::= n::String regex::Regex_R modifiers::SyntaxTerminalModifiers
     "      " ++ convertAssocNXML(modifiers.opAssociation) ++ "\n" ++ -- TODO
     "    </Operator>\n"
     else "") ++
-    "    <Type>common.TerminalRecord</Type>\n" ++ 
+    "    <Type>common.Terminal</Type>\n" ++ 
     "    <Code><![CDATA[\n" ++ 
-    "RESULT = new common.TerminalRecord(lexeme,virtualLocation,(int)getStartRealLocation().getPos(),(int)getEndRealLocation().getPos());\n" ++
+    "RESULT = new common.Terminal(lexeme,virtualLocation,(int)getStartRealLocation().getPos(),(int)getEndRealLocation().getPos());\n" ++
```
tedinski: I think this might be a bug where it should now be using `makeTerminalName` to construct the correct class?

test/cstast/Test.sv
```diff
@@ -34,6 +34,7 @@ global obj::SyntaxRoot =
...
...host\"/></Grammars>\n    <StartSymbol><NonterminalRef id=\"Foo\" grammar=\"host\" /></StartSymbol>\n    <StartLayout></StartLayout>\n  </Parser>\n\n  <Grammar id=\"host\">\n\n    <PP>host</PP>\n\n    <Layout></Layout>\n\n    <Declarations>\n      <ParserAttribute id=\"context\">\n        <Type><![CDATA[common.DecoratedNode]]></Type>\n        <Code><![CDATA[context = common.TopNode.singleton;]]></Code>\n      </ParserAttribute>\n  <Terminal id=\"XTerm\">\n    <PP>XTerm</PP>\n    <Regex><CharacterSet><SingleCharacter char=\"x\"/></CharacterSet></Regex>\n    <Type>common.Terminal</Type>\n    <Code><![CDATA[\nRESULT = new common.Terminal(lexeme,virtualLocation,(int)getStartRealLocation().getPos(),(int)getEndRealLocation().getPos());\n]]></Code>\n    <InClasses></InClasses>\n    <Submits></Submits>\n    <Dominates></Dominates>\n  </Terminal>\n\n  <Nonterminal id=\"Foo\">\n    <PP>Foo</PP>\n    <Type><![CDATA[Foo]]></Type>\n  </Nonterminal>\n  <Production id=\"foo\">\n    <Code><![CDATA[\nRESULT = new foo();\n]]></Code>\n    <LHS><NonterminalRef id=\"Foo\" grammar=\"host\" /></LHS>\n    <RHS></RHS>\n  </Production>\n    </Declarations>\n  </Grammar>\n</CopperSpec>\n", String, csttests );
```
tedinski: This test is so fragile I think we should maybe remove it.

grammars/silver/modification/impide/cstast/Syntax.sv
tedinski: Hah! Terminal is abstract. This code hasn't been tested yet. Hey @TravisCarlson do you remember if these changes so far mean the eclipse code is broken?

remexre:
I'm putting this PR on hold -- the current plan is to remove the Eclipse plugin from the Silver tree, and make an alternate Eclipse plugin that gets information from Monto. Until then, this branch should **NOT** be merged, as it'll probably break the Eclipse plugin.

tedinski:
Er, yeah, those last two commits didn't go in the right direction I don't think, but I'm not sure the earlier commits shouldn't be merged sooner rather than later?

I don't think the plan should be to remove the current eclipse plugin at all, at least not any time soon.

To do highlighting, I suggest NOT having special `MONTO_KEYWORD` lexer classes or the like, but rather having a classifier function that can map lexer class names to coloring. e.g. we should be able to say "map any token with the lexer class 'silver:definition:core:Keyword' to Monto's keyword highlighting group". Does that make sense? Then highlighting is non-invasive.

remexre:
That's what the original one did; it was kind of "yuck" though, so after
talking with Eric I changed it to use the Monto-defined lexer classes.

tedinski:
I'm not sure that's exactly the plan for the eclipse plugin that currently exists. Or at least, the way I'd do it would be to leave it intact until there's a separate monto-based eclipse plugin that we're satisfied is much better all around.

I'm not sure why the approach I suggested is "yuck". The only annoying thing about it is that you need to give it the full names of some lexer classes, but that's easily fixable once version one is there and working. This approach (having to *modify* a language's implementation in order to add IDE support) is more "yuck" in my opinion. (Though of course, that's also how we currently do things with the existing eclipse plugin, too! That's why the plan was to change this.)

Anyway, does this branch (minus those last two commits) break anything? If not, I'd rather see us merge this immediately, rather than delay merging that existing work in favor of getting more done first.

ericvanwyk:
We do plan on leaving the current one there until/if the Monto one is
complete and better.

I think Nathan is going to implement mechanisms to do both versions - then
we can see which is more "yuck" ...  I think each may have situations in
which it is preferred.

Agreed - merging soon is better than later.  The Monto-Eclipse thing should
be its own repo and we should be able to have two Monto and current Eclipse
live together.

remexre:
The problem with merging now is that then we have large changes to the current Eclipse plugin -- I'd rather there were a clean break than a large change and then a change in plugins later on. If you still disagree, I can merge through 7d9a5cc.

tedinski:
Just quick clarification: by large changes, you're referring to what's already in this patch set?

That's fine! Those are all improvements, no? That's actually what I'd be after: build up Silver's support for doing IDE things by making the underlying design work for both Monto and the Eclipse plugin (and maybe even other things.)

Part of the point of this is to tweak how we design things so that building IDE tooling of any kind is easy.

remexre:
Okay, so I should merge 1c7b38e..7d9a5cc, then?

tedinski:
I think so. Can you walk the branch back and then force push (just this branch!) so I can do a once over quick?

At least, I *think* that's how it should be done? Maybe you know a better way?

remexre:
Think that's done?

tedinski:
Ok, let's merge this.

remexre:
As soon as I get my JARs back to what they were at 7d9a5cc, I will. (I really want to try getting Jenkins to save a set of JARs per pushed commit, if for no other reason than my sanity...)

tedinski:
Hah. On the TODO list.

tedinski:
I assume this is because you need to do a publish-jars for this pull request?

remexre:
Yeah. It's not being a fun time...

remexre:
Okay, finally got it. Publishing the JARs, then merging.

krame505:
Currently, the eclipse plugin is totally broken as a result of merging this.  Please, in the future, check that things actually *work* before merging, not just that it compiles.  Can we roll this back for the time being, at least until this is fixed?

krame505:
OK, I am going to roll this back for now so that I have something to use for now.  Let's merge this when the revised (non-monto) plugin is working...

krame505:
@tedinski Since the jars are now published for the updated version, I can no longer run deep-rebuild.  Do we have a backup somewhere or an easy way to roll this back?

tedinski:
Just to be sure I understand what just happened... this broke the eclipse plugin, so you reverted it, and now everything is broken?

krame505:
Pretty much.  I did ask Eric first...

krame505:
OK, we grabbed those jars from Jenkins and have things building again.  Should we go ahead and publish jars?

krame505:
OK, we grabbed those jars from Jenkins and have things building again.  Should we go ahead and publish jars?

tedinski:
Okay.

After that, we need to figure out how to get it to see changes in this branch again. Maybe a rebase?

remexre:
Just rebased. The status is still Merged, but the "merged banner" is gone? I'll make a new PR...

tedinski:
I'm not sure it worked. It now claims to be even with develop.

remexre:
!()[https://imgur.com/a/EE3jX]

Beat me by a second

tedinski:
Does this help? http://stackoverflow.com/questions/18462765/rebase-reverted-merged-branch

remexre:
The monto-in-silver branch seems to be still PR-able, and GitHub says merging it will readd the stuff removed. I'm gonna PR it, and put a huge disclaimer warning at the top...

remexre:
Wait, it doesn't readd most of the stuff. Bleuggggghhhhhhhh

tedinski:
I have a maybe simple idea. What if you add a new commit to this branch that just reverts the revert commit?

remexre:
Oh, hey, now communication's even more fragmented :P

remexre:
Just pushed it, doesn't look like it's showed up here, although it does let me make a new PR now.

tedinski:
Ok. I'm going to go fix bugs with this branch then.

In order to make sure everything goes without conflict, I'll probably branch off this branch, then pull request the changes into this branch. Exciting!

remexre:
/shudder

There might be largeish changes to how `lib:monto` works too, in order to address the concerns raised earlier (e.g. config without State monad, the need for preprocessing, dynamic dependencies with unknown paths, etc.)

I'll probably be making those changes in `feature/monto-in-silver`, though, because they're largely in response to the difficulties of getting `lib:monto` to work with a "real" language (ie. not dcv2), with imports, extensions, directory scope instead of file scope, etc.

tedinski:
Sure. I still want this branch to be merged sooner rather than later, of course. So I'm going to go try to fix a bug with prefixes, and the fix the eclipse plugin. Then I'd like to see this merged again.

AFAIK, these are the only issues this branch introduced.

remexre:
Wrt the eclipse plugin, the commits on [`bugfix/eclipse-plugin-build-fails`](https://github.com/melt-umn/silver/commits/bugfix/eclipse-plugin-build-fails) "fix" it. The problem is that we need to either:

 - hardcode SILVER_HOME into the plugin
 - bundle the Silver jars into the eclipse plugin
 - Make SILVER_HOME part of the Eclipse config for the plugin

`bugfix/eclipse-plugin-build-fails` goes with the first approach; I don't know the Eclipse API or Maven/POM well enough to do the others.

tedinski:
I haven't investigated yet, but I can only say "that's not right." We bundled the silver jars into the eclipse plugin ages ago. I don't know how any of our changes would have disappeared them.

PR #142, opened 2017-04-20
Revert "Feature/monto+highlight"

krame505:
Reverts melt-umn/silver#134


tedinski:
wait what?

krame505:
We agreed to temporarily roll this back until the IDE is fixed.  Now the published jars are for the version including this branch, so I can't run `deep-rebuild`.  Now what?

tedinski:
Okay. Short coordination discussion.

Right now we've got a few things going wrong ("everything is on fire" -Lucas's appropriate description).

1. The biggest issue is a bunch of builds are broken because of my brilliant addition of `concat` to `core`.
2. The next biggest issue is there are several bugs introduced as a result of this work.
3. Jenkins is presently busted. Ugh.

I would like to work on 2 right now, but reverting this merge stops me from doing that. From the other thread, I thought you were working on item 1.

krame505:
When was the last time we published jars?  If we haven't since this pull request got merged, then there should still be a backup somewhere we can restore, right?

tedinski:
You should be able to download build artifacts from jenkins. It appears it isn't so broken that this isn't possible.

I really wish you hadn't done this revert. Github no longer seems to think there are any changes in the monto feature branch. I think it has only made the situation more complicated.

PR #144, opened 2017-04-20
Get eclipse IDE working again with monto changes

tedinski:
I've cherrypicked some of the commits from `bugfix/eclipse-plugin-build-fails`, and fixed another bug. I believe that branch can be retired now.

(There are still some commits in `feature/ide` that are worth preserving though. Eventually we'd like to get that to work, but to the best of my knowledge, that work wasn't completed. (That being to have an `eclipse-plugin` directory at the root, instead of generating the plugin under `generated/ide`.) Is that right @TravisCarlson ?)

This produces the eclipse IDE plugin working the old way, with the new changes to `Terminal`.

This should leave the ide in working state, and make this branch mergable into develop.

**Note that this pull request merges into the feature branch, not develop**


tedinski:
@remexre any thoughts on this?

@krame505 could you glance at and sign off on the prefixes commit here?

krame505:
It's been so long since I worked on that I'm not really sure I know what is going on here.  What exactly did you fix?  

Also I don't really understand your comment that prefix separators
> are a def we need to be scoped to this parser only, not to be overheard by other parsers in this grammar.

The idea behind this is that the host defines what the prefix separator should be as a standard, and composed parser declarations see this automatically and use this when a prefix is specified.

remexre:
lgtm; I don't think anything I've written depends on CopperToken et al

tedinski:
Thanks yall.

@krame505 Perhaps I am misunderstanding something. I see `prefixSeparatorAGDcl` but I also see `prefixSeparatorParserComponent`. This later one uses `liftedAGDcls` to create the former, but this approach would not work if it did as I suggested and actually raised these declarations to the top-level scope. Multiple parsers would see the same declaration in scope.

Really, there's already a problem with multiple parsers and scoping with this design, since if you import multiple host languages, you'll see multiple `_prefix_seperator` symbols. Or the wrong one, if one grammar doesn't define it, but another does. I think this feature is ill-defined right now.

I think my work around doesn't make anything worse though. But perhaps I should also open a new issue about this?

krame505:
Yeah, that is basically correct.  For the record, I was against specifying the prefix separator globally, just make this a convention for the language.  

You are right, importing multiple host languages would cause an issue with this- something we may run into when we do a Silver extension for constructing ableC ASTs.  I guess you could work around this by structuring things in a way where you can import all the host concrete syntax except for the prefix separator, but this makes things even more confusing... 
 I am personally in favor of just removing that feature altogether, since it adds so much complexity for very little actual gain... I agree we should open an issue on this.  

I'm still not totally clear on what you were fixing with this workaround- does this have something to do with the refactoring happening for Monto?

tedinski:
This makes the branch mergeable without breaking eclipse. And fixes that bug, too, which broke a test when this was merged.

krame505:
Sorry, I'm still not clear to which bug you are referring?  The issue of importing multiple host languages that define a prefix separator, or something else?

tedinski:
Ah, the actually fixed bug in this commit. There's a class for Terminals now, so this commit changes prefixes defined within parsers to not have their `genFiles` suppressed.

krame505:
Oh, that sounds reasonable.  This commit LGTM.

ericvanwyk:
I'm missing something here.  Lucas or Ted, lets discuss this next time
you're in the lab.

Perhaps we need a mechanism for building a parser by adding new grammars to
an existing parser.  That existing parser could specify its prefix
separator and then we don't have this at the global level.  The MDA tests
look like this - one names a parser.  But they also name the host grammar,
which is a bit odd.

tedinski:
I'll open an issue about that problem.

PR #158, opened 2017-05-01
Add flowtype declarations to the language

### LINKED ISSUE #114: Flow analysis lacks a way to specify flow types manually
Right now we don't have a way to provide an explicit flowtype for a nonterminal, this can only be inferred automatically.  This is an issue because debugging flow errors in the host is a huge pain, involving grepping through the flow graphs to find what is introducing a dependency.  If we had a way to fix the dependencies for an attribute on a nonterminal, then these issues would simply be reported as errors in the places where these new dependencies are introduced.  
This would also remove the need for 'hackUnused' productions that have a bunch of dummy equations to force the flow types to include extra attributes that we want extensions to be able to use.

### LINKED ISSUE #115: More specific Decorated types
Currently when we decorate a nonterminal and store the decorated value, we require all inherited attributes to be present.  This doesn't really make sense, since there are some attributes that should only ever be used in a very specific way, such as a parameter attribute for a specific analysis which doesn't really need to be present for anything else.  This can be solved by moving those attributes and equations to a separate grammar or specifying dummy equations everywhere and hoping nobody tries to use them, but doing so is either a lot of extra work to introduce a new dependant analysis, or is at least really hacky.  

To remedy this, we want to add some sort of distinction on inherited attributes, between ones that are and aren't required to be present in the Decorated form of a nonterminal.  The flow analysis could then report warnings when a 'non-general' synthesized attribute depending on one of these is accessed on a Decorated nonterminal.  In a more general way, it may be interesting to have some notion of 'partial decoration', or a variation of Decorated type that explicitly has only some of its attributes provided.  This could even be taken to the extent of allowing a partially decorated nonterminal to be further decorated with additional attributes, but this would probably be less useful.

### LINKED ISSUE #153: Flow type specifications
Hi all, give syntax suggestions.

I'm adding the ability to declare flow types. So far I've got this idea:

```
flowtype Expr : forward {context, env}, errors {env}, pp {};
```

Now, there will also be convenience extensions, I think. If I get around to that, too:

```
nonterminal Expr with pp {}, errors {env}, forward {env};
```

but I'm mostly concerned about the standalone syntax, since it seems different than everything else.

Maybe I should add it as part of attribute occurrence declarations?

```
attribute errors occurs on Expr with {env};
```

But actually, that might just be another convenience extension feature. We still need to specify for forwards, after all, which isn't an occurrence.

Also, I'd like this to eventually be able to specify reference sets, too. Thoughts on that syntax?

```
flowtype Env : decorate {};
```

Maybe?

### LINKED ISSUE #155: Flow analysis crashes with missing grammars
I've had a hard time reproducing this for a smaller example, but building and running the flow analysis for an ableC extension artifact without including ableC in the include path causes the following crash:
```
...
(DN.254): While evaling local 'silver:definition:core:forwardsTo:local:mayAffectFlowType' in 'silver:definition:core:forwardsTo' (2a6c751f, SKELETON.sv:12:2)
(silver.driver.util.PisExportedBy in PisExportedBy.java:113): Error while evaluating function silver:driver:util:isExportedBy
(silver.util.Pcontains in Pcontains.java:104): Error while evaluating function silver:util:contains
(core.Pnull in Pnull.java:93): Error while evaluating function core:null
(silver.driver.util.PcomputeOptionalDeps in PcomputeOptionalDeps.java:104): Error while evaluating function silver:driver:util:computeOptionalDeps
(core.Pnull in Pnull.java:93): Error while evaluating function core:null
(silver.util.Prem in Prem.java:104): Error while evaluating function silver:util:rem
(core.Pnull in Pnull.java:93): Error while evaluating function core:null
(DN.254): While evaling local 'silver:driver:util:computeOptionalDeps:local:closeOptions' in 'silver:driver:util:computeOptionalDeps' (433b1597)
(silver.driver.util.PexpandOptionalsIter in PexpandOptionalsIter.java:113): Error while evaluating function silver:driver:util:expandOptionalsIter
(core.Pnull in Pnull.java:93): Error while evaluating function core:null
(DN.254): While evaling local 'silver:driver:util:computeOptionalDeps:local:initPlusExported' in 'silver:driver:util:computeOptionalDeps' (433b1597)
(silver.driver.util.PcomputeDependencies in PcomputeDependencies.java:104): Error while evaluating function silver:driver:util:computeDependencies
(silver.driver.util.PexpandCondBuilds in PexpandCondBuilds.java:122): Error while evaluating function silver:driver:util:expandCondBuilds
(core.Pnull in Pnull.java:93): Error while evaluating function core:null
(silver.driver.util.PexpandExports in PexpandExports.java:113): Error while evaluating function silver:driver:util:expandExports
(core.Pnull in Pnull.java:93): Error while evaluating function core:null
(DN.254): While evaling local 'silver:driver:util:expandExports:local:g' in 'silver:driver:util:expandExports' (6dd2e453)
(silver.definition.env.PsearchEnvTree in PsearchEnvTree.java:104): Error while evaluating function silver:definition:env:searchEnvTree
(silver.util.raw.treemap.Plookup in Plookup.java:102): Error while evaluating function silver:util:raw:treemap:lookup
(core.Phead in Phead.java:93): Error while evaluating function core:head
(DN.254): While evaling local 'silver:definition:core:forwardsTo:local:ntDefGram' in 'silver:definition:core:forwardsTo' (2a6c751f, SKELETON.sv:12:2)
(silver.definition.flow.env.PhackGramFromFName in PhackGramFromFName.java:95): Error while evaluating function silver:definition:flow:env:hackGramFromFName
(core.Psubstring in Psubstring.java:111): Error while evaluating function core:substring
(java.lang.String in String.java:1967): String index out of range: -1
```

tedinski:
This introduces flow type declarations support.

Fixes #153 
Fixes #155 
Fixes #115 
Fixes #114 

I have a separate branch I'll also merge with this one, the additional diff of which can be found here:

https://github.com/melt-umn/silver/compare/feature/flowspec...feature/flowspecuse

This extra branch actually makes use of this feature in Silver itself. I want to merge this one, let Jenkins build it, then merge that one to avoid any bootstrapping issues.

### DIFF
--- a/grammars/silver/analysis/warnings/exporting/Graph.sv
+++ b/grammars/silver/analysis/warnings/exporting/Graph.sv
@@ -58,8 +58,8 @@ String ::= specs::[Decorated RootSpec]
 function makeDotArrow
 String ::= f::String t::String
 {
-  -- A few stupid heuristics to try to make the graphs more readable:
-  return if t == "core" || startsWith("silver:util", t) || startsWith("silver:langutil", t) then "" 
+  -- A heuristic to try to make the graph more readable...
+  return if t == "core" then "" 
   else "\"" ++ f ++ "\" -> \"" ++ t ++ "\";\n";
 }
 
--- a/grammars/silver/definition/core/AGDcl.sv
+++ b/grammars/silver/definition/core/AGDcl.sv
@@ -32,7 +32,6 @@ top::AGDcls ::= h::AGDcl t::AGDcls
 {--
  - A semantically meaningless declaration. Does nothing.
  - Used for: (1) 'nil' counterpart to appendAgDcl
- - (2) annotations or other meaningless AGDcls.
  -}
 abstract production emptyAGDcl
 top::AGDcl ::=
@@ -47,7 +46,6 @@ top::AGDcl ::= e::[Message]
 {
   top.pp = s"{- Errors:\n${foldMessages(e)} -}";
   top.errors := e;
-  forwards to emptyAGDcl(location=top.location);
 }
 
 {--
--- a/grammars/silver/definition/flow/ast/Flow.sv
+++ b/grammars/silver/definition/flow/ast/Flow.sv
@@ -2,8 +2,8 @@ grammar silver:definition:flow:ast;
 
 imports silver:definition:env only quoteString,unparseStrings, unparse;
 
-nonterminal FlowDefs with synTreeContribs, inhTreeContribs, defTreeContribs, fwdTreeContribs, fwdInhTreeContribs, unparses, prodTreeContribs, prodGraphContribs, refTreeContribs, localInhTreeContribs, extSynTreeContribs, nonSuspectContribs, localTreeContribs;
-nonterminal FlowDef with synTreeContribs, inhTreeContribs, defTreeContribs, fwdTreeContribs, fwdInhTreeContribs, unparses, prodTreeContribs, prodGraphContribs, flowEdges, refTreeContribs, localInhTreeContribs, suspectFlowEdges, extSynTreeContribs, nonSuspectContribs, localTreeContribs;
+nonterminal FlowDefs with synTreeContribs, inhTreeContribs, defTreeContribs, fwdTreeContribs, fwdInhTreeContribs, unparses, prodTreeContribs, prodGraphContribs, refTreeContribs, localInhTreeContribs, extSynTreeContribs, nonSuspectContribs, localTreeContribs, specContribs;
+nonterminal FlowDef with synTreeContribs, inhTreeContribs, defTreeContribs, fwdTreeContribs, fwdInhTreeContribs, unparses, prodTreeContribs, prodGraphContribs, flowEdges, refTreeContribs, localInhTreeContribs, suspectFlowEdges, extSynTreeContribs, nonSuspectContribs, localTreeContribs, specContribs;
 
 {-- lookup (production, attribute) to find synthesized equations
  - Used to ensure a necessary lhs.syn equation exists.
@@ -62,6 +62,9 @@ synthesized attribute extSynTreeContribs :: [Pair<String FlowDef>];
 {-- A list of attributes for a production that are non-suspect -}
 synthesized attribute nonSuspectContribs :: [Pair<String [String]>];
 
+{-- Explicit flow type specificiations -}
+synthesized attribute specContribs :: [Pair<String Pair<String [String]>>];
+
 synthesized attribute unparses :: [String];
 
 abstract production consFlow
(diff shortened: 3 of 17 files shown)


tedinski:
Hey @krame505 would you like to take a look?

krame505:
Seems generally fine to me, although I don't really understand the internals of the analysis.  
I still prefer the syntax
`flowtype Foo with ...;`
over
`flowtype Foo = ...;`
since the latter seems to much like
`type Foo = ...;`
which could be confusing, and the former nicely mirrors nonterminal declarations.  I don't remember if we ever came to a final decision on this?

tedinski:
I went with = because `with` didn't make sense to me.

Interestingly, in the other branch where I use it, I ended up most often using the other form (`flowtype forward {env} on Nt1, Nt2, Nt3, Nt4, Etc;`) so I expect that's going to be the one that really ends up getting used.

RE: How it works, it looks a bit subtle, judging by the diff. 

The trick is basically two changes:

1. We compute an "initial flow type" based on the flowtype declaration info.
2. We remove those entries from the "flow vertexes" list stored in the `ProductionGraph`. This means that they don't get updated during the inference procedure.
3. For `decorate` flowtypes, which are special, we just look up and use the specification, if any, instead of the inferred set at the nonterminal declaration.

Everything else is syntax and plumbing.

krame505:
Makes sense.  So then is specifying the default flowtype for an attribute future work?

tedinski:
Yeah, I didn't add anything convenience related yet. I think we need a better sense of how things shake out before that's possible.

e.g. I don't know that adding a flowtype per-attribute is actually all that useful, now that I've tried using it!

Instead, what might be useful is making attributes a super-set of the forward flowtype by default. Presently, we enforce that only for "extension" attributes. I think one possible lesson we might learn after adding these declarations is that we want that for EVERY synthesized attribute, unless there's an explicit flowtype specification for it.

WHICH REMINDS ME... I need to add a check for that in explicit flow types. On it now...

tedinski:
Ok, now I think that should do it...

krame505:
Wait, this was supposed to be happening implicitly?  In ableC we have the `substituted` synthesized attribute which has a flowtype only including the new inherited attribute `substitutions`.  This is defined in a grammar not exported by the host in any way, but there were no flow errors reported.  I am confused what was going on here now.

tedinski:
Yeah, it should have been happening implicitly. Were there `options` involved or anything?

krame505:
Nope.  the grammar is never imported or exported anywhere in the host (just in the template extension), but these don't report any errors: https://github.com/melt-umn/ableC/blob/develop/edu.umn.cs.melt.ableC/abstractsyntax/substitution/Util.sv

tedinski:
I'll take a look next.

PR #161, opened 2017-05-16
Properly check flow type within functions

### LINKED ISSUE #160: Flow analysis falls down on functions
We do not presently build "production graphs" for functions.

This is correct behavior globally (functions cannot affect flow types) but it has disastrous consequences for checking for flow errors in the body of the function, when we'd like to discover transitive dependencies of an access.

For example,

```
function foo
String ::= nt::Nt
{
  nt.inh1 = [];
  return nt.syn;
}
```

if `syn` depends also on `nt.inh2` there is no error raised for missing equations. This expansion was generally done by the production graph, which we don't have. So all that's checked is if the `return` equation _directly_ depended on an equation like `inh1`.

I think I'm going to switch over to locally building a production graph for functions. This will be a lot less broken than the current approach, but it will still be ignorant of `aspect function`s. But oh well?

This is one of the contributing causes to the lack of errors you saw in AbleC for the substitutions thing, @krame505

tedinski:
Fixes #160 

This pull request does 3 things:
1. Fix a flow vertex confusion bug in phantom productions that was causing extension synthesized attributes to not necessarily get their minimum flow type set.
2. A ton of refactoring eliminating `top.signature` and `top.blockContext` merging into `top.frame`.
3. Using this to send the local production graph down, including constructing them for functions, so functions can do flow analysis somewhat more normally.

There is a possibility that the Silver compiler may be more prone to crashes when compiling broken code. (There were some previously silently-wrong dummy-values being used, where now it `error()`s out.) If you run into one, send a bug report!

This patch is pretty extensive I think, so reviewing should be hard. I recommend going commit by commit rather than all changes at once. And probably just skimming it mostly. @krame505 ?

### DIFF
--- a/grammars/silver/analysis/typechecking/core/ProductionBody.sv
+++ b/grammars/silver/analysis/typechecking/core/ProductionBody.sv
@@ -60,7 +60,7 @@ top::ProductionStmt ::= 'forwards' 'to' e::Expr ';'
   errCheck1.downSubst = e.upSubst;
   top.upSubst = errCheck1.upSubst;
 
-  errCheck1 = check(e.typerep, top.signature.outputElement.typerep);
+  errCheck1 = check(e.typerep, top.frame.signature.outputElement.typerep);
   top.errors <- if errCheck1.typeerror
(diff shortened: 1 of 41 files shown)


krame505:
Seems fine to me.  But I'm not super familiar with this part of Silver...

PR #164, opened 2017-05-17
Reduce output slightly, and print error messages better

### LINKED ISSUE #151: Reducing build verbosity - pass 1
Leaving aside the question of a `verbose` flag that reduces output further still...

I'm thinking about making two changes.

One: instead of printing full path and every file when parsing as we currently do, I think we might instead print something like this

```
Compiling Grammar: silver:langutil
	At [/home/tedinski/repos/silver/grammars/silver/langutil/]
	Found [Location.sv Attributes.sv DocConfig.sv Message.sv]
```

The downside to this is that it will make it more difficult to open these files. (Right now, you can double click the name and the whole path gets highlighted, makes things quite easy...) But I think we can improve the error messages to include the full path to the file, solving that issue.

Two: I don't know that there's any reason why we have separate passes for generating interface files and generating other translation. I intend to unify them into one.

tedinski:
This isn't a huge reduction, but it helps a bit. I'd give an example of the new output but my VM has mysteriously stopped syncing copy & paste... >:(

The biggest change, really, is that errors now print out the full path of a file before that file's contents.

When compiling grammars, we now print out the path we found the grammar at, and then each of the files within it. This saves some lines.

Parser errors now report full path as well, where they previously were a bit mysterious (though usually those would be in a file you're already editing...)

Fixes #151

### DIFF
--- a/grammars/silver/composed/idetest/Analyze.sv
+++ b/grammars/silver/composed/idetest/Analyze.sv
@@ -106,8 +106,8 @@ function getIdeMessages
 {
   return map(rewriteMessage(path, _), 
     if !null(spec.parsingErrors)
-    then spec.parsingErrors
-    else spec.errors);
+    then flatMap(snd, spec.parsingErrors)
+    else flatMap(snd, spec.grammarErrors));
 }
 
 function rewriteMessage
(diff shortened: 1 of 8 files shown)


tedinski:
This isn't an especially complicated change, so I'm going to merge and then see how people like the new output when they get experience with it.

PR #168, opened 2017-05-23
doc/determinism

200sc:
Adds some documentation detailing how to run MDA analysis and listing several common errors, as well as some approaches to solve those errors.

There's at least one error case that isn't detailed here because I've never seen it, and that's the non I-subset, and it's possible that my understanding of the non IL-subset is incorrect. This doc also has a lot of pseudo-todo notes in it on determinism error reports, which I'm going to look into improving so the errors reported are more specific and don't crash.

### DIFF
--- a/documentation/concepts/determinism/index.md
+++ b/documentation/concepts/determinism/index.md
@@ -8,7 +8,7 @@ title: Modular determinism
 With a file such as:
 
 ```
-grammar determinism
+grammar example:grammar
 
 import host:grammar;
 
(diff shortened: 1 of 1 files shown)


documentation/concepts/determinism/index.md
```diff
@@ -0,0 +1,102 @@
...
+
+Unlike in well-definedness, determinism tends to report errors quickly. Determinism will usually stop evaluating errors after the first error it reports, as a lot of errors are reported as bugs or crashes in copper (ideally this will change). Determinism will also not evaluate any grammars following one that fails in a list of ```copper_mda``` definitions, and will evaluate those definitions in order from top to bottom.
+
+---
+
+### Types of Errors Reported
+
+1. <... a long stack trace> ... Requested head of nil.
+
+   This is a bug in copper, where it doesn't check the length of a list before obtaining the first element of the list.
```
tedinski: Really this is a bug in Silver, and the explanation should really be that "Silver fails to notice that a necessary grammar was left out of the parser specification."

Likewise below, the solution I'd say something like "Make sure you've included the right set of grammars that add to concrete syntax (don't forget any terminals or lexer classes, if those are separate)"

documentation/concepts/determinism/index.md
```diff
@@ -0,0 +1,102 @@
...
+
+   A non-marking terminal that has been added by your grammar follows a non terminal of the host language. Lookahead spillage tends to come as a result of follow spillage.
+
+   ##### Potential solutions:
+
+   Re-work the grammar so that this is no longer the case. 
+
+5. [javac] ... error: cannot find symbol
+
+   This is the result of outdated java definitions holding on to a production or terminal that has been removed. Silver --clean does not automatically do this removal right now.
```
tedinski: Wait, what? This shouldn't happen. `--clean` should fix everything. Details?

documentation/concepts/determinism/index.md
```diff
@@ -0,0 +1,102 @@
...
+
+2. Missing Bridge Productions 
+
+   Example:
+   ```
+   ... Invalid content was found starting with element 'ProductionRef'. No child element is expected at this point.
+   ... [grammar ext]: Extension grammar is missing the following required attributes: [bridgeProductions]
+   ```  
+
+   This indicates that the automatic scan copper performs to find bridge productions, which are productions that begin with marking terminals, was unsuccessful. Either:
```
tedinski: Silver performs, actually.

And maybe instead of "unsuccessful", say that it didn't find any bridge productions.

documentation/concepts/determinism/index.md
200sc: I always run silver with --clean, or at least have recently, and as long as I've been doing it it hasn't cleared out the generated subdirectory that I'm working with, and this error occurs until I go remove it myself. It's possible that the way I'm running silver uses --clean wrong?

tedinski: OY. I think I recently introduced a bug. I'm going to go fix it.

tedinski: Alright, remove mention of this, it should be fixed in #171

tedinski:
lgtm

PR #170, opened 2017-05-23
dc no longer passes a terminal to the intLiteral abstract production.

remexre:
Also, should I change the left-hand-sides of all the productions to `top`, to fit with what we do in other grammars?

### DIFF
--- a/tutorials/dc/AbstractSyntax.sv
+++ b/tutorials/dc/AbstractSyntax.sv
@@ -57,9 +57,9 @@ quo::Expr ::= l::Expr r::Expr
 }
 
 abstract production integerConstant
-e::Expr ::= i::IntLit_t
+e::Expr ::= i::Integer
 {
- e.pp = i.lexeme ;
- e.value = toInt(i.lexeme);
+ e.pp = toString(i);
+ e.value = i;
 }
 
--- a/tutorials/dc/BetterPP.sv
+++ b/tutorials/dc/BetterPP.sv
@@ -108,7 +108,7 @@ quo::Expr ::= l::Expr r::Expr
 }
 
 aspect production integerConstant
-e::Expr ::= i::IntLit_t
+e::Expr ::= i::Integer
 {
- e.bpp = i.lexeme ;
+ e.bpp = toString(i);
 }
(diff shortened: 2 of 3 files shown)


ericvanwyk:
<rant>No, do not change all lhs to 'top'.  This is not an established
convention and is, frankly, something that I do not like very much.  If
they all have the same name, then why bother giving the lhs a name...</rant>

remexre:
I mean, I'd be fine making `top` a keyword (or keyword-by-convention) like `this` or `self` in most languages... Other than that, does this look good?

ericvanwyk:
I'd say no to 'top' being a keyword.  Besides we have more important things
to worry about.

Otherwise this look good to me - go ahead and merge.

BTW, does this get tested in Jenkins?  Do any of the tutorials?

tedinski:
I should probably write a FAQ entry about `top` someday.

But the short version is that it's exactly like `self` in Python. You can actually name it anything, but writing code without a "we always use a consistent name for this everywhere" rule is _awful_. The choice of `top` is unfortunate, but hey, I get to blame previous students for that! :D 

Eric has (what I think are) good reasons to not use `this`. I actually don't know that we've ever bothered to consider `self` because... you know... not important. Lots of better things to do.

> BTW, does this get tested in Jenkins? Do any of the tutorials?

They should all be, yes, but I haven't double checked that recently.

krame505:
Agreed. It seems ugly, but makes copy/pasting and generally writing code
much simpler.

On May 24, 2017 10:18 AM, "Ted Kaminski" <notifications@github.com> wrote:

I should probably write a FAQ entry about top someday.

But the short version is that it's exactly like self in Python. You can
actually name it anything, but writing code without a "we always use a
consistent name for this everywhere" rule is *awful*. The choice of top is
unfortunate, but hey, I get to blame previous students for that! :D

Eric has (what I think are) good reasons to not use this. I actually don't
know that we've ever bothered to consider self because... you know... not
important. Lots of better things to do.

BTW, does this get tested in Jenkins? Do any of the tutorials?

They should all be, yes, but I haven't double checked that recently.

—
You are receiving this because you are subscribed to this thread.
<https://github.com/melt-umn/silver/pull/170#issuecomment-303757176>, or mute
the thread

PR #177, opened 2017-06-15
Fix/error reporting

### LINKED ISSUE #20: Concrete syntax isn't error checked
```
Provide an explanation of what the project should entail:

The concrete grammar AST (largely under silver:definition:concrete_syntax:ast) 
has the ability to do error checking, but by and large doesn't yet.

Provide a concrete example of a problem this project will fix:

It's possible to produce crashes during parser spec generation time. And it's 
possible to generate silently "corrupted" parser specs.  Try referring to a 
terminal in a production, but with that terminal in a grammar that was not 
included in the parser.

Provide a short guide on how to start this project:

There is already a "cstErrors" attribute on the cs-AST. But the errors checked 
for are WOEFULLY incomplete, and further, this attribute is not even consulted 
before generating parsers!

Provide a rough expectation of what the set of patches completing this
project might look like:

Fleshing out the error checking, with lots of test cases added to the 
copper_features test suite.

A change to the build process to make checking these for error feasible.

Probable challenges in completing this project:

Part of the trouble is figuring out when to do that: the way parsers get 
generated probably has to be changed so that they can be (a) generated (b) 
checked for errors and (c) translated in a 3 phase process.  Currently, they 
are generated ephemerally, only for translation.

Additional Information:

The "long term solution" would be to put semantic checks directly into silver, 
rather than at parser generation time. But, this cannot be done for awhile. In 
the meantime, we should emit good error messages, rather than have the compiler 
crash!
```

Original issue reported on code.google.com by `tux...@gmail.com` on 9 Apr 2012 at 7:52

200sc:
This resolves #20.

Everything outside of the test directory is ready for comments-- I need to work on writing a test file that will successfully parse the other parsers and confirm that they fail, the current one (parsers/Parsers.sv) will call one test and exit as the parser fails.

There's also a couple of things in the tests that should probably be removed as they don't refer to csterrors (the recursive error tests, for example)


grammars/silver/analysis/warnings/defs/Inh.sv
```diff
@@ -571,11 +571,4 @@ top::Expr ::= e::Decorated Expr  q::Decorated QNameAttrOccur
     else [];      
 }
 
-aspect production decorateExprWith
```
tedinski: I'm not sure why this is removed?

grammars/silver/definition/concrete_syntax/ast/CstAst.sv
```diff
@@ -31,20 +31,22 @@ top::SyntaxRoot ::= parsername::String  startnt::String  s::Syntax  terminalPref
...
   
   -- This should be on s1, because the s2 transform assumes everything is well formed.
   -- In particular, it drops productions it can't find an NT for.
   top.cstErrors := s.cstErrors;
   
   production startFound :: [Decorated SyntaxDcl] = searchEnvTree(startnt, s2.cstEnv);
-  -- TODO check if this is found!!
+
+  top.cstErrors <- if null(startFound) then 
+                   ["Missing nonterminal " ++ startnt ++ " for parser "  ++ parsername] else [];
```
tedinski: Presumably the parser name is printed before the list of error messages, so isn't needed here.

grammars/silver/definition/concrete_syntax/ast/CstAst.sv
```diff
@@ -31,20 +31,22 @@ top::SyntaxRoot ::= parsername::String  startnt::String  s::Syntax  terminalPref
...
 
   production univLayout :: String = implode("", map(xmlCopperRef, s2.allIgnoreTerminals));
 
   s2.univLayout = univLayout;
-  top.xmlCopper =
+  top.xmlCopper = if !null(top.cstErrors) then "" else
```
tedinski: I don't think these guards should be here. We should avoid ever accessing `xmlCopper` at all if there are errors. I'd rather these were removed. (In more places that just here, but I'll refrain from commenting on each.)

grammars/silver/definition/concrete_syntax/ast/ProductionModifiers.sv
```diff
@@ -67,8 +67,11 @@ top::SyntaxProductionModifier ::= term::String
 {
   local termRef :: [Decorated SyntaxDcl] = searchEnvTree(term, top.cstEnv);
   
-  top.cstErrors := []; -- TODO check for errors
-  top.productionOperator = just(xmlCopperRef(head(termRef)));
+  top.cstErrors := if !null(termRef) then [] 
+                   else ["Unknown terminal in operator clause " ++ term];
+  top.productionOperator = if null(termRef) then nothing()
```
tedinski: Maybe this guard is okay, but could this be the same situation? i.e. `productionOperator` shouldn't be accessed if there are errors?

grammars/silver/definition/concrete_syntax/ast/ProductionModifiers.sv
```diff
@@ -87,9 +90,34 @@ abstract production prodLayout
 top::SyntaxProductionModifier ::= terms::[String]
 {
   local termRefs :: [[Decorated SyntaxDcl]] = lookupStrings(terms, top.cstEnv);
+  local pairTerms :: [Pair<String [Decorated SyntaxDcl]>] = pairTermRefs(terms, termRefs);
```
tedinski: A possibility to think about (not requirement!): Maybe you could modify `lookupStrings` to just do what you want?

grammars/silver/definition/concrete_syntax/ast/ProductionModifiers.sv
```diff
@@ -87,9 +90,34 @@ abstract production prodLayout
...
+
+  -- TODO: see above, want a util function for this mass head checking, test case
+  top.cstErrors := if null(termRefs) then [] -- layout{} is valid, so this is not an error. 
+                   else foldr(\ a::Pair<String [Decorated SyntaxDcl]> b::[String] ->
```
tedinski: `foldr + lambda` should be a last resort. I think here you'd prefer a `flatMap` (perhaps with that helper function you TODO.)

grammars/silver/definition/concrete_syntax/ast/ProductionModifiers.sv
```diff
@@ -87,9 +90,34 @@ abstract production prodLayout
...
+                     if !null(a.snd) then b
+                     else b ++ ["Unknown terminal in layout clause " ++ a.fst],
+                   [], pairTerms);
 
-  top.cstErrors := []; -- TODO: check for errors
-  top.customLayout = just(implode("", map(xmlCopperRef, map(head, termRefs))));
+  --This causes a concrete syntax error in silver itself 
+  --top.customLayout = if null(termRefs) then nothing() else
+  top.customLayout = just(implode("",
```
tedinski: Here again, not sure, but isn't this doing a lot of guard work for something that's just not accessed if there's errors?

grammars/silver/definition/concrete_syntax/ast/ProductionModifiers.sv
```diff
@@ -87,9 +90,34 @@ abstract production prodLayout
...
+                         [], termRefs))));
+                                
   top.unparses = ["layout(" ++ unparseStrings(terms) ++ ")"];
 }
 
+-- This function is a little useless because it wont report an error if the input
+-- lengths are not equal
+-- This is a transfomration being used to easily reference erroneous strings
+-- within maps/folds checking SyntaxDcls. 
+function pairTermRefs
```
tedinski: This function just does `zipWith(pair, _, _)`.

grammars/silver/definition/concrete_syntax/ast/Syntax.sv
```diff
@@ -312,14 +313,23 @@ top::SyntaxDcl ::= n::String terms::[String] acode::String
...
+  local trefs::[[Decorated SyntaxDcl]] = lookupStrings(terms, top.cstEnv);
+ 
+
+  --local refTermPairs::<[Decorated SyntaxDcl] String> = map(
+  --  \a::[Decorated SyntaxDcl] b::String -> pair(a,b), trefs, terms)
+  
+  -- this 'n' here appears to actually hold the line number of the 
+  -- disambiguation, and the grammar. But we arent supposed to know this?
+  -- TODO: use "indexed" terminal string in print here
+  top.cstErrors := foldr(\a::[Decorated SyntaxDcl] b::[String] ->
```
tedinski: Another place where `flatMap` is probably preferable.

grammars/silver/modification/copper/BuildProcess.sv
```diff
@@ -84,20 +84,32 @@ top::DriverAction ::= spec::ParserSpec  cg::EnvTree<Decorated RootSpec>  silverG
...
+  local specCst :: SyntaxRoot = spec.cstAst;
+
   local ex :: IOVal<Boolean> = isFile(file, top.ioIn);
   local oldSpec :: IOVal<String> = readFile(file, ex.io);
   
   local join :: IO = if ex.iovalue then oldSpec.io else ex.io;
 
+  local err :: IO = 
+    print("CST Errors while Generating Parser " ++ spec.fullName ++ ":\n" ++
+      foldr(\ a::String b::String ->
```
tedinski: Can use `implode` instead of `foldr` here.

grammars/silver/modification/copper_mda/BuildProcess.sv
```diff
@@ -6,88 +6,54 @@ import silver:translation:java:core only makeParserName, makeName;
...
 aspect production compilation
 top::Compilation ::= g::Grammars  _  buildGrammar::String  benv::BuildEnv
 {
-  top.postOps <- [generateMdaSpecs(g.compiledGrammars, grammarsToTranslate, benv.silverGen)]; 
-
-  local targets :: String = mdaBuildTarget(grammarsToTranslate, benv.silverGen);
-  extraTopLevelDecls <- if length(targets) != 0 then ["  <target name='copper_mda'>\n" ++ targets ++ "  </target>\n"] else [];
-  extraGrammarsDeps <- if length(targets) != 0 then ["copper_mda"] else [];
+  top.postOps <-
+    foldr(\ a::Decorated RootSpec b::[DriverAction] ->
```
tedinski: Probably `flatMap((.mdaSpecs), grammarsToTranslate)` then `map` the `mdaBuildSpec`. Avoids the `foldr` and lambda.

grammars/silver/definition/concrete_syntax/ast/CstAst.sv
200sc: Alright, I'll try adding in avoiding referring to copperXml (and others) if there are errors.

grammars/silver/analysis/warnings/defs/Inh.sv
200sc: It wasn't doing anything, there's this and some typos that I took out but could be moved to a different PR considering they aren't related. Do you feel that the empty aspect production with a note that it is not needed should be kept?

grammars/silver/definition/concrete_syntax/ast/CstAst.sv
200sc: That it is, I will remove that second print.

grammars/silver/definition/concrete_syntax/ast/CstAst.sv
tedinski: It should already be in, right? Doesn't your IO code stop writing the file after printing errors?

200sc: It does, but xmlCopper is referred to before writing.

grammars/silver/analysis/warnings/defs/Inh.sv
tedinski: I'm okay with some small unrelated things in pull requests. Too much of a pain to improve little things you stumble across otherwise.

But in this case, I think it's actively documenting *why* nothing is there (and that something could have been there), so it's valuable, imo.

grammars/silver/definition/concrete_syntax/ast/CstAst.sv
tedinski: ...that should read "stop *before* writing the file after printing errors"

grammars/silver/definition/concrete_syntax/ast/ProductionModifiers.sv
200sc: Okay, good, I was hoping there was something that did this in the language already

grammars/silver/definition/concrete_syntax/ast/CstAst.sv
tedinski: I glanced at the file, and I *think* it's fine? (Remember Silver is lazy so just having a variable that does `.xmlCopper` before hand doesn't mean we're accessing it, if that variable isn't used.)

200sc: I only put these in because I was getting a crash for them not being defined, so something in the doWR vs error IO logic needs to change.

grammars/silver/definition/concrete_syntax/ast/ProductionModifiers.sv
200sc: I looked into this, `lookupStrings `is used elsewhere (about 8 uses) so I don't want to change it, but I could have `pairTermRefs `just call `lookupStrings `and do a `zip`

grammars/silver/definition/concrete_syntax/ast/ProductionModifiers.sv
200sc: This guard ended up being unnecessary.

200sc:
That should be everything before getting the test file for the test grammars working, which is looking like it means taking all of the test grammars and writing them as productions directly

200sc:
It's also worth considering changing the 'A's and 'B's in the test grammars to 'Foo's and 'Bars'

grammars/silver/definition/concrete_syntax/ast/ProductionModifiers.sv
```diff
@@ -88,8 +89,13 @@ top::SyntaxProductionModifier ::= terms::[String]
 {
   local termRefs :: [[Decorated SyntaxDcl]] = lookupStrings(terms, top.cstEnv);
 
-  top.cstErrors := []; -- TODO: check for errors
+  -- TODO: a function for this list null checking?
+  top.cstErrors := if null(termRefs) then [] -- layout{} is valid, so this is not an error. 
+                   else flatMap(\ a::Pair<String [Decorated SyntaxDcl]> ->
+                     if !null(a.snd) then []
```
tedinski: I think this `if` is unnecessary, because the flatmap should be over an empty list right, right?

grammars/silver/definition/concrete_syntax/ast/Syntax.sv
```diff
@@ -312,10 +313,16 @@ top::SyntaxDcl ::= n::String terms::[String] acode::String
...
   top.cstDcls = [];
-  top.cstErrors := [];
-  local attribute trefs :: [[Decorated SyntaxDcl]];
-  trefs = lookupStrings(terms, top.cstEnv);
-  -- TODO: check terminal
+
+  local trefs::[[Decorated SyntaxDcl]] = lookupStrings(terms, top.cstEnv);
+  
+  -- this 'n' here appears to actually hold the line number of the 
+  -- disambiguation, and the grammar. But we arent supposed to know this?
```
tedinski: Yah, the "names" of disambiguation groups are made up. Copper wants one, but we don't actually name them in Silver code.

grammars/silver/definition/concrete_syntax/ast/Syntax.sv
```diff
@@ -312,10 +313,16 @@ top::SyntaxDcl ::= n::String terms::[String] acode::String
...
+  top.cstErrors := flatMap(\p ::Pair<String [Decorated SyntaxDcl]> -> 
+      if null(p.snd) then 
+      ["Undefined Terminal " ++ p.fst ++ " in Disambiguation " ++ n]
```
tedinski: I think the error messages might need a final pass: recall that we should strongly suggest that the problem is a missing grammar in the parser with all these kinds of errors. Silver already checked this code and it found that terminal without issue.

maybe `"Terminal ${p.fst} was referenced but this grammar was not included in this parser. (Referenced from disambiguation group ${n})"` or something like that.

tedinski:
> That should be everything before getting the test file for the test grammars working, which is looking like it means taking all of the test grammars and writing them as productions directly

I don't understand, could you explain? What's not working about them? (I haven't tried checking this code out and running it yet...)

> It's also worth considering changing the 'A's and 'B's in the test grammars to 'Foo's and 'Bars'

Not worth it, hah. Made up names are made up names. :)

200sc:
Right now the tests are a group of grammars with their own parsers, so you can run "make -i all" with the associated make file and they should all fail with an error, but actually writing a silver file to confirm that "this parser fails to construct due to an error" is evading me, because attempting to evaluate a parser that  will have an error will exit silver prematurely.

grammars/silver/definition/concrete_syntax/ast/ProductionModifiers.sv
200sc: The `if` at 93 should be unneeded, true. I assume that you're referring to that one and not the one at 95.

tedinski: Oops, yeah, must have misclicked.

200sc:
So this is what the current tests output (trimmed to just hold the error messages):

```
CST Errors while Generating Parser rhs:extendedParser:
Lookup error with parameter test:term_b:B_t of production rhs:rhsMissing

CST Errors while Generating Parser disamb:extendedParser:
Undefined Terminal test:term_b:B_t in Disambiguation disamb:__disam16

CST Errors while Generating Parser startMissing:p2:
Missing start nonterminal test:nonterm_b:B

CST Errors while Generating Parser prodMod:extendedParser:
Unknown terminal in operator clause test:term_b:B_t
Unknown terminal in layout clause test:term_b:B_t

CST Errors while Testing MDA mda:parserWithMissingNt:
Missing start nonterminal test:nonterm_b:B

CST Errors while Generating Parser missingLHS:extendedParser:
Lookup error with LHS nonterminal test:nonterm_b:B

CST Errors while Generating Parser terminalLHS:extendedParser:
LHS of production terminalLHS:terminalLHS is not a nonterminal
```

If we follow the "terminal was referenced but a grammar was not included" pattern that would change these to:

```
CST Errors while Generating Parser rhs:extendedParser:
Terminal test:term_b:B_t was referenced but this grammar was not included in this parser. (Referenced from RHS of rhs:rhsMissing)

CST Errors while Generating Parser disamb:extendedParser:
Terminal test:term_b:B_t was referenced but this grammar was not included in this parser. (Referenced from disambiguation group disamb:__disam16)

CST Errors while Generating Parser startMissing:p2:
Terminal test:nonterm_b:B was referenced but this grammar was not included in this parser. (Referenced as parser's starting nonterminal)

CST Errors while Generating Parser prodMod:extendedParser:
Terminal test:term_b:B_t was referenced but this grammar was not included in this parser. (Referenced from operator clause ???)
Terminal test:term_b:B_t was referenced but this grammar was not included in this parser. (Referenced from layout clause ???)

CST Errors while Testing MDA mda:parserWithMissingNt:
Terminal test:nonterm_b:B was referenced but this grammar was not included in this parser. (Referenced as parser's starting nonterminal)

CST Errors while Generating Parser missingLHS:extendedParser:
Terminal test:nonterm_b:B was referenced but this grammar was not included in this parser. (Referenced from LHS of production ???)

CST Errors while Generating Parser terminalLHS:extendedParser:
LHS of production terminalLHS:terminalLHS is not a nonterminal
```
I think I can add in the production referenced in the second to last error, but I don't think I can find a reference to a given operator or layout clause, and the name of the disambiguation clause, while it has a line number, isn't meaningful otherwise.

200sc:
Also, while there are other CST errors, all of them go unreported as they are masked by other errors before they can be reported. These are almost all duplicate $x errors (nonterminal, terminal, production, etc).

ericvanwyk:
Hi Patrick - thanks these look good.

I think I like the second batch of error messages over the first batch - they seem more informative to me.  Is it difficult to generate these instead?

200sc:
No, they're just as easy to generate, with the exceptions of the issues I noted with disambiguation, operator, and layout identification.

It also could suggest that a symbol refers to a Nonterminal or a Terminal when it actually doesn't, as when we generate the error we know 1) what the type of the symbol should be and 2) that we couldn't identify the type of the symbol. The second error set prints what the type should be.

200sc:
And in case the commit messages didn't make it clear, the tests will now run and pass through test/.runtest, but its worth taking a look at the workaround I had to do to neatly run the failing parsers involving calling silver from within silver

tedinski:
For operator/layout the referenced bit should be about the production they're a part of. Is this hard to determine (it shouldn't be, but I haven't looked at the code in awhile) or is there just some confusion about what these are?

For running the expected-failure tests, instead of the Silver program or the makefile, can you just write a shell script?

```
#!/bin/bash

count=0

silver --clean grammar1 || ((count++))
silver --clean grammar2 || ((count++))
silver --clean grammar3 || ((count++))

test $count == 3
```

tedinski:
In fact, maybe instead of looking for failure, look for the appropriate error string?

```
silver --clean grammar1 | grep "B_t" > /dev/null || ((count++))
```

200sc:
Sure, we could use a shell script that does basically the same thing but maybe in a bit neater fashion. I was trying to do it in a way that the script was a silver file, because that was what I saw in the other test folders, and it shows good practice to write tests in the language that we're testing.

I'm not a fan of checking error strings just in case the error messages change, but just checking that it contains the terminal probably would be okay.

200sc:
For operator/layout, right now their productions just take in strings and I don't think any of their attributes contain information about what production they were on. I'll look into hooking that information in.

tedinski:
IIRC, they occur underneath a production in the tree. If there isn't already, you could probably easily add an autocopy attribute passing the name of the production down to them.

200sc:
I think the only remaining concern is that the disambiguation group name is mangled? There are similar mangled names in the case where one uses `concrete productions ...` instead of a single production, so it might not be something we resolve in this PR?

tedinski:
Right. I wouldn't worry about the name issue for this pull request.

I think I'd still like the makefile/testfile replaced by a shell script. But other than that, I think this looks ready to merge! (Which means congrats, right? First change to Silver and not really even a small one at that!)

PR #179, opened 2017-06-27
Fix/cst errors

200sc:
This adds in more cstErrors and tests that were missed in #177, relating to dominates, submits and lexer class clauses.  

One thing that could be considered would be passing in some sort of name to use for `lexer class` declarations. Right now it just says `(Referenced from dominates clause for lexer class)` with no identifier, but this could go on the same back burner as improving the naming for `concrete productions` and `disambiguate` blocks.

### DIFF
--- a/grammars/core/Location.sv
+++ b/grammars/core/Location.sv
@@ -89,10 +89,6 @@ Location ::=
 function locationLte
 Boolean ::= l1::Location l2::Location
 {
-  -- TODO: We could probaly just compare based on filename and index
-  -- For the moment, though, use line & column instead.
-  return l1.filename < l2.filename || (l1.filename == l2.filename &&
-    (l1.line < l2.line || (l1.line == l2.line &&
-    (l1.column < l2.column))));
+  return l1.filename < l2.filename || (l1.filename == l2.filename && (l1.index < l2.index));
 }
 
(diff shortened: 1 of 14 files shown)


grammars/core/Location.sv
```diff
@@ -89,10 +89,6 @@ Location ::=
 function locationLte
 Boolean ::= l1::Location l2::Location
 {
-  -- TODO: We could probaly just compare based on filename and index
-  -- For the moment, though, use line & column instead.
-  return l1.filename < l2.filename || (l1.filename == l2.filename &&
-    (l1.line < l2.line || (l1.line == l2.line &&
-    (l1.column < l2.column))));
+  return l1.filename < l2.filename || (l1.filename == l2.filename && (l1.index < l2.index));
```
200sc: I assume something in our tests would break if this didn't work, but if there isn't a test that would cover this than I don't mind reverting this change. I also emulated the previous behavior, but the name of this function implies it is less than or equal to when it looks like the old logic would just return true for less than?

tedinski: There probably are no tests for this!

I think I'd prefer to stick to using line/col (there may be older code that still doesn't use indexes? Indexes and "end" locations were a relatively recent addition to locations. I think I only fixed Silver to store locations in interface files properly last month?), but good catch on it being broken and doing `<` only.

If you want to, you could fix that and add some tests (probably a `tests/stdlib/Location.sv` file) to make sure we're actually doing the "and equal" part?

200sc: I'll revert it for now, and put the tests / fix in a new branch.

PR #180, opened 2017-06-29
Fix/disambiguateRefactor

200sc:
This resolves a few TODOs related to disambiguate groups. 

Previously you could put a lexer class into a disambiguate group. This would result in a rather unhelpful copper error message with line numbers meaningless to the source .sv file:

```
   [copper] Parser_disambLexer_p.copper:58.70: cvc-complex-type.2.4.a: Invalid content was found starting with element 'TerminalClassRef'. One of '{"http://melt.cs.umn.edu/copper/xmlns":TerminalRef}' is expected.
   [copper] Parser_disambLexer_p.copper:57.54 [parser disambLexer_p]: Disambiguation function disambLexer___disam20 is missing the following required attributes: [code or disambiguateTo, members]
   [copper] Parser_disambLexer_p.copper:7.73 [parser disambLexer_p]: Undefined reference to disambLexer_Root
```

This PR creates a new nonterminal `TermList` which replaces the previous `TermPrecList` in disambiguate groups. `TermPrecList` no longer has a `defs` attribute, as that was only there for the benefit of disambiguate groups. 

The above error message now looks like this, in an example where `A_l` is a defined lexer class:
`Artifact.sv:20:13: error: Undeclared terminal 'A_l'.`

Because `A_l` is declared, this error message could be improved to something like: `error: Expected terminal, found lexer class 'A_l'.`, but that sort of error message would be helpful in a lot of places in Silver so that might be for a different issue.

### DIFF
--- a/grammars/silver/modification/copper/DisambiguationGroup.sv
+++ b/grammars/silver/modification/copper/DisambiguationGroup.sv
@@ -7,8 +7,7 @@ top::AGDcl ::= 'disambiguate' terms::TermList acode::ActionCode_c
 {
   top.pp = "disambiguate " ++ terms.pp ++ " " ++ acode.pp;
 
-  top.errors := terms.errors;
-  top.errors <- acode.errors;
+  top.errors := terms.errors ++ acode.errors;
 
   acode.env = newScopeEnv(disambigLexemeDef(top.grammarName, top.location) ::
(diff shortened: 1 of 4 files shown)


tedinski:
lgtm.

Lexer classes are a bit strange, because they get their own private namespace. Fixing that would require some design work...

PR #181, opened 2017-07-06
Formalize current behavior in inferring the "refset" in flow analysis.

### LINKED ISSUE #178: Finalize behavior of inferred "ref set" for flow analysis
The normal rules of the flow analysis are that all inherited attribute occurrences exported by the nonterminal's grammar are in the ref set. (The set of inherited attributes required/assumed of decorated references.)

This rule (from my PhD thesis) does not account for `option`s. Normally, options are considered exports for everything flow analysis related. However, this might be a good exception: doing so here would require the use of inherited attribute from an "optional component" essentially making it non-optional.

Right now, the code is actually slightly broken!

https://github.com/melt-umn/silver/blob/develop/grammars/silver/definition/flow/env/NonterminalDcl.sv#L11

We're filtering by `isExportedBy` but we're starting with only the inherited attributes know to exist by the local environment. Normally options aren't imported, so we're accidentally skipping any exported by options.

I propose we codify this bug into a feature. We should introduce an `isStrictlyExportedBy` to actually implement the right check (not counting options), and use it here. And also document things.

This change inspired by https://github.com/melt-umn/ableC/pull/50

tedinski:
Pretty much as discussed in the issue.

Fixes #178

### DIFF
--- a/grammars/silver/definition/flow/env/NonterminalDcl.sv
+++ b/grammars/silver/definition/flow/env/NonterminalDcl.sv
@@ -1,14 +1,17 @@
 grammar silver:definition:flow:env;
 
 import silver:definition:type:syntax only BracketedOptTypeList;
-import silver:driver:util only isExportedBy;
+import silver:driver:util only isStrictlyExportedBy;
 
 aspect production nonterminalDcl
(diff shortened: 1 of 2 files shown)


krame505:
LGTM

PR #182, opened 2017-07-06
Check flow types of attribute sections.

### LINKED ISSUE #47: Missing flow type checks on attribute sections (.attr)
```
1. map((.typerep), list :: [Expr]), where ft(typerep) = {env}
2. No errors or warnings
3. Crash!

We need to add checks for this.

```

Original issue reported on code.google.com by `tux...@gmail.com` on 18 Apr 2014 at 6:47

tedinski:
As mentioned in the associated issue.

Fixes #47

### DIFF
--- a/grammars/silver/analysis/warnings/defs/Inh.sv
+++ b/grammars/silver/analysis/warnings/defs/Inh.sv
@@ -578,3 +578,30 @@ top::Expr ::= 'decorate' e::Expr 'with' '{' inh::ExprInhs '}'
 }
 
 -- TODO: pattern variable accesses.
+
+aspect production attributeSection
+top::Expr ::= '(' '.' q::QName ')'
+{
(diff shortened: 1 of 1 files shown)


tedinski:
Merging this.

cc @ericvanwyk @krame505 it's possible new MWDA errors may appear as a result, FYI. I waited until after our artifact was (presumably?) submitted to make sure I didn't do this at the last minute... hope there aren't any other immediate deadlines like that.

ericvanwyk:
The artifact is in, a bit more limited in explanations of things that I'd wanted but I think it is good.

There are no pending deadlines, except final versions of the OOPSLA paper.  But we should be looking for failing MWDA tests and fixing them.  @TravisCarlson can you watch this as well, when you return - no need to worry about this now while you're away.

PR #184, opened 2017-07-09
New warning: lower-case production names. Refactor regex.

### LINKED ISSUE #46: Silver's separate namespaces for productions and nonterminals can clash in Copper specs
```

Name a nonterminal and a production the same name. e.g.

nonterminal A;

concrete production A
top::A ::=
{}

and generate a parser. 

What is the expected output? What do you see instead?

Copper will raise errors due to the duplicate names 'A', while Silver is happy 
with this arrangement.

Actually, the bug may manifest itself as Copper complaining about mal-formed 
XML, since Silver will happily resolve the name to a production inside, say, 
the right-hand side list of symbols (which should only be terminal or 
nonterminal references)

Please use labels and text to provide additional information.

We had discussed a possibility of required productions to be lower case awhile 
back. That would solve the problem, but is also a breaking change (which is why 
we haven't done it yet.)  Another possibility would be to mangle the copper 
names a little bit more than they are now by adding a namespace prefix like 
"prod__" to productions.
```

Original issue reported on code.google.com by `tux...@gmail.com` on 11 Mar 2014 at 5:42

tedinski:
I probably should have titled this "Sunday afternoon yac-shaving" :/

@ericvanwyk : Thoughts on introducing the warning in the first commit? Raising a warning on productions that start with anything but a lower-case letter?

Presently, this raised no warnings in Silver (except regex, which I then fixed) and none in ableC either. I think this is a pretty universally observed convention, and enforcing it would fix #46.

...then I decided to fix the regex grammar, and it made me angry enough that I did a complete refactoring of it, giving everything nice names, and fixing kludges.

...then I fixed a bug where silver was trying to continue compiling in the presence of parse errors. No one reported this? :(

...then I decided maybe we were printing slightly long lines listing filenames, and reduced that.

### DIFF
--- a/grammars/silver/definition/concrete_syntax/TerminalDcl.sv
+++ b/grammars/silver/definition/concrete_syntax/TerminalDcl.sv
@@ -62,10 +62,10 @@ top::AGDcl ::= t::TerminalKeywordModifier 'terminal' id::Name r::RegExpr tm::Ter
  -}
 nonterminal RegExpr with config, location, grammarName, pp, terminalRegExprSpec;
 
-synthesized attribute terminalRegExprSpec :: Regex_R;
+synthesized attribute terminalRegExprSpec :: Regex;
 
 concrete production regExpr
-top::RegExpr ::= '/' r::Regex_R '/'
+top::RegExpr ::= '/' r::Regex '/'
 {
   top.pp = "/" ++ r.regString ++ "/";
   top.terminalRegExprSpec = r;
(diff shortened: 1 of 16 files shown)


ericvanwyk:
Ok, you have been busy, definitely shaving those yacs.

I say go ahead with this, they all make sense.

I did notice Silver continuing after parse errors but was neck deep in the artifact stuff.  Fixing this is good.

PR #190, opened 2017-07-20
Major refactoring to our scripts

tedinski:
* In an attempt to be nicer to new users (and also just cleaning up), reduce the number of scripts in the project root. Rarely used scripts are now under `support/scripts/`
* Also makes these scripts somewhat more uniform in style and error checking.
* Slightly improves the install scripts. Suggests also installing gedit/emacs/vim support.
* Installs scripts like `sv-grep` by default now.

I don't intend to merge this until next week, because I think this will cause some breakages:

* [ ] Jenkins probably needs updating on `fetch-jars` and `make-dist` locations
* [ ] `generate-documentation` moved. What script uses this?
* [ ] Our AbleC VM artifact needs updating to find the install script in a new location
* [ ] For that matter, the documentation on our website needs updating!
* [ ] ...Also the README.md.

Anything else? Comments @remexre @ericvanwyk @krame505 @TravisCarlson ?

I'd like to merge this before a 0.4.0 release.


krame505:
Do we really want to move all these away from the top level?  I understand moving scripts that are called by update and the publishing scripts (which makes it much nicer for users of silver who aren't developing things), but one that I do use quite regularly is `./make-ide`, which seems like one we may want to keep at the top-level.  I believe this script is also mentioned in documentation in a few places, so we should probably either fix that or just leave it alone for now.  

Also, we have started moving in the direction of using Makefiles at the top-level of extensions for organizing various build and test commands.  For consistency, is this something that we may want to consider for ableC and Silver?

tedinski:
I'm 100% disinterested in using makefiles for Silver or Ablec. I didn't object to them for extensions in part because extensions have to also work with C code/libraries, in part because extensions have multiple "pieces", and in part because I didn't have to deal with it. :)

(In my general opinion, Makefiles are twice as much "awful 70s cruft" as bash scripts, but more complicated, harder to understand, and people are generally a lot less familiar with them. You use them because you have to, not because they're good. For Silver code, the Silver compiler ***is*** the build system, not just the compiler, so we don't need a separate build system like Makefiles, imo.)  ((PS: Also.... for AbleC, ideally we'd eventually get to the point where Silver is smart enough that all you need to build it is `silver build` in the repo, or something similarly "get rid of scripts entirely".))

RE `make-ide`: Would your concern be alleviated if we got Jenkins to auto-update an actual eclipse update site on `melt.cs.umn.edu`, so you generally never have to run it yourself?

krame505:
Sounds reasonable.

If setting up an actual update site would be simple then that would be
great- although this IDE is going away eventually, that is probably long
enough down the road for this to still be useful.

krame505:
Is this PR still planned for 0.4.0?  Or has it been abandoned for now?

tedinski:
Since we're running out of time, I think I'll merge it after 0.4.

Unless something shows up that we need to make an AbleC release, my plan is for Silver as-is to be 0.4 (with some release process stuff done, like getting HISTORY up to date.)

PR #191, opened 2017-08-03
Allow using 'decorate' in flow types

### LINKED ISSUE #187: Allow the reference set to be used in specifying flowtypes
One pattern that I see frequently when adding explicit flowtypes to ableC is that there are many places where the flowtype of an attribute on a nonterminal is the same as that nonterminal's reference set, which may be different between nonterminals.  Thus it would be nice to be able to simply write something like
```flowtype Expr = errors {decorate}, defs {decorate}, ...;```
or even
```flowtype errors {decorate} on Expr, Stmt, Decl, Decls, ...;```
instead of needing to separately handle the nonterminals with extra inherited attributes in the reference set, which has become rather tedious.  
@tedinski what do you think of this?  How difficult would this be to implement?

tedinski:
@krame505 could you review this please? Feel free to merge yourself if you decide it's okay. My internet connectivity is limited at the moment.

After looking into it, it turns out this is simple as long as we require an explicit specification for 'decorate'. I've decided to be conservative for the moment, and just do 'decorate'.

Fixes #187

### DIFF
--- a/grammars/silver/definition/flow/syntax/FlowSpec.sv
+++ b/grammars/silver/definition/flow/syntax/FlowSpec.sv
@@ -88,8 +88,10 @@ top::FlowSpec ::= attr::FlowSpecId  '{' inhs::FlowSpecInhs '}'
     then []
     else [err(attr.location, attr.pp ++ " is an extension synthesized attribute, and must contain at least the forward flow type. It is missing " ++ implode(", ", missingFt))];
   
+  -- We want to put the spec in even if there are errors in 'inhs' so that
+  -- we can look up specs from inhs.
   top.flowDefs = 
-    if !null(attr.errors) || !null(inhs.errors) then []
+    if !null(attr.errors) then []
     else [specificationFlowDef(top.onNt.typeName, attr.synName, inhs.inhList)];
 }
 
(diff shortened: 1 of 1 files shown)


krame505:
Thanks, this is for sure better than what we had before, although we may
want to think about what it would take to do this the 'right' way and
actually infer these.
I'll go ahead and merge this now, but I probably won't get be able to try
this out with ableC until next weekend since this weekend I am traveling.

On Aug 3, 2017 10:27 AM, "Ted Kaminski" <notifications@github.com> wrote:

@krame505 <https://github.com/krame505> could you review this please? Feel
free to merge yourself if you decide it's okay. My internet connectivity is
limited at the moment.

After looking into it, it turns out this is simple as long as we require an
explicit specification for 'decorate'. I've decided to be conservative for
the moment, and just do 'decorate'.

Fixes #187 <https://github.com/melt-umn/silver/issues/187>
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/melt-umn/silver/pull/191
Commit Summary

   - Allow using 'decorate' in flow types

File Changes

   - *M* grammars/silver/definition/flow/syntax/FlowSpec.sv
   <https://github.com/melt-umn/silver/pull/191/files#diff-0> (39)

Patch Links:

   - https://github.com/melt-umn/silver/pull/191.patch
   - https://github.com/melt-umn/silver/pull/191.diff

—
You are receiving this because you were mentioned.
<https://github.com/melt-umn/silver/pull/191>, or mute the thread

krame505:
Forgot to actually hit 'merge'...

PR #192, opened 2017-08-20
Improve error message for flowspecs to say what nonterminal is causin…

krame505:
Minor fix, actually say what nonterminal has its ref set undefined when using `decorate` in a flowtype, so we can tell what to fix when writing, e.g.
```
flowtype host {decorate} on
  Expr, Stmt, Decls, Decl, TypeName, BaseTypeExpr, TypeModifierExpr ...;
```
We should also probably fix the locations somehow to actually highlight the nonterminal that is causing the problem (similar for some other errors related to flowspecs), but I couldn't figure out an easy way of doing that because of how the grammar is structured.


tedinski:
Lgtm

PR #199, opened 2017-10-02
Flow analysis on pattern matching

tedinski:
This is probably the only thing that should block a Silver 0.4 release.

This pull request:

1. Introduces "projection stitch points" as described in my thesis. The data type was already there, but we weren't yet generating them as FlowDefs.
2. Checks for pattern matching on references for transitive dependencies outside the "ref set" of that reference.
3. Checks for dependencies involving pattern variables if the appropriate inherited equation exist in the remote productions.
4. Fixes a bug related to flow graphs for functions: I forgot to add in the "stitch points" to the graph, so it was incomplete.

In base AbleC it exposed a few new minor flow errors reported here: https://github.com/melt-umn/ableC/issues/70

I have not yet tried the AbleC extensions.

@ericvanwyk  @krame505 @TravisCarlson Is it okay if I merge this, or would it upset anything if suddenly there's possibly more flow errors in ablec/extensions? I think it'd be a good idea to get this in ASAP, so we can do a release and get things fixed...

### DIFF
--- a/grammars/silver/analysis/warnings/defs/Inh.sv
+++ b/grammars/silver/analysis/warnings/defs/Inh.sv
@@ -2,6 +2,8 @@ grammar silver:analysis:warnings:defs;
 
 import silver:modification:autocopyattr only isAutocopy;
 import silver:modification:collection;
+import silver:modification:primitivepattern;
+
 import silver:definition:flow:driver only ProductionGraph, FlowType, flowVertexEq, prod, inhDepsForSyn, findProductionGraph, expandGraph, onlyLhsInh;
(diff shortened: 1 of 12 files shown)


krame505:
I guess I don't have an issue with merging now, although I haven't gotten around to patching ableC yet so it might end up showing as failing on Jenkins.  Not a big deal I guess?

ericvanwyk:
Merging seems good to me.  If it breaks things we can then fix them...  I
prefer to merge soon so that we have time to fix it all.

Thanks for all of this Ted - looks good!

PR #201, opened 2017-10-04
Avoid computing flow types without MWDA

tedinski:
This quick fix checks to see if we're doing the MWDA before enforcing the rule for `flowtype` declarations requiring extension synthesized attributes to have flow types that are a super set of the forward flow type.

Also, increase memory for the cstast test, which failed due to lack of memory because of this, and because it builds a lot of the silver compiler along with itself.


tedinski:
Just merging, since I want to get jenkins green here, but of course feel free to yell at me here if anything looks off.

PR #204, opened 2017-10-06
Manually set the sv-nailgun stack size to be the same as silver

### LINKED ISSUE #203: sv-serve has no way of setting the stack size
We basically don't have a way of specifying the stack (or, for that matter, the heap) size when launching `sv-serve`.  The easiest fix is probably to just bump up the defaults in `sv-serve`...
This is a blocking issue for https://github.com/melt-umn/ableC/pull/68, since the increased stack usage for the generated builtins is causing some of the ableC tests to crash.  @tedinski comments?

krame505:
Fixes #203.

### DIFF
--- a/support/nailgun/sv-nailgun
+++ b/support/nailgun/sv-nailgun
@@ -56,7 +56,7 @@ function sv-serve {
     return 4
   fi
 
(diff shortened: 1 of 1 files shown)


krame505:
I'll go ahead and merge this...

PR #208, opened 2017-10-31
Monto3 Builtins for Silver

remexre:
This was going to be the PR for TerminalDescriptors, but I thought that the Monto2 stuff didn't get merged; apparently it did.

Anyway, this has the Silver+runtime code for Monto3, and drops the JeroMQ dependency brought on my Monto2 as a result (GSON is still required though). If we don't want to merge the Silver parts, that's fine, I just want to get the runtime bits merged.


grammars/silver/support/monto/Service.sv
```diff
@@ -0,0 +1,52 @@
...
+    end;
+}
+
+function asDescriptor
+ProductDescriptor ::= pi::ProductIdentifier
+{
+  return productDescriptor(pi.productName, pi.productLang);
+}
+
+-- TODO: Move this to core.
```
remexre: Also, would this be appropriate to do now? I couldn't find anything like this in the existing `core`.

tedinski:
Ok. I haven't reviewed the monto library/runtime because I don't know it and I figure you know what you're doing there. Thoughts on the rest:

1. is `jsonInteger` common? Properly speaking, I don't think JSON has anything but floats, but if it's typical to recognize integers and treat them specially, then ok.
2. I'm okay with moving `find` to core, if you're willing to watch jenkins after the merge and make sure it doesn't break anything. Presently, adding things to core is often dangerous. :/
3. What's with the change to the install script? I think this might conflict with changes I have in another branch, and I'm not sure I like it (it changes some messages to become false if a parameter is supplied, and I don't know why we'd want to put things any place else.)
4. Can you audit and be specific about what files should be deleted from `jars/`? We don't presently have a good system for dealing with removing obsolete files, so we might have to add code to delete them! heh (I'll handle this later, I just want to know what needs removing.)

remexre:
> is jsonInteger common? Properly speaking, I don't think JSON has anything but floats, but if it's typical to recognize integers and treat them specially, then ok.

It's more a convenience for generating JSON than anything else; if it should be a function/forwarding production, I can change it.

> I'm okay with moving find to core, if you're willing to watch jenkins after the merge and make sure it doesn't break anything. Presently, adding things to core is often dangerous. :/

Okay, I'll do that now.

> What's with the change to the install script? I think this might conflict with changes I have in another branch, and I'm not sure I like it (it changes some messages to become false if a parameter is supplied, and I don't know why we'd want to put things any place else.)

I can get rid of that commit; my memory's a bit fuzzy on that, but I believe I was wanting to install the `silver` symlink to `~/.local/bin` [per XDG](https://www.freedesktop.org/software/systemd/man/file-hierarchy.html#~/.local/bin). I don't remember why (or if) that caused any real issue though.

> Can you audit and be specific about what files should be deleted from jars/? We don't presently have a good system for dealing with removing obsolete files, so we might have to add code to delete them! heh (I'll handle this later, I just want to know what needs removing.)

I believe the list is:

 - `ahocorasick-0.3.0-javadoc.jar`
 - `ahocorasick-0.3.0-sources.jar`
 - `ahocorasick-0.3.0.jar`
 - `jeromq-0.3.4-javadoc.jar`
 - `jeromq-0.3.4-sources.jar`
 - `jeromq-0.3.4.jar`
 - `json-simple-1.1.1.jar`

Although not everyone has these; my desktop at home doesn't have the Aho-Corasick stuff, for example.

tedinski:
> It's more a convenience for generating JSON than anything else; if it should be a function/forwarding production, I can change it.

If it's easy to change to a function right now, then maybe we should change it. Is there no code right now that constructs that type from JSON anywhere? Does it get used?

> I can get rid of that commit

Okay. Good to know about that XDG spec. I'll file that away in the back of my mind, we might directly support that someday...

> I believe the list is:

Just to double check I'm reading right, these are files we can delete, right? (I don't have a repo handy right now, does this means it's just gson we're keeping, of the third-party jars?)

remexre:
> If it's easy to change to a function right now, then maybe we should change it. Is there no code right now that constructs that type from JSON anywhere? Does it get used?

`silver:json` doesn't actually include a JSON parser; I do all the parsing with GSON in the runtime. I'll change it to be a function.

> Just to double check I'm reading right, these are files we can delete, right? (I don't have a repo handy right now, does this means it's just gson we're keeping, of the third-party jars?)

That sounds correct, yes.

remexre:
Did the `jsonInteger` change, rebased to remove the changes to `install-silver-bin`.

tedinski:
Oh, sorry for late response. Looks good.

PR #210, opened 2017-11-19
fixes #209, incorrect type in lambda translation

### LINKED ISSUE #209: Case within Lambda causes java translation error
This code, (with a minimal parser), produces a java file with errors after translation

```
nonterminal Foo;

abstract production bar
f::Foo ::=
{}

function fails
(String ::= Foo) ::=
{
    return (\ f::Foo -> case f of bar() -> "str" end);
}
```

The error for this specific example is 

```
    [javac] /mnt/c/linux_work/silver/generated/src/translateBug/Pfails.java:83: error: incompatible types: <anonymous NodeFactory<Object>> cannot be converted to NodeFactory<StringCatter>
    [javac]             return (common.NodeFactory<common.StringCatter>)((new common.NodeFactory<Object>() {
    [javac]                                                             ^
    [javac] 1 error
```

I don't have much insight into what would be causing it, but I do know some changes that stop the error from happening: 1) If there isn't a case statement, it succeeds, and 2) if `bar()` is replaced with `_`, it succeeds. 

I don't suspect the types involved effect the presence of the error, but this first happened with me in a similar function that operated on `QName`s and `DclInfo`s.

tedinski:
Use post-inference types in the translation of lambdas. Fixes bug where Object gets used instead of a more accurate type, and Java generics are invariant so `Foo<A>` is not compatible with `Foo<Object>`.

Fixes #209 

Hey @200sc can you glance at this? If it looks okay to you, we'll merge.

### DIFF
--- a/grammars/silver/modification/lambda_fn/java/Lambda.sv
+++ b/grammars/silver/modification/lambda_fn/java/Lambda.sv
@@ -18,8 +18,8 @@ top::Expr ::= params::ProductionRHS e::Expr
   local finTy :: Type = finalType(top);
   
   top.translation = 
-s"""(new common.NodeFactory<${e.typerep.transType}>() {
-  public final ${e.typerep.transType} invoke(final Object[] args, final Object[] namedArgs) {
+s"""(new common.NodeFactory<${finTy.outputType.transType}>() {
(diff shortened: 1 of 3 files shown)


200sc:
Checks out on the grammar I was working with. LGTM

Issue #212, opened 2017-12-15
General-purpose AST reflection


krame505:
## Overview
I am proposing that we add a general-purpouse AST reflection library to Silver.  This would be useful for implementing embedded host-language AST literals (#194), and may find other uses as well.  

Essentially, I am proposing something sort of like `hackUnparse`, except that it outputs a generic tree structure instead of a string.  

```
closed nonterminal AST with pp;

abstract production nonterminalAST
top::AST ::= prodName::String children::ASTs annotations::NamedASTs
{}

abstract production listAST
top::AST ::= vals::ASTs
{}

abstract production stringAST
top::AST ::= s::String
{}

-- Etc...

nonterminal ASTs with pp;
nonterminal NamedASTs with pp;

-- All the expected cons and nil productions here

nonterminal NamedAST with pp;

abstract production namedAST
top::NamedAST ::= n::String v::AST
{}
```

It would also be possible to designate certain types occuring in the tree that should not be reflected, but instead placed directly in the result tree.  

```
nonterminal ASTProd;
abstract production astProd
top::ASTProd ::= prod::(AST ::= a)
{}

-- Foreign function
function reflectAST
AST ::= escapes::[Pair<String ASTProd>] value::a
{}
```

`escapes` here is a list of pairs of the names of types to escape from reflection and wrapping bridge productions for them.  Note that the `ASTProd` nonterminal is effectively like a `newtype` in Haskell, since the productions in `escapes` vary in type.   

## Example
```
grammar example:abstractsyntax;

nonterminal Expr with ...;

abstract production fooExpr
top::Expr ::= a::Expr b::Foo c::Decorated Bar
{}

abstract production bazExpr
top::Expr ::= prod::(Expr ::= Expr)
{}

nonterminal Foo with ...;

nonterminal Bar with ...;

-- A bunch more productions here

abstract production fooAST
top::AST ::= f::Foo
{}

abstract production decBarAST
top::AST ::= b::Decorated Bar
{}

abstract production unaryProdAST
top::AST ::= prod::(Expr ::= Expr)
{}

-- A production body somewhere
local ast::Expr = ...;
local reflectedAST::AST =
  reflectAST(
    [pair("example:abstractsyntax:Foo", astProd(fooAST)),
     pair("Decorated example:abstractsyntax:Bar", astProd(decBarAST)),
     pair("(example:abstractsyntax:Expr ::= example:abstractsyntax:Expr)", astProd(unaryProdAST)],
    ast);
```

[...]

PR #213, opened 2018-01-04
Implement AST reflection

krame505:
This implements the `reflectAST` operation as described in #212.  Next up is reify.

### DIFF
--- a/runtime/java/src/common/Node.java
+++ b/runtime/java/src/common/Node.java
@@ -96,6 +96,21 @@ public abstract class Node {
 	 */
 	public abstract String getNameOfInhAttr(final int index);
 	
+	/**
+	 * Used for reflection.
+	 * 
+	 * @return An array of the names of all annotations.
+	 */
+	public abstract String[] getAnnoNames();
+
+	/**
+	 * Used for debugging, stack traces especially.
+	 * 
+	 * @param name The name of any annotation returned by <code>getAnnoNames()</code>
+	 * @return The annotation object. (Thunk already evaluated)
+	 */
+	public abstract Object getAnno(final String name);
+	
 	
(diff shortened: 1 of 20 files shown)


krame505:
@tedinski @ericvanwyk Anything that I should fix before merging this?

runtime/java/src/common/Node.java
```diff
@@ -96,6 +96,21 @@ public DecoratedNode decorate() {
...
 	
+	/**
+	 * Used for reflection.
+	 * 
+	 * @return An array of the names of all annotations.
+	 */
+	public abstract String[] getAnnoNames();
+
+	/**
+	 * Used for debugging, stack traces especially.
```
tedinski: Text from copy and paste?

tedinski:
General thoughts:

1. Not sure if we should have AST have a `pp`, but for now okay no problem. But I think we could probably avoid the dependency on langutil in the final version?
2. In the end, we probably need tests that do better than compare to string (too brittle?), but obviously I think that's hard to do yet? So ok.

Probably good to merge.

ericvanwyk:
Ok with me.

runtime/java/src/common/Node.java
krame505: Oops...

krame505:
> 1. Not sure if we should have AST have a pp, but for now okay no problem. But I think we could probably avoid the dependency on langutil in the final version?

Yeah, I suppose those equations could move to aspects in the langutil library.  

> 2. In the end, we probably need tests that do better than compare to string (too brittle?), but obviously I think that's hard to do yet? So ok.

Agreed, we really need a derivable equality typeclass for this...
I'll go ahead and merge.

PR #214, opened 2018-01-18
Implement AST reification

krame505:
This implements the reify operation essentially as described in #212.  
I'm opening this now since this feature is basically complete, although before this is merged more documentation, testing, and (probably) bug fixes are needed.  There are also additional features that this library should possess as well (e.g. `AST` automatic serialization and deserialization), which I will implement in the feature/reflection branch and a future pull request.


tedinski:
* This may have an easier answer but I just wanted to make sure you thought about it: if type variables are unified with mutation, are you sure you're not "permanently" mutating one that you expect to re-use later or something like that?
* skolem constant and unification variables are different. EZ trick might be to make skolem constants just a `BaseTypeRep` with a special name. (This trick might fail to work depending on fancy type features, but I *THINK* those won't interact with reification? They only show up when pattern matching on GADTs afaik)
* I think a separate `ListTypeRep` might be unnecessary, it can also be just another `BaseTypeRep`. e.g. just a static function on `BaseTypeRep` that constructs list types in a standard way. Maybe there's a reason this is easier though, idk.
* I feel like there's more generated code that I would have preferred but maybe I'm wrong, it's just a feeling, I don't see anything wrong. Just saying: if it can be in the runtime instead of generated, that's probably a good idea.
* I don't like the use of `List` `Set` `Map` here, try to stay arrays (which is what you start out with!) Also, I believe annotations maintain an alphabetical order invariant, so it should be okay to do in-order comparison there, if you were worried about that.
* `getType` on list seems busted. Remember: you don't need to type check silver types. Silver does that. The type of a list is `list-type-rep(head-element.get-type)` OR `list-type-rep(fresh-var)` for empty lists. No need to traverse checking each element. (Checking each element's type is what we should be doing when constructing a list from a `listAST` or whatever its called, not trying to get a type for a list.)

That's my brief look-over.

krame505:
> * This may have an easier answer but I just wanted to make sure you thought about it: if type variables are unified with mutation, are you sure you're not "permanently" mutating one that you expect to re-use later or something like that?

Yeah, basically the approach is that I'm taking is that every time you call `getType` you get back an object containing all fresh type variables.  So I did take this into account.   

> * skolem constant and unification variables are different. EZ trick might be to make skolem constants just a `BaseTypeRep` with a special name. (This trick might fail to work depending on fancy type features, but I *THINK* those won't interact with reification? They only show up when pattern matching on GADTs afaik)

I think I understand the type theory here, but I was a bit confused by how this works in Silver, because it seems like skolem constants act like flexible variables in some situations?  For example if I have `production foo :: (Foo<a> ::= a)`, and I call hackUnparse on its signature I get something only involving `skolemType`, even though `a` should be flexible.  
But I'm pretty sure the way I am doing this is actually valid.  Essentially, `ty.unify` has two parameters: the other type to unify with, and a flag of whether to allow substitutions to be performed on the other type.  This is only true when performing unification for the LHS type, when we do want to make these vars flexible, but is false anywhere else.  AFAIK there isn't a way for rigid type variables to occur on the LHS of a production, or flexible var on the RHS of a production.  I have worked through a number of complex examples and haven't found a case where this doesn't work... let me know if you have a specific example.  
Also that trick with BaseTypeExpr wouldn't quite if existential type vars are skolems within the context of a particular production signature.  When these are passed down recursively as the expected type for children to reify into, we want to update the var with whatever the actual type of that child is.  That is how I ensure that existential type vars used in multiple places in a signature actually refer to the same thing.  

> * I think a separate `ListTypeRep` might be unnecessary, it can also be just another `BaseTypeRep`. e.g. just a static function on `BaseTypeRep` that constructs list types in a standard way. Maybe there's a reason this is easier though, idk.

Basicly, the `toString` is different, and we need to handle lists as a special case anyway, so everything just got simpler when I moved it into a seperate class.  

> * I feel like there's more generated code that I would have preferred but maybe I'm wrong, it's just a feeling, I don't see anything wrong. Just saying: if it can be in the runtime instead of generated, that's probably a good idea.

The only big things are the reify function literal, and the `getType` and `reify` functions on every production class, all of which require generating code based on the type in some way.  There isn't anything really that can easily move.  

> * I don't like the use of `List` `Set` `Map` here, try to stay arrays (which is what you start out with!) Also, I believe annotations maintain an alphabetical order invariant, so it should be okay to do in-order comparison there, if you were worried about that.

Any big reason for this?  The main place that I use them is when unpacking the child and annotation `ASTs` in `Reflection.reify`, where I don't know how big to make the arrays initially.  I suppose I could avoid constructing a `Set` in the generated `Pwhatever.reify` function, and I suppose I could make this function take arrays as parameters, knowing that these are sorted.  But is there a problem with using them internally to `Reflection.reify`?  

> * `getType` on list seems busted. Remember: you don't need to type check silver types. Silver does that. The type of a list is `list-type-rep(head-element.get-type)` OR `list-type-rep(fresh-var)` for empty lists. No need to traverse checking each element. (Checking each element's type is what we should be doing when constructing a list from a `listAST` or whatever its called, not trying to get a type for a list.)

The issue is you may have flexible type vars embedded in the first element that need to get further constrained - for example, what about `[[], [1]]`?  The type of this should be `[[Integer]]`, not `[[a]]`.  Maybe slightly inefficient, I know, but I don't see a better way of doing this elegantly.  
But yes, I did forget to write the override for `getType` in `NilConsCell`.

krame505:
> > if I have production foo :: (Foo<a> ::= a), and I call hackUnparse on its signature I get something only involving skolemType, even though a should be flexible.
>
> The way I like to think of it is that every abstraction has an "outside" interpretation and an "inside" interpretation...

So do you mean that a different NamedSignature is available within in the production declaration than is placed in the env?

> > AFAIK there isn't a way for rigid type variables to occur on the LHS of a production, or flexible var on the RHS of a production.
>
> Correct on the first part (a consequence of productions always being at the top-level), but any variable that appears in the LHS will be flexible also in the RHS, but I think maybe you meant 'exclusively' on the RHS...

Yeah that's what I meant.

> > When these are passed down recursively as the expected type for children to reify into, we want to update the var with whatever the actual type of that child is. That is how I ensure that existential type vars used in multiple places in a signature actually refer to the same thing.
>
> Ah, hmm, maybe that does work. I can't think of a reason immediately that it wouldn't. ok.

To be clear: do you mean that this method of handling existentials seems OK, or my handing of type vars as a whole is fine?

> > he only big things are the reify function literal
>
> Can that one be done in the runtime instead? (actually, at the end of this post I mention some stuff that might make us want to...)

Literally the only thing happening here is creating an anonymous `NodeFactory` subclass with an `invoke()` that constructs the appropriate type and calls `Reflection.reify`.  I don't see what else there is to be moved to the runtime?  FYI, having `RefiyFn` that subclasses `NodeFactory` living in the runtime with a constructor taking a `TypeRep` doesn't work out, because I can't construct a `TypeRep` in a single expression (due to using shared objects to represent type variables.)  

> This made me think about something: maybe we should get rid of `ASTs` and `NamedASTs` and just use lists for that? Then you've got a `.length()` method to use right here, and I think that `AST` (unlike, uh, actual ASTs) doesn't have a reason for custom list types (usually, the use of inherited attributes.)
>
> Actually, unless you think of a good reason, I strongly think we should make this change. `ASTs` to `[AST]` and `NamedASTs` to `[NamedAST]`. Sorry I didn't notice this earlier!

I did it this way precisely because I can easily imagine analyses involving inherited attributes.  For example, checking whether a certain production exists within an AST, or performing some sort of substitution.  My idea is that extensions making use of this library would aspect the `AST` productions, adding whatever analyses that they are interested in performing.  

> I think we should allow failed reification to be handled. Maybe it should be `reify :: (Either<String a> ::= AST` ? (I really wish we had a 'Result' type instead of `Either` but oh well)

Makes sense, I suppose - so something a bit like parsers. 

> I think right now you make efforts to chain exceptions down, but probably we'd just want the originally thrown one. If you're okay with this idea. This might take of exception handling code that might be complicated (e.g. the `ReifyFailed` exception might be hidden inside `SilverTraceExceptions` or whatever they're called, so we might have to do work to decide if its actually a reification failure and return differently...)

I'm sorry, I don't follow what you are trying to say here.  I do think that this error message should *somehow* indicate where in the AST an error occurred - which is what I accomplish by chaining exceptions right now.  I guess I don't understand what you are proposing?

tedinski:
> So do you mean that a different NamedSignature is available within in the production declaration than is placed in the env?

Maybe maybe not, I forget how that worked. But certainly there's multiple interpretations of a production's type.

> To be clear: do you mean that this method of handling existentials seems OK, or my handing of type vars as a whole is fine?

Probably both?

> I don't see what else there is to be moved to the runtime?

Okay, nevermind.

> I did it this way precisely because I can easily imagine analyses involving inherited attributes.

Hmm. I guess stylistically I would expect people to write functions over `AST` rather than aspect them.

> I don't follow what you are trying to say here. I do think that this error message should somehow indicate where in the AST an error occurred - which is what I accomplish by chaining exceptions right now. I guess I don't understand what you are proposing?

Just thinking that if we have to render the error message as a string, maybe dumping the whole trace in there would be annoying rather than helpful. But maybe we can just see what this looks like.

But we do need to distinguish between a type error while reifying and some other error. Other errors should propogate out, but type errors should turn into `left` return values. I think this might take work because they're wrapped.

(Another thought: I didn't go back and look right now, but we should "handle" a production class `blah.Pwhatever` not existing as a "type error" that results in a `left` instead of crash.)

krame505:
> Just thinking that if we have to render the error message as a string, maybe dumping the whole trace in there would be annoying rather than helpful. But maybe we can just see what this looks like.

Hmm, seems reasonable - it should be possible to render this a bit more nicely.  I'll play around with this a bit.  

> Another thought: I didn't go back and look right now, but we should "handle" a production class blah.Pwhatever not existing as a "type error" that results in a left instead of crash.

Yeah I am checking for this now and reporting it in the same way as type errors.

krame505:
I implemented error handling for reification.  Basically, I subclassed `TraceException` with variants that are specialized to trace recursive elements of the reification process.  These have functions that reconstruct a representation of the partially-built AST (with irrelevant portions elided away.)  A wrapper function, `reifyChecked`, calls `reify` and catches any `SilverExceptions`, re-throwing any that are not do not have a `SilverError` as their root cause.  `SilverError`s are then transformed via the above to a displayable string error message to be returned.

krame505:
So, it turns out that my approach to type variables doesn't *quite* work in some corner cases, where we do not, in fact, want to freshen all type variables.  This occurs when the inferred result type of `reify` contains skolem constants, or if a function is wrapped up in `anyType` that is really a lambda with skolems as part of its type.  

I don't see too many potential cases where reification of something involving skolems would actually happen, but this could reasonably occur by accident.  For example:
```
function reifySkolem
Either<String a> ::= x::AST
{
  return reify(x);
}
```
The way I am currently doing things, the skolem constant a gets freshened to a new type var which is used in dynamic checking during reification, meaning that `reify` will accept any AST!  Then calling this function will basically allow you to cast anything into anything else, causing a runtime crash somewhere unrelated.  A similar issue is possible when reifying a tree containing a lambda that has a scoped type variable as part of its type, which could be illicitly reified into something more general without giving any errors.  

The good news is that this is easy to fix - there is nothing wrong with representing *unfreshened* skolem type constants with `BaseTypeRep`, a point I was confused on earlier since I freshen almost everything.  As part of this I also realized that I could significantly simplify unification, and refactor a whole bunch of other things.  

Also just an aside- there really isn't that much info out there about how skolem constants are supposed to work.  One of the top hits on Google is a reddit link to a blog post by Ted from 8 years ago complaining about exactly this!

tedinski:
> Also just an aside- there really isn't that much info out there about how skolem constants are supposed to work. One of the top hits on Google is a reddit link to a blog post by Ted from 8 years ago complaining about exactly this!

My biggest contribution to this world so far! Hah

krame505:
I think that this is basically ready to merge, unless there are any other comments.

krame505:
Also, I can't figure out why Jenkins is failing, it builds fine using fresh jars on my machine.  It seems to not be able to find some core:reflection generated files?

tedinski:
Because it's building tests actually-clean, having deleted all generated files first.

Generated productions reference `core:reflect` but if nothing ever causes it to be imported then it doesn't get built.

Either we have to be able to omit those references, or we have to make core import `core:reflect` I guess.

krame505:
OK, I'm importing `core:reflect` in `core` and it seems to be working.  This does mean that `core:reflect` can't (for now) rely on anything new in the runtime, so I had to temporarily disable the function I added to print the type of any object, which required changing the tests.  I will add this capability back in the next phase, where I refactor the reflection library.  

Since Jenkins is now happy, any reason not to merge?

tedinski:
You can merge if you like.

BUT, at the risk of annoying you, my thoughts on what I would do differently:

1. We _could_ move the definition of the `AST` nonterminal and its productions to `core`, thus eliminating the need to import `core:reflect`, which would still contain everything else (such as adding `pp` or functions or whatever.) Then we don't get this added dependency to all builds.
2. I still feel like interaction with `AST` would be done with functions over it rather than attributes, typically. I still would prefer using `[AST]` instead of having an `ASTs`. But if I can't convince you that's better, then oh well.

krame505:
As to 1 what I ended up doing with the refactoring in #217 is putting just the bare-bones AST nonterminal declaration in `core:reflect`, and moving everything else (attribute definitions, functions, etc.) to `silver:reflect`.  `core` imports `core:reflect` to trigger building the nonterminal (since generated code depends on this), but `AST` isn't visible to the user unless they import `silver:reflect` (which exports `core:reflect`.)  This avoids needing to explicitly pre-build `core:reflect` in `deep-rebuild`, but also keeps things a bit cleaner; in my opinion, we should keep `core` and anything under it to basically just build dependencies of Silver, and any optional libraries that the user imports explicitly should be somewhere under `silver`.  
But all that is in a seperate PR because it refactors and adds a lot of stuff not directly related to reification.  And it currently crashes Silver sometimes.  

As to 2, most of the cases I can imagine are equally well expressed with attributes as with functions, and while you can still always pattern match over `ASTs`, dealing with inherited attributes and lists gets ugly.  So I think we should wait and see how we end up actually using this library before making that change.  Also I'm lazy and don't really feel like changing everything.  

I'm going to go ahead and merge now.

Issue #152, opened 2017-04-27
Removing obsolete grammars, and naming new libraries


tedinski:
I wanted to write down naming intentions here.

We have a number of libraries that should be considered obsolete and removed:

```
grammars/silver/util/treemap  (use raw:treemap)
grammars/silver/util/fixedmap  (ditto)
grammars/silver/util/command  (use cmdargs)
grammars/lib/errors  (obsolete by langutil)
grammars/lib/system  (??!)
grammars/lib/stringmap  (obsolete by raw:treemap)
```

This library is one that we preserve only because it was used by the adventium grammars, which we don't want to break:

```
grammars/lib/xml
```

These should be mined of whatever value they have, and then removed as grammars, imo:

```
grammars/lib/extcore
grammars/silver/util (the files directly in this grammar, not the grammars below it)
```

We have a slight problem with the following grammars:

```
grammars/lib/json
grammars/lib/monto
grammars/lib/monto/helpers
grammars/ide
grammars/silver/definition/regex
```

I believe these are mis-placed. I'm not sure what a good naming scheme would be. I would consider it reasonable to perhaps place them all under `langutil`. But other suggestions very welcome! (Perhaps a "silver:languages:" grammar prefix for trivial library items like regex and json? Not sure if I like that idea though... gives the impression that's where languages should go, and it's really only special cases...)

The following are all fine: (Although I kinda regret the `raw` name. Perhaps we should hope to migrate away from that? And perhaps not have `tree` in the name for map/set either.)

```
grammars/silver/util/cmdargs
grammars/silver/util/deque
grammars/silver/util/raw/treemap
grammars/silver/util/raw/graph
grammars/silver/util/raw/treeset
grammars/silver/langutil
grammars/silver/langutil/pp
```

These are... maybe okay. I think it should be renamed `silver:util:testing`. (And they need cleanup.)

```
grammars/silver/testing
grammars/silver/testing/sampleTests
grammars/silver/testing/bin
```

Finally, I'll obsolete the old issue #12 and restate intentions for these grammars:

These should be corralled under `silver.compiler`:

```
grammars/silver/definition
grammars/silver/modification
grammars/silver/extension
grammars/silver/host
grammars/silver/composed
grammars/silver/analysis
grammars/silver/driver
```

And these should be renamed `silver.lang`:

```
grammars/core
grammars/core/monad
```

[...]

PR #217, opened 2018-01-27
Implement utilities for reflection

krame505:
This implements step 3 of the plan outlined in #212 - the `silver:reflect` library.  Everything has been moved out of `core:reflect` except for the actual `AST` nonterminal and production declarations.  `silver:reflect` now includes:
* The `reflect` and `reflectTypeName` functions
* Equations for `pp` and `unparse`
* A parser (and driver function `parseAST`) that parses a `String` created by `unparse` back into an `AST`.

Since the `silver:reflect` library is *not* exported by `core:reflect` (to avoid a build dependency), this library is optioned by `silver:langutil` where these attributes are defined.  If we wanted to avoid such a cross-dependency, duplicate attributes could be defined in `silver:reflect`, although this would be sub-optimal since these are both standard libraries, and I don't think that such a cross-dependence is a major issue.  

The only outstanding issue (besides lack of documentation - I intend to fix this as well!) is that strings are not being properly escaped and unescaped by `unparse` and `parseAST`.  I'm not really sure how best to go about this, or if this issue has already been solved - thoughts?  

This shouldn't be merged before #214 is merged, since it includes some of the same commits.


krame505:
Also for some strange reason this makes Silver crash when building some of the tutorials (e.g. lambda) - seems to be something to do with the flow analysis?  @tedinski could you take a look when you get a chance?

tedinski:
Unfortunately I probably won't get a chance to look at this till Thursday

tedinski:
It's just running out of stack space because you're importing the whole silver host language(!??) for `astParser`.

So... probably just don't do that.

krame505:
Oh, hmm. I guess I just wanted to reuse `QName` and a bunch of terminal declarations... annoying that this isn't factored out.  I suppose I can just duplicate the relevant stuff.

tedinski:
Honestly, it's a lot simpler than the silver grammar, just writing it from scratch is probably easiest.

krame505:
I want to try and finish this up in the next couple of weeks, there just a few things that need to be finished:
* Escape special characters in strings when unparsing or pretty-printing an AST, and un-escape strings when parsing an AST.  This somehow doesn't seem to be part of Java's standard libraries, but there are a few third-party libraries that do this... I'll probably end up needing to just re-implement it in part of the runtime.  `escapeString` and `unescapeString` do seem like things that belong in `core`... thoughts?  
* Call an object's native `toString()` function when pretty-printing an `anyAST` - fairly straightforward to do with a foreign function.  
* Deprecate `hackUnparse` as having reflection means that this should never be used in actual code.  `hackUnparse` can probably still exist as a quick debugging tool in `core` (do we want to add an `unparse` function that just does `show(reflect(x).pp)`?  It can't be in `core` anyway though to avoid a dependency on `silver:reflect`), but it would be nice to have some sort of warning attached to using `hackUnparse` in actual code (not sure how deprecation in Silver works, can we actually do this?)  Thoughts?  
* Replace `unparse` on `AST` with something that is a `Maybe<String>` or `Either<String String>`, instead of erroring out immediately when unparsing `anyAST`.  
* If interdependence with `silver:langutil` really is a massive issue, make duplicates of the `pp` and `pps` attributes with different names for `AST` (although I still don't understand why having `silver:reflect` an option of `silver:langutil` is a problem at all.)

tedinski:
> escapeString and unescapeString do seem like things that belong in core... thoughts?

Sure. They might even deserve a native Java implementation.

> Deprecate hackUnparse

Ideally, we'd just leave the function there for now, but have it call the reflection code instead of its current implementation.

> interdependence with silver:langutil

Could you write out what the packages are and what dependencies you're envisioning here? The main thing is we don't want core to depend on langutil. The only problem I can remember off the top of my head was that, because we have to generate code per-production, we always have to build whatever has `AST` which suggests that should type be in `core` to me...

krame505:
> Could you write out what the packages are and what dependencies you're envisioning here?

The `AST` nonterminals and productions are defined in `core:reflect`, completely by themselves with no attributes - this grammar has no dependencies.  `core` imports `core:reflect` to trigger the build but not make `AST` automatically visible in the environment.  All function and attribute definitions live in `silver:reflect`, which exports `core:reflect`, but `core` and `core:reflect` know nothing about `silver:reflect` or `silver:langutil`.  

`silver:reflect` imports `pp` and the `Document` constructors from `silver:langutil`, defining `pp` on `AST`, etc.  But since these are aspect productions on `AST` for an imported attribute, `silver:langutil` needs to have `silver:reflect` as an `option` to avoid an 'orphaned occurs' error (we can't have the option in `core:reflect` since that would add a build dependency.)  

> Ideally, we'd just leave the function there for now, but have it call the reflection code instead of its current implementation.

Right now doing so would add a dependence of `core` on `silver:reflect` (and also `silver:langutil`.)  I suppose we could move the definition of `reflectAST` to `core:reflect`, but I'm not sure how to get around the langutil dependency, and doing so would make the organization even more confusing anyway.  

It does seem, though, that "stuff needed to build the runtime" and "stuff we want in the environment by default" are seperate concepts; for example, there is no problem with having `hackUnparse` itself depend on `silver:langutil`, etc.  We could probably refactor `core` into two seperate grammars - move `hackUnparse` and all other utility functions not needed by the runtime into `silver:prelude`, which gets automatically imported instead of `core`.  `core` would still contain things that are dependencies of the runtime (e.g. `Maybe` and productions), and would be exported by `silver:prelude`.  Things that are runtime dependencies that we *don't* want in the environment by default, such as `AST`, can remain in sub-grammars of `core` that are imported by `core`, such as `core:reflect`.  

This is a slightly bigger task, though, and doesn't need to happen as part of this pull request.  Thoughts?

krame505:
... I guess this really also is tied in with the refactoring discussed in #152, so maybe we would have `silver:lang` and `silver:lang:core` instead of `silver:prelude` and `core`.  Further discussion of this specific issue should maybe happen over there.

tedinski:
Okay. The option doesn't sound like the worst thing, for now. I still don't like it, and reserve the right to have thoughts when I see what that code looks like.

I can see why you want to use the `Document` stuff, but I also think this could probably be implemented as just a `String` if that simplifies anything...

> Right now doing so would add a dependence

Oh, right. Well, step one, let's not worry about it yet. Maybe all we do is move the function out of core into the reflection code, so the breaking change can be fixed by just adding an import.

krame505:
Per the discussion in #152, I think I will go ahead and move the `pp` equations to `silver:langutil:reflect`, conditionally exported by `silver:langutil`.  I will leave `hackUnparse` alone for now.  

For now the rest of the library will stay in `core:reflect` and `silver:reflect`, but these will be combined into  `silver:lang:reflect` when the #152 refactoring occurs.

tedinski:
Sounds like a solid plan!

krame505:
@tedinski What's the procedure for getting Jenkins to build after making simultaneous changes to the runtime and core?  I have working jars on my local machine, but I'm not sure if there's a way to just publish jars for a specific branch?

krame505:
^ fixed a bug where the `escapeString` and `unescapeString` functions in `silver:definition:env` (that only handle backslashes and newlines) were being shadowed in some places, causing some weird bugs with translation in non-clean builds.  Anyway, this should be totally ready to merge, as soon as I know how to make Jenkins build after core changes that depend on the runtime.

tedinski:
On Thursday, I will attempt to review and add a jenkins feature that lets you send a build parameter with a different jars location. It wouldn't automatically work, but at least it'd let you manually issue a rebuild with the jars to use...

krame505:
Great - once a branch has built successfully once, do we use the resulting jars in future (automatic) builds of that same branch?

tedinski:
Hmm. That would make it work pretty well. I don't know. Maybe I could find a way to get that behavior. I'll see.

krame505:
OK - I'll open an issue with some further thoughts.

krame505:
@tedinski I have absolutely no idea why some of the tutorials are failing on this branch for Jenkins - they build fine on my machine and my local checkout on coldpress.

tedinski:
Ah ha. The major thing to note about tutorials is they build against what gets shipped with silver in the tarball. If tutorials mysteriously fail on jenkins, that's a clue of a simple problem: you added a grammar and didn't tell `make-dist` to include it in the tarball. Tutorials should always work in what we distribute, so we deliberately test this.

Should be a one line addition to the `make-dist` script.

krame505:
OK, thanks.  Let's see if that fixes it...

And one other issue: This branch requires some slight changes in ableC (added a `pps` attribute to langutil since that is used everywhere.)  This is in the ableC `feature/langutil` branch.  I suppose if I renamed the ableC branch to `feature/reflection`, we could add a similar thing to the Silver Jenkinsfile that we have with ableC for building the extension branch of the same name, if a branch of that name exists.

krame505:
Since Jenkins is (finally) happy with this branch, I'm going to merge now, and immediately merge `feature/jenkins` in ableC before the Silver build gets that far on Jenkins.

PR #221, opened 2018-03-02
Better errors, changes JSON number output

remexre:
This changes error messages to include the names of terminals that have a `named`... property?

```
terminal Identifier_t /[A-Za-z][A-Za-z0-9]*/ lexer classes {IDENTIFIER}, named "an identifier";
```

Also, if there is only one string a terminal can expand to, that string will be shown. For example, an error message might be:

```
Expecting $, '&&', '/', 'else', 'end', '==', '>=', '>', 'in', '<=', '<', a comment, '-', '*', '!=', '||', '+', ')', 'then', or some whitespace, got an identifier
```

---

Also, this changes `jsonFloat.jsonString` to print a float that exactly represents an integer as an integer. (Because apparently not everyone actually follows the spec w.r.t. deserializing `3.0` into a `int` as being valid.)

---

Also, since I've got a line in my vimrc that strips trailing whitespace (`autocmd BufWritePre * :%s/\s\+$//e
`), there's a decent amount of that.


krame505:
Looks generally OK after a (very) quick glance, although I would like to wait on merging this until #217 is merged, since these are likely to have merge conflicts.  Also we are fixing some issues with Jenkins right now so it might be a bit.

remexre:
Okay, I'm fine with waiting since my stuff works on this branch since I rebased it onto develop.

grammars/core/List.sv
```diff
@@ -172,6 +172,20 @@ function removeAllBy
         else removeAllBy(eq, tail(ys), removeBy(eq, head(ys), xs));
 }
 
+{--
+ - Returns the initial elements of a list.
+ -
+ - @param lst  The list to examine
+ - @return  The initial elements of 'lst'. If 'lst' is empty, crash.
+ -}
+function allButLast
```
tedinski: Haskell calls this function `init`. So does Scala. Ug, so does Ocaml.

I don't think we should call it `init`, it's too generic to drop into our global namespace. (Someday, if we have type-base namespaces to work with...) But I do think we should think carefully about names before adding them. Since everyone seems to use some variant on `init`, what about `initList` or `initialList`?

grammars/silver/definition/concrete_syntax/TerminalDcl.sv
```diff
@@ -1,9 +1,11 @@
 grammar silver:definition:concrete_syntax;
 
 import silver:definition:regex;
+import silver:modification:ffi:util;
```
tedinski: An import from the core language to a 'modification' should not happen. (I mean, we screw this up all over already, but we shouldn't make it worse, probably.) Can this be fixed?

grammars/silver/definition/concrete_syntax/ast/PrettyNames.sv
```diff
@@ -0,0 +1,149 @@
+grammar silver:definition:concrete_syntax:ast;
+
+function asPrettyName
+Maybe<String> ::= r::Regex
+{
+  return mapMaybe(\x::String -> "'" ++ xmlEscapeString(x) ++ "'", r.asLiteral);
+}
+
+synthesized attribute asLiteral::Maybe<String> occurs on Regex, RegexSeq,
```
tedinski: Can I get a little documentation here about what this attribute does? I *think* it's trying to recognize literal regexes, and returns `nothing` otherwise, but I like to not have to read code to know that.

tedinski:
For reference, this makes error messages look like this (for silver itself):

```
Errors for silver:composed:Default
 [/project/melt4/People/ted/git/silver/grammars/silver/composed/Default/Main.sv]
Main.sv:11:0: error: Syntax error:
Error at line 11, column 0 in file Main.sv
         (parser state: 88; real character index: 243):
  Expected a token of one of the following types:
   [$,'ignore',
    'marking','abstract',
    'annotation','aspect',
    'attribute',silver:definition:core:BlockComments,
    'closed',silver:definition:core:Comments,
    'concrete','function',
    'global','import',
    'inherited','nonterminal',
    'synthesized','terminal',
    silver:definition:core:WhiteSpace,'flowtype',
    'deprecated','{@comment',
    '{@config',silver:extension:doc:core:NoDclComment_t,
    'equalityTest','mainTestSuite',
    'makeTestSuite','wrongCode',
    'wrongFlowCode','derive',
    'testFor','autocopy',
    'disambiguate','lexer',
    'parser','prefix',
    'copper_mda','temp_imp_ide_font',
    'temp_imp_ide_dcl','type']
   Input currently matches:
   [silver:extension:doc:core:CommentText_t]
```

That's pretty great. I like it. I hadn't thought of the approach of just analyzing the regex to figure out if it's a literal and we can use a short name. Nice.

I don't like the change to `ParseResult` `parseErrors` though. For ableC we're getting errors like:

```
Expecting '=', edu:umn:cs:melt:ableC:concretesyntax:BlockComment, ',', edu:umn:cs:melt:ableC:concretesyntax:LineComment,
edu:umn:cs:melt:ableC:concretesyntax:NewLine_t, ';', edu:umn:cs:melt:ableC:concretesyntax:Spaces_t,
edu:umn:cs:melt:ableC:concretesyntax:cppTags:CPP_Location_Tag_t, 'asm', '__asm__', '__attribute__', '__asm', '__attribute',
or edu:umn:cs:melt:ableC:concretesyntax:gcc_exts:PragmaMark, got edu:umn:cs:melt:ableC:concretesyntax:Identifier_t,
edu:umn:cs:melt:ableC:concretesyntax:TypeName_t, and edu:umn:cs:melt:ableC:concretesyntax:gcc_exts:AttributeNameUnfetterdByKeywords_t
```

Not only is this harder to read, there's no line/column info. :)

Copper's string might be not the prettiest, but I think we should leave that the default for `ParseResult`, and we can (if we want to, later) make things pretty inside the individual compilers.

grammars/core/List.sv
krame505: I kinda do prefer just plain `init` - after all this is part of Haskell's prelude, so I don't really see how this is much different in our case.  Also we don't suffix any of our other generally-named standard list functions (e.g. `concat` or `implode`) even though there are other non-list functions that do want to use the same names (`sconcat`, `ppImplode`) - IMHO being less consistent will make the problem worse.

tedinski: I have a feeling changing the name to `init` might cause breakage in already written code somewhere. But I guess if it doesn't then that'd be okay with me.

remexre:
Would you be opposed to me making either a new attribute or a function to give this error? I'm using the shorter error in Monto, since the error is displayed at the correct line/column already. Or would you prefer I move the whole of the code that makes that error to Monto?

tedinski:
Hm, would you mind moving that into Monto for now?

grammars/core/List.sv
krame505: After some quick grepping I don't see any conflicts.

grammars/silver/definition/concrete_syntax/TerminalDcl.sv
remexre: I need that for `cleanStringLexeme`; is there a better solution somewhere else that I should be using?

grammars/core/List.sv
krame505: Oh, I forgot about ableC - I guess there is a conflict there.  But that production/nonterminal could probably use a better name anyway, so I vote we change it there.

remexre:
I'm not sure how to read the log output; all I can see is:

```
[Pipeline] [ableC] }
[ableC] Failed in branch ableC
[Pipeline] // parallel
[Pipeline] }
[Pipeline] // stage
[Pipeline] echo
Setting failure flag
```

krame505:
On http://coldpress.cs.umn.edu:8080/job/melt-umn/job/silver/job/develop/ if you over over the Integration stage of the build that failed,  click on Logs, and then expand the item that failed, it will give you a link to the corresponding ableC build that failed.  You should look at the console output for that.

remexre:
I think this might have something to do with the conflict between the `init` in core and the one in ableC; let's try switching that? Or should I leave `allButLast`?

Also, what should I rename `init` to? It's a `production` for the `Init` type, so it's not like it makes 100% sense to rename it to `initialize` or something...

krame505:
I guess you didn't see my comment from earlier:
> Oh, I forgot about ableC - I guess there is a conflict there. But that production/nonterminal could probably use a better name anyway, so I vote we change it there.

krame505:
Maybe rename the nonterminal and productions to`InitItem`?  Or maybe just rename `init` to `positionalInit`?  IDK which one I like better.

tedinski:
> Or maybe just rename init to positionalInit?

Seems fine to me. Who's going to do it?

remexre:
I'm doing this right now on the ableC init branch, but I think I broke the pps attribute somehow...

tedinski:
Lucas made some breaking changes recently there.

Oh, probably merge develop or rebase your branch of Silver here, that's likely the problem.

remexre:
Okay, it works on my machine now; there's still the stuff in `grammars/silver/definition/concrete_syntax/TerminalDcl.sv` though.

grammars/silver/definition/concrete_syntax/TerminalDcl.sv
tedinski: Oh, probably just do the `substring(1,length(s)-1, s)` yourself here, I don't think you actually want `cleanStringLexeme` since that's potentially adding newlines and such to the string?

tedinski:
I made a commit to develop, to increase the stack limit. If you think everything should be working, feel free to merge all these branches. Just be ready to fix anything in case you're wrong. :)

krame505:
@remexre FYI, you will need to merge develop into your branch to get ableC-refcount-closure to build.

remexre:
I think it already worked?

krame505:
Nope: http://coldpress.cs.umn.edu:8080/job/melt-umn/job/ableC-refcount-closure/job/develop/9/
The overall silver build hasn't finished yet, though.

remexre:
Looks like ableC wasn't the only thing using `init`...

krame505:
OK.  I'm not really familiar with ableP, but go ahead and fix it, I guess...

Also, any reason why there are so many force pushes on this branch?  Generally we try to avoid those when possible.

krame505:
I'm not any more familiar with ableP than you, but I think Ted did migrate this to github recently.  

I think in our workflow the general policy is try to avoid rebasing when possible - merge commits are annoying but less bad when things go wrong.  But you should check with @ericvanwyk about this.

remexre:
Oh, looks like I don't have push access to `ableP` as well; it looks like the `init` production is for Promela's `init` block, which contains initializers; would it make sense to name this `initializers` or `initUnit`, or something else?

krame505:
I fixed the permissions issue.  
It seems like `init` is actually the name of some sort of construct in Promela, so IDK - @ericvanwyk thoughts on renaming this?

ericvanwyk:
Renaming ``init`` in ableP is fine.  Just call it ``initConstruct`` - assuming that isn't also used.  The add a comment about the change.

We may review Promela someday, it is still the best example of the power of context aware scanning.

remexre:
Okay, looks like everything's building; is this good to merge?

PR #222, opened 2018-03-03
Tweaks to Jenkins

krame505:
Since Silver and ableC are both building successfully now on feature/jenkins, I'm going to go ahead and merge these both back into develop.


krame505:
I'm merging these now.

PR #231, opened 2018-03-16
Fixes to built-in type conversion functions

krame505:
I ran into a bug where Silver crashes when translating `toInt(Boolean)`, so I decided to actually fix this.  I also implemented the `toBoolean` builtin function, converting `Integer`, `Float`, or `String` values to `Boolean` values.  

This still needs tests, but I figured I'd open a PR first to check if I'm on the right track.  

While I'm fixing these conversion things, I've also been thinking of renaming `toInt` -> `toInteger` to be consistent, and depericating `toInt`.  Thoughts?

### DIFF
--- a/grammars/silver/analysis/typechecking/core/BuiltInFunctions.sv
+++ b/grammars/silver/analysis/typechecking/core/BuiltInFunctions.sv
@@ -17,6 +17,18 @@ top::Expr ::= e::Decorated Expr
   top.upSubst = top.downSubst;
 }
 
+aspect production toBooleanFunction
+top::Expr ::= 'toBoolean' '(' e1::Expr ')'
+{
+  e1.downSubst = top.downSubst;
+  top.upSubst = e1.upSubst;
+  
+  top.errors <-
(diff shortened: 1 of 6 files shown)


krame505:
Also, I wasn't totally sure of the expected behaviour of `toFloat(Boolean)`, but I made this return either 0.0 or 1.0, for false and true inputs, since I couldn't really think of a better option.

tedinski:
1. All of these should have a test. Obviously we were missing that...
2. I kinda dislike Float<->Boolean at all, but oh well.
3. If you want to add `toInteger` go ahead, but you don't have to. I wouldn't add any warning or anything for `toInt` yet though.

krame505:
> I kinda dislike Float<->Boolean at all, but oh well.
Agreed, but that would make type checking more complicated...

I'll add `toInteger` and tests soon when I get a chance.

krame505:
Done, and done.  Once Jenkins finishes building, is this good to merge?

tedinski:
lgtm

Issue #33, opened 2015-08-20
Function type errors in arguments are wrong and confusing


GoogleCodeExporter:
```
This stems from the 'reverse' done on the type list.

Too many args, and all types will be reported wrong.
```

Original issue reported on code.google.com by `tux...@gmail.com` on 1 Mar 2013 at 9:50

tedinski:
I'm going to add another note to this problem: when calling `doesNotExist(x)` we raise an error about `Too many arguments provided to function` when this should only be raised if we actually know how many arguments the function should have.

PR #240, opened 2018-04-19
Introduce `errorType` and suppress unnecessary error messages

### LINKED ISSUE #19: Wrong aspect signature forgets all production attributes
```

Any aspect production that has a wrong signature (for example, missing a 
parameter) will not see any production attributes.

The trouble is that the type of some production attributes may depend on the 
aspect signature.  So this isn't a "bug" per se, it's just awful error 
reporting.

What is the expected output? What do you see instead?

No errors for undefined production attributes, even if there are type errors as 
a result of not having a "correct" aspect signature to necessarily determine 
the types for production attributes.

Please use labels and text to provide additional information.

As a stop-gap measure, I'm going to make the signature error message properly 
appear before the body's errors, so at least handling errors one at a time, 
top-down won't be confusing...
```

Original issue reported on code.google.com by `tux...@gmail.com` on 13 Mar 2012 at 12:40

### LINKED ISSUE #197: Annotation 'occurs on' not exported by the nonterminal is allowed
This should be an error, since other grammars where the occurrence is not visible, calling the productions for the nonterminal would not give a value for the annotation.  
I haven't actually tried it, but trying to access the annotation value from trees constructed where the occurrence is not visible probably would result in a crash.

tedinski:
We'd long ago planned this and I felt like getting it done. This patch turns an error message list like this one:

```
Test.sv:6:5: error: Undeclared type 'E1'.
Test.sv:6:15: error: Undeclared type 'E2'.
Test.sv:8:6: error: Undeclared attribute 's1'.
Test.sv:8:11: error: Undeclared value 'func'.
Test.sv:8:11: error: func has type a and cannot be invoked as a function.
Test.sv:8:16: error: Too many arguments provided to function 'func'
Test.sv:8:16: error: LHS of '.' is type a and cannot have attributes.
Test.sv:8:18: error: Undeclared attribute 's2'.
Test.sv:8:22: error: Undeclared value 'r'.
Test.sv:8:22: error: LHS of '.' is type a and cannot have attributes.
Test.sv:8:24: error: Undeclared attribute 's3'.
Test.sv:9:2: error: Undeclared value 'e'.
Test.sv:9:2: error: Cannot define attributes on e
Test.sv:9:4: error: Undeclared attribute 'inh'.
Test.sv:10:6: error: Attribute 'realsyn' does not occur on 'a'. Looked at:
	errors:realsyn (Test.sv:3)
Test.sv:10:16: error: Undeclared value 'x'.
Test.sv:10:16: error: LHS of '.' is type a and cannot have attributes.
Test.sv:10:18: error: Attribute 'realsyn' does not occur on 'a'. Looked at:
	errors:realsyn (Test.sv:3)
```

into one like this

```
Test.sv:6:5: error: Undeclared type 'E1'.
Test.sv:6:15: error: Undeclared type 'E2'.
Test.sv:8:6: error: Undeclared attribute 's1'.
Test.sv:8:11: error: Undeclared value 'func'.
Test.sv:8:16: error: Too many arguments provided to function 'func'
Test.sv:8:18: error: Undeclared attribute 's2'.
Test.sv:8:22: error: Undeclared value 'r'.
Test.sv:8:24: error: Undeclared attribute 's3'.
Test.sv:9:2: error: Undeclared value 'e'.
Test.sv:9:4: error: Undeclared attribute 'inh'.
Test.sv:10:16: error: Undeclared value 'x'.
```

I've added a note to #33 about eliminating that last outlier (too many args) in the future there.

This required more plumbing changes that I hoped it would when I started it for funzies earlier today. I had to introduce `x.found` instead of using `null(x.errors)` because now the relevant error can be suppressed. This is especially vexing in one location where we might use an error production with no errors --- because we know they were raised elsewhere. Hm.

This also fixes #197 which was a typo I came across while adding `found`.

This also fixes #19 by making use of error types instead of just giving up.

I also discovered a couple of long standing bugs because error types refuse to translate, including a rather surprising type inference bug in the second commit on this branch. Yikes.

Given that unexpected, and very old, bugs were revealed by this change, we might see some crashes when trying to compile erroneous silver code going forward. Hopefully not.


ericvanwyk:
Excellent!

PR #241, opened 2018-04-19
Merging docgen into develop

tedinski:
I'm going to begin changes to how the website gets generated, and as part of that, I want to get rid of this branch, which is currently what the documentation for the site uses. It looks like this merges cleanly, as it only contains changes to the documentation itself and the doc gen stuff.

This looks unobjectionable to me, thoughts @TravisCarlson  @ericvanwyk ?


tedinski:
Okay, I want to get going with this, so merging.

ericvanwyk:
I'm too late, but this is OK with me.  Thanks.

PR #244, opened 2018-05-03
Improved logging for Monto.

remexre:
This is a bit short for a PR, but it fails Jenkins and idk why...

### DIFF
--- a/runtime/java/src/monto/Server.java
+++ b/runtime/java/src/monto/Server.java
@@ -57,13 +57,17 @@ public class Server {
 			this.inner = inner;
 		}
 		public void handle(HttpExchange t) throws IOException {
-			System.out.println(t);
+			System.out.println("Starting to handle " + t.getRequestURI() + "...");
+			long startTime = System.nanoTime();
 			try {
 				inner.handle(t);
 			} catch(Exception ex) {
 				ex.printStackTrace();
 				throw ex;
 			}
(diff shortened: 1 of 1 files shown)


tedinski:
Probably the trouble is this branch is older, so it has its own 'jar' files in Jenkins, but then you merged develop (or rebased onto it), where we'd had some tricky runtime changes made awhile back. So now those changes are using your branches older jars and so it doesn't work. Yuck.

Any thoughts on how to avoid this problem @krame505 ? Maybe we ought to only do the "fetch jars from branch's last build" for specially named branches or something, idk.

The workaround is just to re-run your branch's build, but explicitly setting the `OVERRIDE_JARS` parameter in jenkins to `/export/scratch/melt-jenkins/custom-silver/` to force it to use normal jars. I did this just now, probably this build will succeed.

tedinski:
(Whoops should be set to `/export/scratch/melt-jenkins/custom-silver/jars`...)

krame505:
Isn't this essentially the intended behaviour though?  We normally want to keep using the jars on the feature branch, as we are assuming that any development involving weird bootstrapping changes to jars happened there.  There isn't really an automatic way possible of handling multiple active feature branches...
Maybe a fix that could make this a tad nicer is to check if `OVERRIDE_JARS` is just set to "develop", in which case we should use `/export/scratch/melt-jenkins/custom-silver/jars`?

remexre:
Is this fine to merge?

PR #247, opened 2018-05-09
Adding silver-custom script

krame505:
You know, I think this approach is perfectly fine.  Thoughts?


ericvanwyk:
I like it.  It doesn't change an existing bit of software and leaves room
for changing the approach later.

PR #250, opened 2018-05-11
Further tweaks to silver-custom script

krame505:
This makes the `silver` script call the `silver-custom` script to avoid having the JVM parameters in 2 different places.  Also further bump the default heap space, since I was crashing things when doing a clean build of Silver+ableC.

### DIFF
--- a/support/bin/silver
+++ b/support/bin/silver
@@ -24,5 +24,3 @@ export SILVER_HOME=${ME/support\/bin\/silver/}
 SILVER="$SILVER_HOME/jars/silver.composed.Default.jar"
 
 silver-custom "$SILVER" "$@"
-
-
--- a/support/bin/silver
+++ b/support/bin/silver
@@ -23,4 +23,5 @@ export SILVER_HOME=${ME/support\/bin\/silver/}
 # Find the silver jar
 SILVER="$SILVER_HOME/jars/silver.composed.Default.jar"
 
-silver-custom "$SILVER" "$@"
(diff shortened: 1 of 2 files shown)


tedinski:
Not a big fan of either change.

This change to the script will break everyone's existing installs. We could maybe do it the other way around though.

Why do we need more memory? Increasing this number is a very big problem because it means we're demanding a lot more of users, who may be running silver in a small vagrant vm or something like that. If you have to, we have to, but this shouldn't really be done without figuring out why memory demands have increased, and checking there's no other easy way. If it's just the compilation of silver-ablec, I think we should just change it there for now.

krame505:
> This change to the script will break everyone's existing installs. We could maybe do it the other way around though.
Hmm, I suppose so.  Not sure what you mean by "do it the other way around" - do you mean adding an overridable environment variable to the silver script with the jar to use?  

>  If it's just the compilation of silver-ablec, I think we should just change it there for now.
Yeah, building Silver + ableC + other stuff all at once sets the record for the largest single silver build.  I suppose that would be fine.

tedinski:
> do you mean adding an overridable environment variable to the silver script with the jar to use?

Probably. I can probably figure this out later... I've been meaning to get back to my refactoring scripts PR, it's just jenkins and melt-website and doc stuff is higher on my todo list at the moment.

krame505:
OK, reasonable.  I'll just close this for now, then.

PR #255, opened 2018-06-22
Use reflection library to replace interface file parsers

krame505:
Note that this PR includes the changes from #254 and should be merged after that.  

This uses the silver reflection/serialization/deserialization/reification library to replace our existing tedious method of pp'ing and parsing the entire environment in a custom way.  All environment data that would have been printed manually is instead wrapped into a `GrammarProperties` nonterminal, which gets reflected and serialized into the interface file.  The reverse process of deserialization and reification is used instead when loading an interface file.  

The only downside to this is that there a slight impact to performance - my guess is that this is due to the heavy use of java reflection and the runtime type checking needed to implementing `reify`.  I haven't measured it, but the initial phase when compiling Silver seems to be a couple of seconds slower on my machine.  Since this is minor compared to the amount of time spent in error checking and translation, which are unaffected by this change, this seems like a reasonable price to pay.


tedinski:
I definitely like the idea of getting rid of sviParse. Don't have time to review, but that performance regression is concerning. We should try to figure out why that is before merging, I think...

krame505:
Possible idea about improving the performance: make reification lazy - we always reconstruct the entire env tree, but in practice would probably only demand a small portion.  

Also, there seems to be a weird bug with this where Silver crashes with infinite recursion during type checking when running `self-compile` on this branch when only certain branches (e.g. `silver:modification:copper`) are touched.  I have no idea of what is going on here, but it doesn't seem to be a simple shortage of stack space.  My best guess is that this is an existing bug with the type checking code that was being masked by the old SVI parser somehow.  Any ideas on things to investigate?

tedinski:
Not many. I'd look at the stack trace, and try to find a place to put an `unsafeTrace` print of information I think might help me understand what's happening better.

I won't have time to try anything myself for at least a week or two though.

tedinski:
Once the other branch is merged, would you might rebasing this one, to help me review it more easily?

krame505:
I think just merging develop into this branch was sufficient to have only the changes in this branch show up on github?  Not sure what I would be rebasing.  Or did you mean something else?

tedinski:
Ok, at first glance, as someone who loves deleting code, this is beautiful. There's even less changed code necessary than I thought we'd need.

I think we should change the name `GrammarProperties`, but I don't know to what yet. Don't worry about it. I'll do it later or something.

If I understand right, we have two issues to investigate:

1. There's a crash bug or something related to actually using deserialized trees, yes?
2. Possible performance loss compared to the old interface files.

I'll start looking to reproduce that crash...

krame505:
Thanks, yeah.  I agree `GrammarProperties` isn't the best name, I couldn't think of anything better.  
You can reproduce the crash just by bootstrap compiling Silver on this branch, then touching random files and running self-compile until it fails.

tedinski:
My current, unconfirmed best guess about the problem:

Serializing things means we're exactly replicating `tyVar`s between two invocations of the compiler. I don't know to what extent this is a real problem, most code should be fine with it.

However, I believe there's one case where we're NOT fine! When building core, we replicate out `tyVar(140)` to `Silver.svi`. Then, when re-building in a certain order, we end up in a situation where (I *THINK*) we're calling `freshenCompletely` on this type for `cons` and `freshTyVar` just unluckily comes back with `tyVar(140)` as the replacement for... `tyVar(140)`. Infinite loop.

Currently pondering what the simplest, least invasive way to ensure this doesn't happen is. Maybe `freshenCompletely` and whatnot could use a non-recursive shallow variant of substitution, where this wouldn't cause a loop.

krame505:
Hmm.  Isn't the deeper issue here that `genInt()` just doesn't provide a value that's unique between runs?  It seems like this could quite possibly break a lot of other things as well.  
What about changing it to start off counting from the current unix time instead of 0, or something like that?  Or maybe we could be smarter and actually save the last value of `genInt` in a `generated/` file somewhere and somehow and re-load it on startup?  
Another, slightly smarter possible option is we introduce a `Unique` foreign type in Silver, from which we can obtain a unique identifier.  When this type gets reified, we can do some magic to ensure that the underlying unique integer that it contains gets mapped to a new value that doesn't conflict with any other `Unique` values that have been created or loaded.  
Or maybe we just add a function that generates a sufficiently large random alphanumeric string that is highly unlikely to ever conflict with anything.

tedinski:
Also note to self because I'm not going to get to it yet, but there's also a fair bit of `unparse` code that can be removed too!

tedinski:
>  It seems like this could quite possibly break a lot of other things as well.

It actually shouldn't... Well, there's some places where we generated names with `genInt` but I think those all should have some extra information that prevents any problems.

tedinski:
I have added a commit that fixes the crash. I believe this should be a complete fix, but I will have to take some time next week to audit that.

test/cstast/Test.sv
```diff
@@ -1,91 +0,0 @@
-grammar cstast;
```
tedinski: Why was test/cstast removed?

krame505: As far as I can tell this is just testing serialization/deserialization of the cstast using the env parser, correct?  This is also handled automatically now, so I'm not sure what the point of still testing this would be?

krame505:
> I have added a commit that fixes the crash. I believe this should be a complete fix, but I will have to take some time next week to audit that.

Thanks, as far as I can tell this is working fine for me now.  

Also I ran some basic performance checks with my lazy reification experiment and it turns out that's about 20% slower, so yeah... I guess we aren't going that route.

test/cstast/Test.sv
tedinski: It was testing `xmlCopper` at bit, but looking at it now, I think we're better off removing this test suite. It's atrophied down to nothing useful, I think.

But that means we need to remove calling it from Jenkinsfile.

tedinski:
OK. I think I have completed that audit. We should now not have any more bugs related to re-use of tyvar numbers from interface files. Probably. Assuming this build succeeds, this is probably ready to merge.

tedinski:
Probably okay, since I can't actually measure any performance loss, I think?

krame505:
The initial phase of compilation takes maybe a couple of seconds longer on my laptop than it did before, but I agree that's not enough to worry.  Also not building sviParse means building Silver is still faster overall!  
Also, did you want to rename `GrammarProperties` before this is merged?

tedinski:
No don't worry about that renaming, I'll make a separate branch to do some more stuff later.

tedinski:
One possibility, maybe, is to write a hand parser for deserialization. We could avoid a lot of overhead that way, it's a trivial one to write I think, and it's not like that's a parser that is subject to extension. But this can definitely wait until much later, when we'd be surer the parser doesn't need to change at all anymore.

tedinski:
Ah good, the build now goes through. lgtm.

krame505:
> One possibility, maybe, is to write a hand parser for deserialization.

Not sure how much of an improvement that would be - Copper is pretty fast, and we were using a Copper parser for interface stuff previously without much of a penalty.  I haven't checked in more detail, but I highly suspect the major use of reflection and dynamic type checking of the entire interface tree is a larger contribution to the runtime for deserializing to Silver trees.

krame505:
Actually I'm not so sure on second thought - the serialized interface files are a lot bigger than they were before (`silver:definition:core` is over 1 MB, for example.)  But it would be interesting to profile this to see what actually is the major contributor here.  I'm honestly surprised it's as fast as it is, tbh.

PR #261, opened 2018-08-07
Remove unparsing code that is no longer used, thanks to serialization.

tedinski:
Just a whole lot of deleted code, really.


tedinski:
Pretty sure this failure is something busted with the silver-ablec jenkinsfile. @krame505 ?

krame505:
Jenkinsfile should be fixed, minor bug.  No idea why it was even working in the first place.

Issue #92, opened 2016-08-31
Document and Crispen IDE Interface


TravisCarlson:
1. Documentation on how to install the Silver IDE plugin, currently at http://melt.cs.umn.edu/silver/doc/dev/eclipse/getting-started/, should no longer be under `dev`.
2. The IDE interface and features that are available for modification when developing a new plugin should be cleaned up and documented.
   - As one example, rather than expecting the plugin writer to specify fonts for all lexer classes, creating standard IDE lexer classes would allow the work of selecting terminal fonts to be done in the Silver code for new languages rather than in the Java code for the plugin.
   - So, if a language defines `terminal 'if' lexer classes { MY_KEYWORDS }`, the current plan is to add a line to the plugin TokenClassifier like `map.put("grammar:path:MY_KEYWORDS", TokenType.silver_definition_core_font_keyword);` then rebuild the plugin. Having a standard `IDE_KEYWORDS` (or maybe just `KEYWORDS`?) would allow the font to specified in the Silver code as in `terminal 'if' lexer classes { MY_KEYWORDS, IDE_KEYWORDS }`.
   - An advantage to this method is that no modification to the plugin or knowledge of its Java internals is needed when selecting fonts.
3. Documentation on how to create an IDE plugin for a new language should be created on the website.

tedinski:
Regarding (2): A good way to do this would be to drop some standardish lexer classes into `silver:langutil` and then implement a "default colorer" in the silver-eclipse runtime that recognized those and colored them in a default way.

The downside of doing this, though, is that we'd have to start including `silver:langutil` in our parser declarations.

Regarding (3): We might look into creating a Maven Archetype for this. I'm not sure how flexible the maven archetype plugin is, though. The documentation I found in a stupid quick google was too simplistic. Wasn't clear if we could generate a "maven project with three maven subprojects" skeleton.

TravisCarlson:
Here is the current list of features that are available for plugin writers to modify. Please let me know if I am missing anything, if anything here looks like it should not be exposed, or if you have any other ideas for improvement.

Coloring and folding seem to be the most desirable features that are also relatively simple to implement in a new language. File stubs are nice to have and relatively simple. Building and exporting are also desirable, but seem to me to maybe be less straight-forward to update for new languages.

## Coloring

[...]

PR #263, opened 2018-08-08
Adopt langutil in the Silver compiler

### LINKED ISSUE #258: Using langutil in the silver compiler
`silver:langutil` provides a bunch of small common things we recognized from the Silver compiler for all compiler implementations in Silver to use. But then we never moved the Silver compiler over.

The root reason is that `pp` is a `Document` in `langutil` but a `String` in Silver. So switching over is non-trivial.

One major issue is that the Silver `pp` attribute is used in weird ways. I think it's not actually a `pp` really. So the correct fix is probably to switch over to using `unparse` in Silver instead of `pp`!

To convince ourselves of this, we should do a minor audit. I've tried to filter down things as follows:

```
git grep -w pp | grep -v "top.pp =" | grep -v "location, pp,"
```

(The point is to find uses of `pp` that aren't themselves `pp` equations or just occurrences.)

I haven't been thorough with this, but what I see are:

1. Debugging uses (e.g. translation)
2. Error messages **that probably shouldn't be using pp** (e.g. they should be accessing a `.name` attribute or something like that instead.)
3. Some stuff in the unfinished `doc` extension that's probably temporary hacks.

So I think we should fix 2, and be content renaming `pp` to `unparse` for 1 & 3. Then we can switch over to using langutil.

## Replaced stuff

```
silver/definition/env/Unparsing.sv:synthesized attribute unparse :: String;
silver/definition/core/Attributes.sv:synthesized attribute errors :: [Message] with ++;
silver/definition/core/Message.sv:nonterminal Message with loc, pp, severity, msg;
```

## Additional step

In #81 I think, we were talking about having to make some changes to the Silver IDE code to accommodate this change. That shouldn't be forgotten about.

tedinski:
This is a big patch, mostly for the middle commit that does an automated rename.

Fixes #258 

Enables us to make progress on old IDE stuff in #81 and #92 which I want to take a stab at.


tedinski:
Hm, looks like this one will need corresponding changes to silver-ablec.

krame505:
I'm not 100% sure, but I think most of the errors in silver-ableC are the result of a bug in handling imports introduced with the recent changes to the environment - there are a bunch of 'ambiguous reference to attribute' errors for the same attribute, even though there are no duplicate imports.

tedinski:
Those are errors from a very old bug, because `silver:definition:core` is now exporting `silver:langutil` so the explicit import is a duplication. But that project still needs changes, because it explicitly referenced `Message` and it is using `pp` instead of `unparse` on silver.

If this looks good to you, I'll go merge everything and shepard it through jenkins: https://github.com/melt-umn/silver-ableC/compare/cleanup/ppuse

krame505:
Ah, I missed the export.  Why exactly is `silver:langutil` being exported by `silver:definition:core`?  We don't do this in other places that we use langutil (e.g. ableC.)  

Aside from that, this and the changes to silver-ableC LGTM.

tedinski:
Literally just that I didn't want to have to edit all those grammars to add another import. There were a **lot**. I'm lazy.

PR #264, opened 2018-08-09
Refactoring of ide code

### LINKED ISSUE #81: Restructuring IDE features

tedinski:
1. This takes advantage of #263 to move the imp ide code over to using langutil `Message` and deletes the old `IdeMessage` type that's no longer necessary.
2. Next, I complete some work that @TravisCarlson started ages ago, and actually change the colorer over to using the lexer class information that's now on `Terminal` to decide colors.

In pursuing this second one, I've developed a strategy I plan to continue with as far as I can take it: Instead of generating java code, we'll generate information in `plugin.xml` that the Java code from the silver-eclipse runtime (`runtime/imp/`) can just use instead.

Here, instead of generating a colorer, we just generate the mapping of "font->colors" and "lexer classes->fonts" and then the runtime can just do that coloring. For Silver's own IDE it looks like this:

```
<extension point="org.eclipse.imp.runtime.tokenColorer">
  <tokenColorer class="edu.umn.cs.melt.ide.imp.services.Colorer" language="Silver">
    <font name="silver_extension_silverconstruction_font_escape" r="160" g="32" b="240" bold="true" italic="true" />
    <font name="silver_definition_core_font_comments" r="82" g="141" b="115" bold="false" italic="true" />
    <font name="silver_definition_core_font_literal" r="50" g="50" b="250" bold="false" italic="false" />
    <font name="silver_definition_core_font_keyword" r="123" g="0" b="82" bold="true" italic="false" />
    <font name="silver_definition_core_font_modword" r="41" g="95" b="148" bold="true" italic="false" />
    <font name="silver_definition_core_font_type" r="41" g="95" b="148" bold="false" italic="false" />
    <font name="silver_extension_doc_core_font_doc" r="82" g="100" b="166" bold="false" italic="true" />
    <font name="silver_extension_doc_core_font_doc_kwd" r="82" g="100" b="166" bold="true" italic="true" />
    <font name="silver_extension_doc_core_font_doc_id" r="82" g="100" b="166" bold="false" italic="false" />
    <coloring lexerclass="silver:definition:core:BUILTIN" font="silver_definition_core_font_keyword" />
    <coloring lexerclass="silver:definition:core:COMMENT" font="silver_definition_core_font_comments" />
    <coloring lexerclass="silver:definition:core:KEYWORD" font="silver_definition_core_font_keyword" />
    <coloring lexerclass="silver:definition:core:LITERAL" font="silver_definition_core_font_literal" />
    <coloring lexerclass="silver:definition:core:MODSTMT" font="silver_definition_core_font_modword" />
    <coloring lexerclass="silver:definition:core:SPECOP" font="silver_definition_core_font_keyword" />
    <coloring lexerclass="silver:definition:core:TYPE" font="silver_definition_core_font_type" />
    <coloring lexerclass="silver:extension:doc:core:DOC" font="silver_extension_doc_core_font_doc" />
    <coloring lexerclass="silver:extension:doc:core:DOC_ID" font="silver_extension_doc_core_font_doc_id" />
    <coloring lexerclass="silver:extension:doc:core:DOC_KWD" font="silver_extension_doc_core_font_doc_kwd" />
    <coloring lexerclass="silver:extension:silverconstruction:Escape" font="silver_extension_silverconstruction_font_escape" />
  </tokenColorer>
</extension>
```

If nothing puts up a road block, I plan to continue this strategy all the way through removing ALL generated java code from the ide. Every bit, hopefully. So all eclipse java code will live in the runtime, and what we generate would just be a nice full `plugin.xml` and a completely bog-standard jar of the silver-compiled code with nothing special about it.

Probably I'll stop there, but this should take us a lot closer to being able to actually remove the ide "extension" to silver. We're much closer to having refactored all this so that we can use **plain** silver jars with a supporting silver-imp-eclipse runtime and minor amounts of addition configuration. (Like the above list of color mappings.)

edit: oh and I think this fixes #81


krame505:
A little late, but LGTM.

PR #269, opened 2018-08-30
alter dont-translate to work in a more useful fashion

### LINKED ISSUE #248: Add an option to run Silver without generating translation
Building a jar isn't needed when we are running modular analyses, so being able to disable this phase would significantly speed up building extensions on Jenkins.  We may still want to write interface files though maybe?  

One possible idea: For now, simply add a flag to disable generating the buildfile.  The `silver` script would then check if `build.xml` exists before running `ant`.

tedinski:
We have an existing option that suppressed java translation, but it wasn't the most useful implementation. This cleans it up.

New behavior is to:

1. Check for errors.
2. Run MDA analysis, if any.
3. No java translation or compilation or jar files or interface files. Also no parser building. Wasn't sure about that last one, but it seems to be the behavior we usually want when using this option.

Here is an example of a change we need to make to take advantage of it:

https://github.com/melt-umn/ableC-condition-tables/compare/feature/notranslation

I call "NOT IT" on having to make this change to all the extensions... :P

Fixes #248

### DIFF
--- a/grammars/silver/modification/copper_mda/BuildProcess.sv
+++ b/grammars/silver/modification/copper_mda/BuildProcess.sv
@@ -12,17 +12,22 @@ top::Compilation ::= g::Grammars  _  buildGrammar::String  benv::BuildEnv
   top.postOps <- map(generateMdaSpec(g.compiledGrammars, _, benv.silverGen ++ "src/"),
     flatMap((.mdaSpecs), grammarsToTranslate));
 
-  local targets :: String = 
-    foldr(\ a::Decorated RootSpec b::String ->
-      foldr(\ c::MdaSpec d::String -> 
-        mdaBuildSpecTarget(c) ++ d,
-        "", a.mdaSpecs) ++ b,
-      "", grammarsToTranslate);
+  -- TODO: consider examining all grammars, not just grammarsToTranslate?
+  -- I believe this choice was originally because we weren't serializing MdaSpecs to
+  -- interface files, but I think we could easily start doing that new with the new serialization code?
+  local targets :: [MdaSpec] = flatMap((.mdaSpecs), grammarsToTranslate);
 
-  extraTopLevelDecls <- if length(targets) == 0 then []  
-                        else ["  <target name='copper_mda'>\n" ++ targets ++ "  </target>\n"];
-  extraGrammarsDeps <- if length(targets) == 0 then [] 
-                         else ["copper_mda"];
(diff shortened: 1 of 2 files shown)


krame505:
Fortunately, the `modular_analyses` Makefiles are all the same, so I should be able to do this with a quick script.  Also see the comments I left at https://github.com/melt-umn/ableC-condition-tables/commit/b6430f0bf888f25f65cfe760c69ea013a0189fa3, if you want I can make that change.

tedinski:
I pushed a change to that makefile to add timestamps to avoid rebuilds. Any further thoughts?

krame505:
Looks good, but just to be pedantic we have been using `.test` for the empty timestamp files (also this extension is listed in the `.gitignore`.)

krame505:
Hmm, this actually doesn't work for me.  I tried running `make mwda` for ableC-condition-tables and got
```
(silver.composed.Default.Pmain in Pmain.java:104): Error while evaluating function silver:composed:Default:main
(silver.driver.PcmdLineRun in PcmdLineRun.java:113): Error while evaluating function silver:driver:cmdLineRun
(silver.driver.PperformActions in PperformActions.java:101): Error while evaluating function silver:driver:performActions
(silver.driver.util.PrunAll in PrunAll.java:104): Error while evaluating function silver:driver:util:runAll
	(last line repeats 5 more times)
(core.PunsafeTrace in PunsafeTrace.java:102): Error while evaluating function core:unsafeTrace
(DN.315): While evaling syn 'core:io' in 'silver:modification:copper:parserSpecUnit' (495f3965)
(DN.254): While evaling local 'silver:modification:copper:parserSpecUnit:local:doWR' in 'silver:modification:copper:parserSpecUnit' (495f3965)
(core.PwriteFile in PwriteFile.java:111): Error while evaluating function core:writeFile
(U.371): java.io.FileNotFoundException: /home/krame505/lrk/PL_research/silver/generated/src/well_definedness/Parser_well_definedness_extendedParser.copper (No such file or directory)
(java.io.FileOutputStream in FileOutputStream.java:-2): /home/krame505/lrk/PL_research/silver/generated/src/well_definedness/Parser_well_definedness_extendedParser.copper (No such file or directory)
```

tedinski:
Yeah I just pushed a patch directly to try to fix that.

krame505:
Works for me, I'll go ahead and update all the other extensions.

PR #273, opened 2018-09-20
Introduce SILVER_HOST_GEN

### LINKED ISSUE #270: Finding pre-built artifacts
Right now, in Jenkins, we `cp -r` a `generated` folder from one job to the next. These folders are very large, and this creates a lot of disk IO. A lot a lot. 230 MB and 43k files.

This task is necessary because it's faster than re-generating all that data, and because Silver requires pre-built artifacts (svi and java files) to be in a single `generated` location. We could lift this restriction with some effort.

The goal here would be to have some flag like `-I` except instead of indicating a search path for source files (adding to the grammar path) we should be able to search for generated files at the specified location. That is, adding more locations to look for `svi` files.

We'll need this capability eventually anyway: to be able to use pre-compiled artifacts (like jars) we'd need to be able to search for generated files inside those jars. So the current logic that assumes `svi` files can only exist in `generated` will need replacing eventually.

Ultimately, I think we'll want to see `-I some.artifact.jar` as the flag to search an existing compiled jar. For a more short-term hack, we could go several different directions:

1. We could add a new flag (what?) for adding to an "artifacts search path" (Anyone got a better name?) and probably an environment variable, too. But this would be temporary, because presumably this would be done via jars in the future?
2. We could stick with `-I` and just automatically search all these paths for `svi` files. This wouldn't hurt anything really, and it would work.

I'm leaning towards 2.

Then we can change the jenkins builds for extension to just re-use ablec generated, instead of copying it.

tedinski:
This introduces an environment variable which can be directed towards a previous Silver's build `generated` directory, allowing the Silver compiler to discover already compiled files, without having to copy them into the current build's `generated` directory. The goal of this is to reduce file IO on Jenkins.

Fixes #270 

This branch:

1. Adds the `SILVER_HOST_GEN` environment variable, and has the driver search multiple locations for `Silver.svi` files. This actually makes the code clearer from its previous version. The found location is not stored and tracked for interface files.
2. Fixes jar file construction, to source files from wherever that grammar was generated to.
3. Fixes copper file generation to *ignore* parsers found via `SILVER_HOST_GEN` since these should not have changed.
4. Does nothing to `copper_mda` since those aren't presently stored in interface files anyway, so they're already ignored.

This does not yet actually use this new feature. That will come later, by modifying `jenkins-lib`, once this is merged.


tedinski:
Last plain AbleC build was 24m32s. Using this feature reduced it to 18m46s. So AbleC integration testing runtime reduced by about 20%. Not quite as much as I was hoping, but still decent.

PR #274, opened 2018-09-22
Add conditional statements to action blocks

krame505:
This adds if/else and { } statements to action blocks, inspired by C syntax (which fits fairly well here.)  These are basically pieces of imperative code, and this will make things a bit cleaner for some of the more complex tasks we are planning for modifying the parser context when parsing ableC.  I suppose we could also in theory add while loops, but I don't see that as generally useful.  
Currently we have the syntax `pushToken(...) if (...);` to conditionally push a token, this lets us get rid of that production and just do `if (...) pushToken(...);` instead.  We are also thinking of adding `include <lexerclass>;` and `exclude <lexerclass>;` action statements, this allows us to avoid adding conditional forms of those as well.  
Thoughts?  Also @tedinski I'm not sure if I got the substitution threading right in the new productions, could you take a look?

### DIFF
--- a/grammars/silver/modification/copper/ProductionStmt.sv
+++ b/grammars/silver/modification/copper/ProductionStmt.sv
@@ -126,7 +126,8 @@ top::ProductionStmt ::= '{' stmts::ProductionStmts '}'
   top.translation = stmts.translation;
   
   stmts.downSubst = top.downSubst;
-  top.upSubst = top.downSubst;
+  top.upSubst = error("Shouldn't ever be needed anywhere. (Should only ever be fed back here as top.finalSubst)");
+  -- Of course, this means do not use top.finalSubst here!
 }
 
 concrete production ifElseStmt
@@ -147,10 +148,10 @@ top::ProductionStmt ::= 'if' '(' condition::Expr ')' th::ProductionStmt 'else' e
   errCheck1.downSubst = condition.upSubst;
   top.upSubst = errCheck1.upSubst;
   
-  th.downSubst = top.finalSubst;
+  th.downSubst = top.downSubst;
   th.finalSubst = th.upSubst;
   
-  el.downSubst = top.finalSubst;
+  el.downSubst = top.downSubst;
   el.finalSubst = el.upSubst;
 
   errCheck1 = check(condition.typerep, boolType());
--- a/grammars/silver/modification/copper/ProductionStmt.sv
+++ b/grammars/silver/modification/copper/ProductionStmt.sv
@@ -128,11 +128,61 @@ top::ProductionStmt ::= 'pushToken' '(' val::QName ',' lexeme::Expr ')' 'if' con
   errCheck2 = check(condition.typerep, boolType());
   top.errors <-
        if errCheck2.typeerror
-       then [err(condition.location, "pushToken condition has type " ++ errCheck1.leftpp ++ " which is not a Boolean")]
+       then [err(condition.location, "pushToken condition has type " ++ errCheck2.leftpp ++ " which is not a Boolean")]
        else [];
 }
 
+concrete production blockStmt
+top::ProductionStmt ::= '{' stmts::ProductionStmts '}'
+{
+  top.unparse = "\t{\n" ++ stmts.unparse ++ "\n\t}";
+  top.errors := stmts.errors ++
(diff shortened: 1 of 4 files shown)


grammars/silver/modification/copper/ProductionStmt.sv
```diff
@@ -93,46 +93,79 @@ top::ProductionStmt ::= val::Decorated QName  e::Expr
...
+
+  condition.downSubst = top.downSubst;
+  errCheck1.downSubst = condition.upSubst;
+  top.upSubst = errCheck1.upSubst;
+  
+  th.downSubst = top.finalSubst;
+  th.finalSubst = th.upSubst;
+  
+  el.downSubst = top.finalSubst;
+  el.finalSubst = el.upSubst;
```
tedinski: All this needs help. You shouldn't be turning `final` into `down`.

This should just thread through `condition`, `errCheck1`, `th`, `el` and out.

tedinski:
Ok, so it turns out there are some comments:

```
{--
 - These need an initial state only due to aspects (I think? maybe not. Investigate someday.)
 - They otherwise confine their contexts to each individual Stmt.
 -}
attribute downSubst occurs on ProductionBody, ProductionStmts;
```

and

```
{- A NOTE ABOUT THREADING:
   The pair of attributes (upSubst/downSubst) represent the state of unification.
   downSubst tells a node its current state, and upSubst represents the final state.
   finalSubst gives information from elsewhere in the tree.
   
   The scope of that threading is not universal.
   That is, the threading BEGINS generally just above expressions, or individual
   production STATEMENTS. Not Production bodies.
   
   There is a slight hack to this... because aspects need to do type unification
   (actually, not sure if we need to preserve this, but let's assume we do!)
   ProductionBody and ProductionStmts (plural, *NOT* ProductionStmt)
   do not have an upSubst and finalSubst. Only downSubst.
   These will take that initial state, but otherwise confine things to each Stmt.
   -}
```

So:

* `downSubst` on ProductionStmt, ProductionStmts, ProductionBody are all really more like `initialUnificationContext`. It's not threading like it does elsewhere.
* Each ProductionStmt does have a `finalSubst` but this is a convenience: we just set `child.finalSubst = child.upSubst` from the couple of parent productions. This avoids having to explicitly set finalSubst on each child of a ProductionStmt production.

So here's the bit of code to understand to get ProductionStmt threading:

```
aspect production productionStmtsSnoc
top::ProductionStmts ::= h::ProductionStmts t::ProductionStmt
{
  h.downSubst = top.downSubst;

  t.downSubst = top.downSubst;
  t.finalSubst = t.upSubst;
}

aspect production productionStmtAppend
top::ProductionStmt ::= h::ProductionStmt t::ProductionStmt
{
  -- We treat this as though each is independent here as well.
  h.downSubst = top.downSubst;
  h.finalSubst = h.upSubst;

  t.downSubst = top.downSubst;
  t.finalSubst = t.upSubst;
  
  top.upSubst = error("Shouldn't ever be needed anywhere. (Should only ever be fed back here as top.finalSubst)");
  -- Of course, this means do not use top.finalSubst here!
}
```

grammars/silver/modification/copper/ProductionStmt.sv
```diff
@@ -93,46 +93,79 @@ top::ProductionStmt ::= val::Decorated QName  e::Expr
...
+{
+  top.unparse = "\t{\n" ++ stmts.unparse ++ "\n\t}";
+  top.errors := stmts.errors ++
+               (if !top.frame.permitActions
+                then [err(top.location, "Block statement is only permitted in action blocks")]
+                else []);
+  top.translation = stmts.translation;
+  
+  stmts.downSubst = top.downSubst;
+  top.upSubst = top.downSubst;
```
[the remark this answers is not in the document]
tedinski: Ok. Following the append production I mentioned below, probably our `upSubst` should be an `error` here, like it is there. These are similar situations.

grammars/silver/modification/copper/ProductionStmt.sv
tedinski: Ok. Following append below, the only change to make here I think is to send `top.downSubst` down to `th` and `el`, not `finalSubst`.

PR #275, opened 2018-09-22
Set SILVER_GEN build parameter

krame505:
(Also see corresponding changes in silver-ableC and jenkins-lib.)
This uses the new SILVER_HOST_GEN feature to somewhat speed up the build of the silver-ableC extension.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -60,8 +60,8 @@ melt.trynode('silver') {
     }
     // Build
     sh "./deep-rebuild"
-    // Clean
-    sh "./deep-clean -delete all"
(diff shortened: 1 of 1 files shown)


tedinski:
This probably doesn't do anything, since it's `deep-cleaning` away all the generated files.

krame505:
Good catch, fixed.  
But for that matter it also doesn't seem to be reusing ableC generated files either - did you ever test this for multiple items in SILVER_HOST_GEN?

tedinski:
Hm, maybe not? I could look into it next thursday.

PR #293, opened 2018-12-13
Refactor flow analysis organization

tedinski:
This doesn't really change behavior at all, I just tried to rename and reorganize the flow analysis code a bit, to hopefully make it a little bit more accessible to "people other than me" in the future.


tedinski:
The build failure is caused by my renaming a grammar, which broke silver-ableC. I'm going to merge this and hot-fix silver-ableC, since it's a simple rename.

krame505:
Sounds good to me.

Issue #276, opened 2018-09-27
Silver (compiler) performance meta-bug


tedinski:
There's a bunch of different categories of things that could use improvement. This is just a thought-collection list.

## Startup time

I'm retiring issue #30 and reformatting those ideas here:

* Many classes (e.g. productions) have information stored in static lists. This information gets filled in at class-load time. However, SOME of this information is not necessary until (e.g.) an error has occurred and stack traces are being printed. We could add a `debuggingInit` function, and move this static list initialization step until later, when it's needed. (Major culprit: names of attributes.) (Alternative possibility: In future JVMs, class files may gain the ability to store "lists of strings" as constants. So if we just do nothing long enough...) _(This fix already is done for terminal lexer classes. This is initialized to null, and then an accessor can fill it in later, in the unlikely event it's needed. Perhaps we could do the same for things on productions, functions, and nonterminals.)_
* Autocopy requires dynamically examining the types of every production at start-up time to fill in copy equations. This is awful. I believe I had some notes about doing this statically in my PhD thesis, as part of verifying the MWDA was sound. Perhaps we could begin enforcing those rules, and doing it statically instead.

## Silver in general

* We could probably pretty desperately use arrays in many locations we currently use lists. Arrays aren't automatically better than lists, but they'll be good probably for anything that gets iterated over more than once. Adding arrays (and good array syntax) to Silver would probably require a fair bit of work, though.
* We need strictness analysis. Even a naive simple approach (i.e. non-dataflow) would be helpful. Perhaps even JUST being able to annotate strictness on critical functions/productions in the standard library.

## Interface files

* See issue #260 
* Copper parsing (or rather, probably lexing?) performance could possible use some improvement? These files take a lot longer to parse than the old ones, and likely because it's just injesting 10 times as many bytes, since the new format is less terse.
* I haven't put Copper performance on this list otherwise, but that's something to consider, too.

## Compilation process

* Silver builds "all-at-once" instead of building "strongly connected component at a time, in topological sort order." This means we use significantly more memory than we could otherwise manage.

## MWDA

* Algorithm is very slow. Likely would benefit from arrays, other small Java-native enhancements.
* More investigation is needed.
* Would also benefit from "strongly connected component" build order: flow inference can be done on smaller parts at a time.

## Type checking

* Most of the silver error checking process is, **presumably**, type checking. So it must be slow.
* Once possible culprit is the inefficient way we do unification and track unification substitutions.
* More investigation is needed.

## Generated files

[...]

PR #295, opened 2018-12-15
Remove unneeded failing reify operations from silverconstruction extension

krame505:
This should fix the reification exceptions being caught during Silver builds that were noted in #276.

### DIFF
--- a/runtime/java/src/common/Util.java
+++ b/runtime/java/src/common/Util.java
@@ -354,17 +354,13 @@ public final class Util {
 	 * <p>Avoids demanding a StringCatter.
 	 *
 	 * @param file The filename
-	 * @param content Either a String or {@link StringCatter} object.
+	 * @param content A {@link StringCatter} object.
 	 * @return singleton IO token.
 	 */
 	public static IOToken writeFile(String file, Object content) {
 		try {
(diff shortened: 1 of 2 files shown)


tedinski:
Just copying the explanation over from the issue:

>  In ableC there are a few productions on nonterminals lacking a location annotation that have Location as the rightmost child, so I made this into a general check that works by attempting to reify the last child and failing if it isn't actually a Location. This got copied as part of the silver construction extension, but seeing as there aren't any such productions in Silver's abstract syntax, this check can be removed.

This makes sense. Possibly in AbleC we might even want to change that code so it tries to pattern match on the `AST` to see if it's a location, instead of using the reify code to check? (This might narrow its logic to "only use that location if it's actually a `loc` but perhaps that's acceptable or even better? I didn't investigate why this logic exists really...)

krame505:
> This might narrow its logic to "only use that location if it's actually a loc but perhaps that's acceptable or even better?

Yeah, that's the main reason - we have `Location` productions other than `loc`.  But I don't think it is really that significant of a performance impact, since if it is going to fail it will do so immediately when it sees a top production that isn't a `Location` production.

PR #302, opened 2019-01-21
Implement parser attribute aspects

### LINKED ISSUE #300: Parser collection attributes
One idea related to the design of disambiguation classes (#193) is the ability to contribute initial values to the "context" parser attribute, for extensions that wish to introduce particular names as marking terminals without actually making them keywords and resorting to lexical precedence.  To do this, we need something like a collection attribute containing bindings of strings to terminal identifiers.  However, normal collection production/function attributes aren't possible here, since terminal identifier literals can only occur in Copper semantic action blocks.  
One option is to add a global "initial parser action" declaration, containing action code to be run after all collection attributes have been initialized but before beginning parsing, for example
```
initial parser action {
  context = addIdentsToContext(..., context);
};
```
This would be the most general approach and easiest to implement in Silver, but would require a change to Copper as there is currently no way of specifying code to be executed after parser attribute initialization.  

Another option is to allow additional action code to be "aspected" on to an existing parser attribute, so that all "contributions" would end up in that parser attribute's initialization block:
```
aspect parser attribute context {
  context = addIdentsToContext(..., context);
};
```
This means that the values of other parser attributes can't be reliably used in contributing values to the collection, but this isn't much of a loss, since with global initial parser actions only non-collection parser attributes could be reliably used anyway.  

Thoughts?

krame505:
Fixes #300.  

This seems to work as intended when used with `context` to eliminate keywords from the string extension.

### DIFF
--- a/grammars/silver/definition/concrete_syntax/ast/CstAst.sv
+++ b/grammars/silver/definition/concrete_syntax/ast/CstAst.sv
@@ -25,6 +25,7 @@ top::SyntaxRoot ::= parsername::String  startnt::String  s::Syntax  terminalPref
   s.containingGrammar = "host";
   s.univLayout = error("TODO: make this environment not be decorated?"); -- TODO
   s.classTerminals = error("TODO: shouldn't by necessary to normalize"); -- TODO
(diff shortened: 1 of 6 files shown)


krame505:
@tedinski Would you mind hitting "approve" for this?

PR #313, opened 2019-03-28
Move host embedding translation code to a library

krame505:
Previously this code was duplicated in silver-ableC and silver construction extensions; since any host language can utilize this in specifying a corresponding silver extension, it really belongs in Silver as a library.  

Various escape productions (and their translation) are specified by collection production attributes on `nonterminalAST`.  Since these are distinguished by their full grammar name, there is no issue of conflicts between escape productions specified by different extensions with the same name.  

This branch is currently failing due to ongoing issues with Jenkins, but it should be ready to merge otherwise.


tedinski:
Oh, hmm. After a quick scan, the only issue I take with this is that it doesn't really belong in `langutil`. (That supposed to be a library where there should be no dependencies on the Silver compiler internals.) But putting it in an actually good place would require actually fixing the grammar structure of the Silver compiler.

Do you think just `silver:hostEmbedding` would be fine for now? In some future refactoring it could move to `silver:compiler:hostEmbedding` or something like that...

krame505:
My thinking was that this is something that various host languages would use in a somewhat standard way, for building language-specific extensions to Silver, thus this is code intended to be used outside of the Silver compiler.  And this grammar isn't being exported by `silver:langutil`.  
But if you think that's still an issue, I suppose moving it would be fine.

tedinski:
If you prefer to leave it where it is, it can be moved from here to a more appropriate place just as easily later as now. I just wanted to post that thought. I guess the easiest thing to do is just tag issue #152 here, where this kind of thing was mentioned before.

krame505:
I suppose there isn't much harm in moving it now, as I guess some people might find this confusing - I'll go and do that.  
@ericvanwyk can you please take a look and review this soon?  Thanks.

PR #315, opened 2019-04-04
Cleanup builtins and reduce poor error messages

### LINKED ISSUE #242: Built-in functions error when a parameter has an error
For example, I managed to get the error
`error: Operand to length is not compatible. It is of type <err>`
I haven't tried any other built-in functions, but presumably others have this issue as well.

tedinski:
Fixes #242 

I started out just cleaning up what I found in built-in functions, but then ended up with a messy patch that also significantly reduces the amount of error messages we get. My test bit of bad code went from:

```
ASDF.sv:2:28: error: Operands to ++ must be the same type. Instead they are [Integer] and Integer
ASDF.sv:2:28: error: Argument 2 of function 'core:append' expected [Integer] but argument is of type Integer
ASDF.sv:2:28: error: Operands to ++ must be the same type. Instead they are [Integer] and Integer
ASDF.sv:2:28: error: Argument 2 of function 'core:append' expected [Integer] but argument is of type Integer
ASDF.sv:4:29: error: Operand to new must be a decorated nonterminal.  Instead it is of type <err>
ASDF.sv:4:33: error: Undeclared value 'asdkjfs'.
ASDF.sv:6:31: error: First operand to 'terminal(type,lexeme,location)' must be a Terminal, instead it is <err>
ASDF.sv:6:31: error: Undeclared type 'ASDFASDASDF'.
ASDF.sv:8:22: error: Integer is not a concatenable type.
```

to:

```
ASDF.sv:2:28: error: Operands to ++ must be the same concatenable type. Instead they are [Integer] and Integer
ASDF.sv:4:33: error: Undeclared value 'asdkjfs'.
ASDF.sv:6:31: error: Undeclared type 'ASDFASDASDF'.
ASDF.sv:8:22: error: Integer is not a concatenable type.
```

### DIFF
--- a/grammars/silver/definition/core/BuiltInFunctions.sv
+++ b/grammars/silver/definition/core/BuiltInFunctions.sv
@@ -1,15 +1,11 @@
 grammar silver:definition:core;
 
---import silver:analysis:typechecking:core;
-
 concrete production lengthFunction
 top::Expr ::= 'length' '(' e::Expr ')'
 {
   top.unparse = "length(" ++ e.unparse ++ ")";
 
-  top.typerep = intType();
-  
-  top.errors := e.errors ++ forward.errors;
+  top.typerep = intType(); -- is this necessary? for flowtype reasons?
 
   forwards to performSubstitution(e.typerep, e.upSubst).lengthDispatcher(e, top.location);
 }
@@ -21,7 +17,13 @@ top::Expr ::= e::Decorated Expr
 
   top.typerep = intType();
 
-  top.errors := [err(e.location, "Operand to length is not compatible. It is of type " ++ prettyType(performSubstitution(e.typerep, top.finalSubst)))];
+  local resolved :: Type =
+    performSubstitution(e.typerep, top.finalSubst);
+
+  top.errors :=
+    e.errors ++
+    if resolved.isError then [] -- suppress additional error message
+    else [err(e.location, "Operand to length is not compatible. It is of type " ++ prettyType(resolved))];
 }
 
 abstract production stringLength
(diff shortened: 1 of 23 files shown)


grammars/silver/definition/core/BuiltInFunctions.sv
```diff
@@ -106,61 +105,89 @@ top::Expr ::= 'terminal' '(' t::TypeExpr ',' es::Expr ',' el::Expr ')'
...
-      stringConst(terminal(String_t, "\"??\""), location=$6.location),
-      intConst(terminal(Int_t, "-1"), location=$6.location),
-      intConst(terminal(Int_t, "-1"), location=$6.location),
-      intConst(terminal(Int_t, "-1"), location=$6.location),
-      intConst(terminal(Int_t, "-1"), location=$6.location),
-      intConst(terminal(Int_t, "-1"), location=$6.location),
-      intConst(terminal(Int_t, "-1"), location=$6.location)
-    ]), $6, location=top.location);
+
+  local bogus :: Expr =
```
krame505: Idea: what if instead of a bogus location here we put the location of this production, so the terminal location would point to the actual location in the Silver source file?

tedinski: Well, it'd be confusing if you compile something with AbleC and get an error message pointing to an `sv` file that's nowhere to be found! :)

It's a good idea to better track where things originate, but I'd file it under general ideas for improving how we track locations. Less a change we can make here, more a change we could make to location information, then modify something like this to use.

krame505: We already sort of have this with the silver-ableC host embedding stuff, since an ableC tree constructed from an AST literal has the locations in the Silver source file as its location.  And getting such a built-in location in an error message always points to a compiler bug, but its way nicer for debugging then just getting "builtin: line -1, column -1" in an error.  
But yeah that is something we should discuss more, I agree this doesn't need to be done now.

PR #319, opened 2019-04-11
Towards some more better error messages

### LINKED ISSUE #307: Missing environment information could be better diagnosed
I recently got this:

```
Errors for silver:extension:testing
 [/home/tedinski/repos/silver/grammars/silver/extension/testing/EqualityTest.sv]
EqualityTest.sv:67:95: error: Attribute 'grammarFlowTypes' does not occur on 'Decorated silver:driver:util:RootSpec'. Looked at:
	silver:definition:env:grammarFlowTypes (Attributes.sv:82)
EqualityTest.sv:68:103: error: Attribute 'productionFlowGraphs' does not occur on 'Decorated silver:driver:util:RootSpec'. Looked at:
	silver:definition:env:productionFlowGraphs (Attributes.sv:81)
```

Technically, this error is correct, but the root cause was `silver:driver:util` not being imported, which is where `RootSpec` and its occurrences are declared.

Since `nonterminal Foo with ...` is a common case, I think if we run into an error involving a type from a grammar that's not imported, perhaps we should point that out as part of the error message.

tedinski:
This fixes #307 by introducing a new heuristic error message.

I also started looking at #33, but ended up just doing some clean-ups, and not getting to the issue. I don't know when I'll get back to this, so I'm stopping here.

### DIFF
--- a/grammars/silver/definition/core/Expr.sv
+++ b/grammars/silver/definition/core/Expr.sv
@@ -1057,20 +1057,29 @@ Boolean ::= l::Pair<Integer Decorated Expr>  r::Pair<Integer Decorated Expr> { r
 function mkStrFunctionInvocation
 Expr ::= l::Location  e::String  es::[Expr]
 {
-  return mkFunctionInvocation(l, baseExpr(qName(l, e), location=l), es);
+  return mkFullFunctionInvocation(l, baseExpr(qName(l, e), location=l), es, []);
 }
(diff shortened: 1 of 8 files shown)


krame505:
Looks like something broke in silver-ableC that needs updating now.

krame505:
Should be fixed, so I'm going to go ahead and merge this now.

PR #320, opened 2019-04-15
Implement silver extension providing nicer syntax for constructing AST trees

krame505:
This re-uses the reflection library's `AST` deserialization grammar to provide another "tree literal" convenience extension for the `AST` type, analogous to the silver-ableC and silverconstruction extensions.  
These changes are taken from the feature/astTerm branch, for which we are deferring further work until later.


krame505:
@ericvanwyk Could you review this when you get a chance?  Thanks.

PR #321, opened 2019-04-20
Add support for patterns using language-specific syntax

krame505:
Using extensions such as silver-ableC, we can nicely construct trees using expressions that embed the syntax of another language.  However, pattern matching on such trees is still annoying.  We can create extensions that use language-specific syntax to specify patterns in a similar way to those that create expressions, by just computing another translation on `AST`.  

This was a quick test of the idea, I added new pattern syntax for the `AST` type itself, and used that to simplify some of the annoying patterns in silver-ableC.  
I also added patterns for matching `Float`s as it seems Silver was lacking those for some reason?  

One slight issue is that we don't have a way of matching on terminals at the moment, so currently attempting to do so just results in a wildcard pattern instead.  I'm not sure of the best way to deal with this case, besides adding support for terminal patterns, I guess.  Shouldn't really be an issue if we use this for languages with abstract syntax (like ableC), but for a language like Silver unexpected behavior might occur if we try to match using a pattern containing identifier terminals.

### DIFF
--- a/grammars/silver/definition/flow/env/Expr.sv
+++ b/grammars/silver/definition/flow/env/Expr.sv
@@ -756,6 +756,12 @@ top::PrimPattern ::= i::Int_t '->' e::Expr
   top.flowDeps = e.flowDeps;
   top.flowDefs = e.flowDefs;
 }
+aspect production floatPattern
+top::PrimPattern ::= f::Float_t '->' e::Expr
+{
+  top.flowDeps = e.flowDeps;
+  top.flowDefs = e.flowDefs;
+}
 aspect production stringPattern
 top::PrimPattern ::= i::String_t '->' e::Expr
 {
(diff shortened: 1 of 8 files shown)


tedinski:
> I also added patterns for matching Floats as it seems Silver was lacking those for some reason?

Equality on floats is generally a bad idea, that's the only reason it didn't exist.

I endorse introducing terminal patterns, if you want to. For the moment, it should probably just mimic the terminal creation syntax i.e. `terminal(Term_t, lexeme, location)`.

Also, can you give an example of what you expect this to look like when used? Just curious.

krame505:
After a brief chat, pattern matching on `Float`s isn't any more of an issue that having `==` work for `Float`s.  We just don't really want (or have a need) to use it, generally.  

Terminal patterns would make this nice but I don't have time to implement this at the moment - I'll open an issue if there isn't one already.  Just a note for the future though is that when using terminal patterns in the translation for this we would want to wildcard the location rather than using the location we have from the concrete pattern, similarly to how we discard annotations for generating production patterns.  

One example of how we might use this is for ableC, we might have an extension production that wraps a host `Stmt` and wants to perform some interesting structured analysis, such as checking if the `Stmt` is a for loop and extracting the components.  This gets kind of messy using abstract syntax so having concrete syntax for patterns would be useful here.

Issue #304, opened 2019-01-28
No type checking for pluck return type


krame505:
Currently a value of any type may be plucked.  This is an error, as only terminals in disambiguation functions and terminal identifiers in disambiguation classes should be pluckable.

602p:
The easy part of this is enforcing that the type passed to pluck is a terminal identifier (I have a patch that does this that I'll PR shortly.) This solves the nasty cases that `pluck 3` would compile, and that `pluck "foo"` would only get caught by javac.

The harder part is making sure that it's one of the terminals being disambiguated amongst - that might deserve it's own issue, or could co-opt this one. I'm not sure what a good solution to that is.

There's a hacky approach I found suggested in a comment by @tedinski , which is to bind over the names of the legal terminal identifiers with some marking type, and enforce that type when typechecking `pluck`. I tried implementing this using a `newSkolemConstant()` per action block as the marker type. This seems to work-ish for the restricted case of disambiguation functions. It's a hacky solution though, and I don't think it solves the more challenging case when disambiguation functions pluck terminal identifiers from parser attributes.

In that case we need to enforce that all assignments to that attribute are legal terminals to be plucked in the disambiguation function. Additionally, it is currently legal to have multiple disambiguation functions pluck from the same parser attribute, since as long as the attribute only ever holds terminals in the intersection of the groups being disambiguated this works fine.

As a short term change, we might consider adding runtime checks to make sure the returned terminal identifier is amongst the legal ones.

krame505:
That is more or less the situation as it stands.  We do have runtime checking (in Copper) for disambiguation functions and classes to ensure that the plucked terminal is currently in the set of shiftable terminals - failing this currently raises a syntax error, I think.  But ensuring that the plucked terminal of a disambiguation class is actually in the class is a tough problem using types - really the only nice way of handling this (and the similar case of disambiguation functions) is to have type classes, with an type class automatically defined for each disambiguation function/class.  
For now I would recommend only checking that what is `pluck`ed is a `TerminalId`, and leave function/class membership to be checked at runtime, until we have added type classes to Silver.

602p:
I was poking around in Copper, and can't find such a check. The silver code
```
terminal Id_t /[a-zA-Z][a-zA-Z0-9]*/;
terminal IntLit_t /[0-9]+/;
terminal If_t /if/;
ignore terminal WhiteSpace_t /[\t\r\n\ ]+/  ;

disambiguate Id_t, If_t {
[...]

PR #325, opened 2019-05-28
Fix some issues with `pluck` in disambiguation functions and groups

### LINKED ISSUE #305: Code after pluck in action block yields unreachable code
Currently doing anything after `pluck` in an action block will generate a return statement followed by some other code, which is illegal in Java.  Somebody should either make this an error in Silver or just change the generated code for pluck to `if (true) return ...;` to suppress this error.

### LINKED ISSUE #306: Missing pluck in disambiguation function/class action block generates non-compiling code
Not writing `pluck` in a place where it is expected results in code that doesn't compile due to a missing return statement.  Also see #305.

602p:
Fixes #305 , fixes #306 , fixes the "easy" part of #304 .

### DIFF
--- a/grammars/silver/modification/copper/ActionCode.sv
+++ b/grammars/silver/modification/copper/ActionCode.sv
@@ -48,10 +48,36 @@ top::ActionCode_c ::= '{' stmts::ProductionStmts '}'
   top.actionCode = sflatMap(hacklocaldeclarations, stmts.defs) ++ stmts.translation;
 
   top.errors := stmts.errors;
+  top.errors <- if top.frame.permitPluck && !stmts.containsPluck then
+    [err(top.location, "Disambiguation function without pluck")] else [];
   
   stmts.downSubst = emptySubst();
 }
 
+
+-- Support code to check the validity of disambiguation blocks. True if any elements
+-- contained in the snoc-list (so this statement or before) are a pluck. Handles
+-- raising errors if there are statements after a pluck.
+synthesized attribute containsPluck :: Boolean occurs on ProductionStmts;
+
+aspect production productionStmtsSnoc
+top::ProductionStmts ::= h::ProductionStmts t::ProductionStmt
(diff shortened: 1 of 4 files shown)


grammars/silver/modification/copper/ActionCode.sv
```diff
@@ -48,11 +48,37 @@ top::ActionCode_c ::= '{' stmts::ProductionStmts '}'
...
+-- Support code to check the validity of disambiguation blocks. True if any elements
+-- contained in the snoc-list (so this statement or before) are a pluck. Handles
+-- raising errors if there are statements after a pluck.
+synthesized attribute containsPluck :: Boolean occurs on ProductionStmts;
+
+aspect production productionStmtsSnoc
+top::ProductionStmts ::= h::ProductionStmts t::ProductionStmt
+{
+  local immediateChildIsPluck :: Boolean = case t of
+    | pluckDef(_, _, _) -> true
```
krame505: Minor formatting note: `case` should be on a new line, with the same indentation as `end` and all the `|` clauses.

grammars/silver/modification/copper/ProductionStmt.sv
```diff
@@ -28,7 +28,15 @@ top::ProductionStmt ::= 'pluck' e::Expr ';'
                else [])
                ++ e.errors;
 
-  -- TODO: figure out wtf is going on with type here! (needs to be a terminal, plus one of the ones in the disgroup)
+  local tyCk :: TypeCheck = check(e.typerep, terminalIdType());
+  tyCk.downSubst = e.upSubst;
+  top.errors <- if tyCk.typeerror
```
krame505: Minor formatting note: `if` should be indented the same as `then` and `else`.

602p:
Closed this because I am re-PRing from a branch on this repo instead of from my fork so we get CI.

PR #328, opened 2019-07-17
When clauses for pattern matching

RandomActsOfGrammar:
This adds `when` clauses to pattern matching with `case`, allowing Boolean conditions to be added to match clauses.  This enables something like the following code to be written:
```
case x of
| bar(_, _) -> 5
| foo(a,b) when a > b -> a - b
| foo(a,b) when b > a -> b - a
| _ -> 0
end
```

### DIFF
--- a/grammars/silver/extension/patternmatching/Case.sv
+++ b/grammars/silver/extension/patternmatching/Case.sv
@@ -13,6 +13,7 @@ terminal Of_kwd 'of' lexer classes {KEYWORD,RESERVED};
 terminal Arrow_kwd '->' lexer classes {SPECOP};
 terminal Vbar_kwd '|' lexer classes {SPECOP};
 terminal Opt_Vbar_t /\|?/ lexer classes {SPECOP}; -- optional Coq-style vbar.
+terminal When_kwd 'when' lexer classes {KEYWORD,RESERVED};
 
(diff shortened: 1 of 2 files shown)


krame505:
FYI Jenkins is still slightly broken, so you'll need to trigger a build manually before github will let you merge.

RandomActsOfGrammar:
How do you trigger a build manually?

krame505:
I don't have access at the moment, but if you go to the Jenkins console at
http://coldpress.cs.umn.edu:8080/, you should be able to navigate melt-umn
-> silver -> feature/when-clauses where there should be an option to start
a build.  You will need to be logged in to Jenkins for that option to show
up (did we ever set up an account for you?)
Lucas Kramer

RandomActsOfGrammar:
No, I don't have a Jenkins account that I know of.

tedinski:
I might come in today, to take a look.

How’d you decide on the ‘when’ keyword?

Ted

RandomActsOfGrammar:
OCaml, F#, and C# all use `when` for conditions on patterns.

krame505:
The only language I am aware of that differs is Haskell, which uses `|` - but that would probably be more confusing for us as we already use it as our pattern separator.

Issue #36, opened 2015-08-20
Build fails without --clean meta-bug


GoogleCodeExporter:
This bug is intended to track the ways in which builds without using --clean can fail.

There is only one known way presently:

1. Make a change to a grammar that causes compiles to still succeed, but with different translation that matters to downstream grammars.  For example:
    * Move a function from one grammar to another. (Where existing users import both grammars.)
    * Change a production into a function.

tedinski:
Presently, our rebuild logic works like this:

1. Scan all the grammars, if any have files changed, then we re-parse the grammar. These are then translated.
2. Determine which grammars directly depended on changed grammars. These are then re-analyzed (for errors), but not re-translated.

We could solve this one known build bug by _conditionally re-translating files_. Basically, instead of deleting the old translation and re-generating, we can take each file, compare its contents with what's already there, and do nothing if it hasn't changed. This is what we do with generating parsers right now, too.

Which would be another advantage. We generate parsers in a separate way than other files because parsers need to change if one of their included grammars changes. I believe this change, in addition to fixing the one know build bug, would also let us unify how we generate parsers with all the rest of `genFiles`.

tedinski:
I think to solve this problem fully we have to give up on the idea that there's only two phases necessary. ("The changed grammars" and "the grammars that directly depend upon them.")

Instead, we should just embrace an iterative algorithm. Start with changed grammars, and see if the interface files change. If so, trigger re-building dependent grammars. Iterate until there are no changed interface files left (or we've re-translated everything.)

The note in #25 is only one way that problem can happen. Consider moving a nonterminal from one grammar to another, and other grammars having `type` declarations involving it. This, too, can mean we'd have to re-build grammars more than once-removed from the change.

tedinski:
There's a problem with builds currently where, when a grammar directly changes, its generated dirs get their files deleted. This means even if the generated parser spec doesn't change at all, it still has to get recompiled by Copper. It'd be nice to somehow preserve the `.java` file generated from the `.copper` file for this special case, to avoid that recompliation.

> instead of deleting the old translation and re-generating, we can take each file, compare its contents with what's already there, and do nothing if it hasn't changed.

This works for parsers because we know it only generates a single `.java` file translation, but it's more complicated for regular java files. We'd have to remember, when we find that `filename.java` has changed, to also delete all `filename$generated.class` files it potentially generated, so nothing stale persists.

602p:
Was going to open a new issue for this, but then found this one.

[...]

PR #333, opened 2019-10-02
Origin Tracking!

602p:
Here are some high level issues to be resolved before this can be merged for real:
 - [x] OI interaction with Reflection
   - [x] Reflective ASTs originate from objects
   - [x] Reified objects originate from ASTs
   - [x] Persist notes
   - [x] Reflective invoke
 - [x] Parser action blocks
 - [x] Future of the `location` annotation
    - [x] Enforcement of `location`-or-origins on concrete productions
 - [x] Interaction with forwarding and lambdas
    - [x] Tracking notes through lambdas
    - [x] Tracking forwarding links
 - [ ] Incremental recompilation fix -- see #36 
 - [x] Better tooling and ergonomics for client languages
    - [x] No-redex mode
    - [x] Compile-time switch
    - [x] Call-path note option
    - [x] Note syntax
    - [ ] Patching in `tracked`-ness for stdlib constructs (e.g. `Message`)
 - [x] Move `tracked` tracking into NT type proper to fix spurious `OriginInfo`s being constructed
 - [x] Performance overhead
 - [ ] `cast` expression in core silver (or other link polymorphism solution) -- moved to own branch
 - [x] Remove "bogus" fallback OI
 - [x] Bring back to parity with `develop`

Let's use this PR as a tracker for work that needs to be done to polish this up for merging.


krame505:
@602p I noticed you have added a new `cast` expression to Silver.  What exactly is the anticipated use case of this?  

Note that we already have this functionality by way of `reify(anyAST(a))`, although I'm not against making this less awkward.  

From looking at the code it appears that `cast` is also intended to auto-undecorate, which is fine, however there are a couple of issues with the implementation:
* Checking `arg instanceof ${resultType.transType}` isn't sufficient to ensure that `arg` has the expected type, as this doesn't properly handle polymorphic types such as `Maybe`.  You need to consult the run-time `TypeRep` representation of the type, which includes this information.  
* `cast` appears to always undecorate it's argument - this means that casting `Decorated Foo` to `Decorated Foo` would fail - which I believe isn't the desired behavior.

Other than that, this looks good!

602p:
@krame505 thanks for pointing out `reify(anyAST(x))`. I think that's actually a much better solution to the problem `cast` was intended for. The case at hand is that when we have some value that's been computed and has origins info attached, we'd like to be able to look at the origin of it. We don't however know the type of it's origin (the nonterminal signature for `originOriginInfo`/`originAndRedexOriginInfo` uses unconstrained tyvars like `anyAST` does) and what's more it might be some nonterminal type from an extension that the host language code can't even name. The question is how to ask a question like "did this originate from an `Expr` and if so which?"

Previously my approach was to allow the code to get a reference to the origin as an unconstrained tyvar (standing in for an `Any` type) and then try `cast`ing that to nonterminal types it knew about. That was problematic because (1) it needed `cast`, which is kinda too ugly to live IMO and (2) `a` isn't actually `Any`, and has issues if you try to use it that way.. I think a better approach is setting up the origins API in terms of `AST`s, so that `getOriginLink` returns `Maybe<AST>` instead of `Maybe<a>`, and then as an implementation detail make it so that it's always an `anyAST` so `reify` is free. I've been doing that and I'll remove `cast` once it's happy, but that seems like a much better approach - thanks for reminding me of `reify(anyAST(x))` :)

krame505:
Hmm, I'm not actually sure if I like that.  Generally an `anyAST` holds something that can't be introspected in any way (e.g. a function or foreign type), so wrapping up things that we actually should be able to pretty-print, serialize, etc. in `anyAST` is kind of a hack.  I don't really see how having `Maybe<AST>` improves anything over just using a skolem type like `Maybe<a>` - if someone wants to do a reflective operation or `reify` into some type then they can call `reflect` or `anyAST` themselves.  

Also just to be clear, `origin` and `redex` in `originOriginInfo` / `originAndRedexOriginInfo` point to the decorated or undecorated tree?  At least in `originAndRedexOriginInfo` using `a` for both `origin` and `redex` requires them to be the same type, which is nice (assuming that this is what you wanted - remind me what `redex` is, again?)  

> (1) it needed `cast`, which is kinda too ugly to live IMO

I don't think it's actually that bad, assuming the above-mentioned bugs are fixed.  If auto-undecoration isn't an important feature, `cast` could be made to forward to
```haskell
case reify(anyAST(a)) of 
| left(x) -> just(x)
| right(_) -> nothing()
end
```
instead of implementing it in the runtime.  

> and (2) `a` isn't actually `Any`, and has issues if you try to use it that way..

What exactly do you mean by this?  What is `Any`?

There may be another solution possible in the future, involving type class magic.  Maybe some sort of class `HasOrigin a b` where `a` and `b` are an instance if `a` can have `b` as its origin?  These instances could be auto-generated by the compiler.  This does require us to support multi-parameter type classes, though, and I haven't quite thought through yet how this would actually be useful in practice.

602p:
Re: the `origin` and `redex` fields: them both using the same tyvar is a bug -- fixed on my local copy now (these `*OriginInfo` objects get constructed from java and don't get checked.) Thanks. `redex` represents a node that catalyzed the movement of the node from one part of the tree to another as part of a transformation (it's just another type of origin information.) Both links (`redex` and `origin`) are undecorated in the OI nodes.

Re: cast: no, the auto-undecoration isn't important - I'd be happy to see it implemented that way, and it could be done just in the stdlib and avoid adding an expr for it. This does work, but trying to actually use it exposes some issues with `reify`... it seems like our implementation of tyvars means that functions parameterized over a tyvar can't call `reify` where the result is parameterized. For example, the following

```haskell
imports core;
imports silver:reflect;

function reify2
Either<String b> ::= x::AST
{
  return reify(x);
}

function main 
IOVal<Integer> ::= largs::[String] ioin::IO
{
  local m :: Either<String Integer> = reify2(reflect(123));
  return ioval(print(hackUnparse(m), ioin), 0);
}
```

fails because the call to `reify` can't tell what type it's supposed to return, and so tries to reify against a bare `common.BaseTypeRep("b2")` which always fails. Not sure what to make of this - seems like a bug that it's possible to write this and not have it fail until runtime; is reifying against a bare tyvar ever intended? Should it error out rather than generate that?

The case this comes up in is wanting to write helpers that look like
```haskell
function getOriginLink
Maybe<Either<String b>> ::= val::a
{
 return case getOriginInfo(val) of
     | just(originOriginInfo(_, o, _, _)) -> just(reify(anyAST(o)))
     | just(originAndRedexOriginInfo(_, o, _, _, _, _)) -> just(reify(anyAST(o)))
     | _ -> nothing()
 end;
}
```
and encountering the same issue of the doomed-to-fail reification against a tyvar. That this worked with `cast` is only because of the first bug you pointed out in it that it doesn't check that it reifies - if it does then it also always fails on trying to reify with a tyvar.

602p:
@krame505 do you have time for a call tomorrow to talk through this stuff? I'm free most of the day 9-5.

krame505:
Yeah, the issue with `reify` is a known problem but apparently this wasn't documented outside of the source code - I just opened #368.  

Also there is a slightly more detailed discussion of this in the (not yet published) journal paper on reflection - https://github.umn.edu/melt/pubs/tree/master/drafts/2020_COLA_Reflection_in_AGs, I can send you a copy if that doesn't render for you.

602p:
Code style/review questions:
 - [ ] Should `tracked`-ness be part of pretty-printing a nonterminal type? Without this errors can come up in partial-recompilation-bug situations where we report `NT expected but got NT` (because the trackedness of the interface file dosen't match) and that's very unhelpful. In `--clean` builds/if we resolve #36 this issue will never occur.
 - [ ] `attachNote` ergonomics - is the inline form useful? Should it look like a function or not?
 - [ ] `--no-origins`, `--force-origins`, `--no-redex`, `--tracing-origins` flags: very useful, acceptable? These introduce global compiler modes in a way we don't already have.
 - [ ] Things that we have to decide in the compiler if they are tracked or not: `Location`, `Message`, `Either`, IDE stuff
 - [ ] Runtime warning options - keep?
 - [ ] Are we actually using monto?

krame505:
If I understand correctly, trackedness is now a feature of the type of a nonterminal, so one could have both tracked and untracked versions of a nonterminal in the same program?  If that is the case then it should definitely be indicated in some way when the type is pretty-printed.  Writing `tracked Foo` everywhere though might get rather verbose, perhaps something like `@Foo` for tracked nonterminal types?

Also is there a way to convert between tracked and untracked values?

I have no opinion (yet) on `attatchNote`, I would recommend coming up with some examples to discuss.

Compiler flags are fine as long as they don't affect whether or not the code compiles.  Seems useful to have for measuring performance impact of origin tracking. 

I would strongly lean towards 'no' for all of those things, unless there is a reason to do otherwise - these types generally shouldn't occur directly in ASTs, and when they do they are essentially in leaf position.  

What runtime warnings?

Not really, the plan AFAIK is to get better Atom plugin generation working, and perhaps add support for building Language Server plugins.  Check with Eric, but if the monto code is a barrier to getting origins working then I would be fine with its removal.

602p:
The most recent work I've been doing has been fixing some laziness issues and tightening performance. Here are some performance numbers for ableC when using origins in different modes (measured using [hyperfine](https://github.com/sharkdp/hyperfine)):

```
HEAD 			3.060 s ±  0.155 s
OFF			3.093 s ±  0.052 s		+1% 	(w/in measurement error of HEAD)
SELECTIVE-NR		3.356 s ±  0.083 s 		+9%
SELECTIVE 		3.465 s ±  0.070 s 		+13%
ALL-NR 			3.464 s ±  0.042 s 		+13%
ALL			3.518 s ±  0.086 s 		+15%
```

`HEAD` means the ableC head built with the silver head (so no tracked nonterminals, compiler does not support origins and does not have any overhead to support it.) 

The rest are my origins-enabled branch of ableC that marks all the CST and AST non-terminals as tracked, and built with origins-enabled silver.

`OFF` means origins-enabled ableC built with origins-enabled silver with `--no-origins` passed, and serves to show the within-margin-or-error performance change if you aren't using origins in your project.

`SELECTIVE` means 'normal' mode, tracking only those nonterminals marked `tracked`.

`ALL` means passing `--force-origins` and tracking _every_ nonterminal in the compiler (including stdlib stuff, etc.) You probably won't ever do this except to experiment with what you want to track (e.g. build with `--force-origins` and look at what nonterminals appear in origins traces, then mark them `tracked` and go back to building normally.)

`-NR` as a suffix means built with `--no-redex`. This disables tracking redexes which is relativley expensive and not required for the primary use case (which is finding the source location of error messages.)

In my opinion 9% is a little steep for `SELECTIVE-NR`, which is the mode one would probably want to use if all they want is to not have to use `location`. If we started using origins in ableC we could conceivably chuck out `location`, and that might mean that the net performance decrease is lower since we no longer need to shuffle that around. That said, I think that the performance story is now thoroughly OK because the `HEAD` vs `OFF` difference is almost zero - within measurement error.

Fixing the laziness issues also was the last blocker to building silver itself to use origins! This works great and using origins-based location info for errors emitted by silver works just fine. I haven't done rigourous performance testing there, but it looks like the time for OI-enabled silver to build silver with OI is only about 3s slower than head silver to build head silver, which is less than 10%.

602p:
For completeness' sake, here are the perf stats when run against a large (7kLOC) C file without errors:

```
HEAD 			7.876 s ±  0.202 s
OFF 			8.081 s ±  0.285 s 		+2% 	(w/in measurement error)
SELECTIVE-NR 		8.470 s ±  0.294 s 		+7%
SELECTIVE 		9.204 s ±  0.458 s 		+17%
ALL-NR 			8.895 s ±  0.373 s 		+13%
ALL 			10.391 s ±  0.481 s 		+32%
```

It seems that the performance impact of selective origins is about the same, but using origins with redexes on everything must result in a lot of overhead in the last part of the translation process. Regardless I think this is still OK because the `HEAD` vs. `OFF` difference is still within the margin of error.

krame505:
Very nice, thanks for collecting some concrete performance numbers on this - certainly we can live with a ~10% performance regression, I was worried it would be much worse.  I suspect that getting rid of location annotations on everything will have a significant positive impact as well.  

If you are able I would be interested in seeing the performance numbers for building the ableC-prolog extension's test suite with origins enabled, as there is a quite large (and possibly exponential) amount of forwarding going on here.

ericvanwyk:
@krame505 - please don't be mean to @602p ... :)

602p:
@krame505 I meant those questions mostly for myself, but thanks for answering!

> If I understand correctly, trackedness is now a feature of the type of a nonterminal...

Trackedness is a feature of the nonterminal type in the same way that closedness is - that is, a type is tracked or not throughout the codebase, controlled by the `tracked` keyword where it's defined. So you don't have to write `tracked Foo` (or `@Foo`) everywhere, since you can't have both tracked and nontracked `Foo` in the same compilation. The semantics of a tracked vs non-tracked type are exactly the same except for whether `getOriginInfo(x)` will return something or not (and of course codegen for them.) Per our discussion some months ago the reason trackedness had to move into the `nonterminalType` instead of being in the `dcl` for it is because of situations where code must be generated that needs to know if a value is tracked or not in a context where the `dcl` is not in scope (e.g. importing `err` but not `Message`.)

The question of if to include it in the prettyprint or not came up when running into another instance of #36 where changing the trackedness would produce meaningless errors (`expected NT but got NT` where the first is tracked and the second is not) or runtime crashes. These build problems are still unsolved barring fixing #36 - the only sane way to build after changing the trackedness of something (including via changing compiler flags) is to build `--clean`ly. Thoughts?

> I have no opinion (yet) on `attachNote`, I would recommend coming up with some examples to discuss.

:+1:. I'll put up a copy of the docs page I'm writing for origins semantics + constructs soon.

> Compiler flags are fine as long as they don't affect whether or not the code compiles. Seems useful to have for measuring performance impact of origin tracking.

:+1:. None of these flags change the correctness of code - accessing origins results in a `Maybe<OriginInfo>` regardless.

> I would strongly lean towards 'no' for all of those things [...]

I agree, and that's how it is now, with the exception of `Message`. It's pretty handy for that to be tracked and I think it makes sense to be liberal when we have to pick behavior in the compiler since code that isn't gonna use origins can compile with `--no-origins`. That being said, the only reason we have to pick a trackedness for `Message` in the compiler instead of the stdlib is because it's used directly in `impide`... similarly to monto, is that living code still? (The only reason types need to have a trackedness picked in the compiler is when they are referred to directly as e.g. `nonterminalType("silver:langutil:Message", [], true)`. So if we can excise the mention of `Message` in impide then we don't need to pick trackedness in the compiler. That said, we do still need to pick it where it's defined in `langutil`, so it's kind of a moot point.)

> What runtime warnings?

I removed them. I had had some warnings (enabled via `-Dsilver.origins.rtwarn=1`) in the generated code for cases where FFI stuff or bugs were breaking origins behavior, but I reached a point where all the issues that were tripping them were fixed (or in the case of FFI, won'tfixed) and all they were doing is slowing it down so I removed them.

> ...if the monto code is a barrier to getting origins working then I would be fine with its removal...

The only problematic stuff is the monto interop stuff in the java runtime library. It could be set up to work with origins via the catchall FFI origin, but I just have it commented out of the build for now.

krame505:
Ah, I had forgotten about the discussion about moving trackedness into types - I still don't like it, but I don't see a better solution without changes to how we handle imports.  

Given that, trackedness definitely shouldn't be included in the pp.  In fact if there ever is a mismatch in trackness for a nonterminal I would recommend raising an internal error with a link to #36 and a suggestion to try a clean rebuild, rather than attempting to report it as a regular type error.  

One other idea here, though - we could change the semantics of imports to enforce that when a production is imported, its nonterminal is always implicitly imported (at least in its fully-qualified form) as well.  I haven't thought through too carefully how hard this would be to implement or what the other consequences might be, but it might be less annoying than putting this all in production signatures.  

The rest of your points all sound good to me.

krame505:
FWIW we will probably merge feature/strategyattr before this branch - so you could get a head start by merging feature/strategyattr into this.

602p:
Here's a quick update to the performance story. Since the headline use of origins is to replace needing to use the `location` annotation and get rid of all that swizzling, I did that to AbleC. You can find my patch on the[ `feature/nolocation` branch](https://github.com/melt-umn/ableC/tree/feature/nolocation) and it should build with the silver built from this branch. It completely removes `location` and the associated `exprLocation` and `sourceLocation` attributes on some containers and env items (tracking them instead.) It also removes `Location`s from signatures throughout the codebase (e.g. cases where they are passed into helpers etc.) All the `err`s and `wrn`s are converted to `errFromOrigin`s etc. In cases where a `Location` was passed into a helper to be the source for messages one of the other arguments' origins is used, or `ambientOrigin()` is used to retrieve the origin context of the function call. Here are the perf numbers for that:

```
errs.c (small file with errors)
HEAD			2.941 s ±  0.241 s
OFF			2.405 s ±  0.072 s -18%
SELECTIVE-NR		2.508 s ±  0.055 s -15%
SELECTIVE 		2.494 s ±  0.069 s -15%
ALL-NR 			2.573 s ±  0.095 s -13%
ALL 			2.565 s ±  0.073 s -13%

bzip2.c (very large file without errors)
HEAD			7.679 s ±  0.260 s
OFF			7.751 s ±  0.435 s +1% (w/in measurement error)
SELECTIVE-NR		8.360 s ±  0.320 s +9%
SELECTIVE 		8.619 s ±  0.308 s +12%
ALL-NR 			8.865 s ±  0.430 s +15%
ALL 			9.140 s ±  0.360 s +19%
```

`HEAD` has the same meaning as above, so that version is using location and no origins. `OFF`, `SELECTIVE-NR`, `SELECTIVE`, `ALL-NR`, and `ALL` all mean the same except that the codebase they are building has had `location` totally removed. This means that `OFF` is a much less useful metric since it won't work (all the locations will be garbage.)

krame505:
Thanks Louis, those are somewhat promising numbers.  I'm still curious about the impact of forwarding if you get around to it, though.  

Also by the way have you been triggering Jenkins manually?  I'm just curious since the webhooks are still broken AFAIK, but I've been seeing build failure notifications on this branch.

602p:
Looks like I spoke too soon on the good performance news. The first set of measurements from my last comment was badly skewed by some very strange intermittent out-of-stack bug I am encountering in `develop` AbleC (but not my branch.) I updated the comment above to reflect the real state. In actuality it looks like removing `location` and adding `origins` is either break-even or about 5-10% slower. I'm still not entirely sure. It seems like the performance numbers for a lot of these things depend on the phase of the moon...

@krame505 I have not been triggering Jenkins myself... I don't know why that's happening (the moon...?)

602p:
I'd like to figure out what the next steps are for getting this into mainline. This currently has no outstanding issues/merge blockers (knock on wood.)
 - Should I be merging `feature/strategyattr` into this in the expectation that that will be merged first?
 - Is there a more real review process you'd like to go through @krame505 ? Could we set a time for a call for this?
 - Are there other considerations on bringing it into `develop` @ericvanwyk ?
 - Should I be preparing a matching PR to remove `location` from ableC (the `feature/nolocation` - or should it be `fix/nolocation`? :stuck_out_tongue: - branch is ready)? Do we want to look at doing the same for silver?

krame505:
Thanks for working on this, Louis!  

I will try to do a more detailed code review here on github soon - this weekend is busy for me, but maybe sometime next weekend.  If there are things to discuss after that we can definitely set up a call.  

Yes, it would be good to have a matching PR for ableC to actually use this before we merge origin tracking in Silver.  This will help in finding remaining warts and bugs that only show up through more widespread use and in cases we haven't considered.  The biggest task here is updating the various ableC extensions to utilize origin tracking instead of manual locations.  My advice here (once we have Jenkins working again, anyway) is to "follow the build failures".  Protip: if you put all these changes on branches with the same name across all of the repos (i.e. naming them all `feature/origintracking`), then Jenkins will automatically check out and build with the corresponding versions of everything.  

Also once we have updated things to make use of origins, we should perform some manual inspections to verify that the error locations are still sensible.  Perhaps we should consider automating this for the error tests in ableC extensions by creating files with the expected outputs, which would alert us to when there is a change in the location for an error message.  

One potential issue I just realized with using origins for errors is that we want it to be clear when an error arises from generated code, as this usually is indicative of a bug.  Traditionally we have achieved this with locations - when using concrete syntax AST literals, the location annotations of the constructed AST point to the location in the Silver source files, which is rather helpful for debugging.  Is there any way to keep this behavior when using origins?  The way that these concrete syntax extensions (such as silver-ableC) work is by forwarding to a bunch of production calls with their locations annotations set from the locations of the concrete syntax literal that was parsed as a part of the Silver specification.  Is there some expression that we could generate instead to explicitly set the origin of these constructed trees to be the origin of the parsed concrete syntax?

ericvanwyk:
That last point @krame505 makes is where the use of origins holds the most promise. Really getting to the point of doing some interesting things with origins instead of the location stuff is what I'm looking forward to.

602p:
@krame505 let's try to set a date for a code review. Does tomorrow or Friday work?

I have just created downstream PRs for ableC and silver-ableC. These work just fine in my (limited) testing. Unfortunately your suggestion to use `OVERRIDE_JARS` is insufficient to get Jenkins building this branch. This is because it overrides the jars for the "bootstrapping" (from `develop` silver) compile as well as the self-hosting compile and the new "ABI" is not backwards compatible. What needs to happen (what `rebuild-for-child.sh` does) is:
 - Install the `develop/HEAD` `SilverRuntime.jar`
 - Install the hack/old versions of `core/originsimpl/Impl.sv`, `core/reflect/AST.sv`, `silver/reflect/Util.sv`, and `silver/langutil/Message.sv`
 - Build the `feature/origintracking` compiler with a `develop` `silver`
 - Use the `feature/origintracking` compiler to translate (but not java-compile) `core` and `lib:xml:ast` using the new ABI
 - Build and install the `feature/origintracking` runtime
 - Switch back to the new versions of `core/originsimpl/Impl.sv`, `core/reflect/AST.sv`, `silver/reflect/Util.sv`, and `silver/langutil/Message.sv`

It seems this might be above the pay grade of `OVERRIDE_JARS`, and I'm not sure there's a sane way to get Jenkins to do it. Looking at the Jenkins script it invokes `deep-rebuild` (I think... one of the scripts) as the build step. We could maybe get away with replacing `deep-rebuild` with `rebuild-for-child.sh` or adding a `OVERRIDE_BUILD` flag, but those both seem gross. Not sure what the best way forward is.

Regarding AbleC extension origins, I've added some code the the `feature/nolocation` branch of AbleC to allow extensions to mark themselves. If I write a simple extension that looks like this:

```haskell
...
global MODULE_NAME :: String = "edu:umn:cs:melt:exts:ableC:demo";
global MODULE_NOTE :: OriginNote = extensionGenerated(MODULE_NAME);

abstract production sayHiStmt
top::Stmt ::= e::Expr
{
  attachNote MODULE_NOTE; -- we could consider a file-wide version of this statement for use in extensions

  forwards to ableC_Stmt{printf("Hello %s!\n", $Expr{e});}; -- also works with manual exprStmt(...) construction
}
```

and compose an AbleC with it and compile a file:

```C
// Note no #include<stdio.h> so printf is undefined

int main() {
	say_hi(foo);
	return 0;
}
```
I get the following output:

```
INTERNAL ERROR: The following error message originated in extension-generated code.
This is probably indicative of a bug in the extension as opposed to your code.
The offending extension was: 'edu:umn:cs:melt:exts:ableC:demo' - please report this to it's developers.
The error was: error: Undeclared value printf.
Origins chain terminates without location.
Origins chain follows:
  -> core:originOriginInfo(core:setAtConstructionOIT(), edu:umn:cs:melt:ableC:abstractsyntax:host:name("printf"), [], true)
  -> core:originOriginInfo(core:setAtConstructionOIT(), edu:umn:cs:melt:exts:ableC:demo:abstractsyntax:sayHiStmt(edu:umn:cs:melt:ableC:abstractsyntax:host:directRefExpr(edu:umn:cs:melt:ableC:abstractsyntax:host:name("foo"))), [edu:umn:cs:melt:ableC:extutil:extensionGenerated("edu:umn:cs:melt:exts:ableC:demo")], true)


test.xc:4:8: error: Undeclared value foo
```

Note that the error on `foo` being undefined isn't flagged as extension generated because it wasn't manipulated :). If it's really wanted I could probably restore the old logic of using the silver location of the template via `logicalLocation` but that seems kinda like it was a hack in the first place. If flagging where in the extension a piece of code is generated is important for debugging you can always do `attachNote dbgNote("arbitrary string that will appear in origins chain") on {...some code...}`.

krame505:
Tomorrow or Friday evening would work for me, I am free after about 6 - let me know what you prefer.  

Not sure I understand the Jenkins issue.  The jars you provided via OVERRIDE_JARS were the already self-hosting ones resulting from `rebuild-for-child.sh` locally, right?  And once origin tracking is released that script should no longer be needed, correct?  After providing the fully boostrapped jars the `feature/origintracking` branch would become self-hosting on Jenkins.  Note that since `feature/strategyattr` is also currently self-hosting, we may want to hold off on this for `feature/origintracking` until `feature/strategyattr` is merged.  

Also note for the downstream builds to (eventually) succeed on Jenkins the corresponding branches will also need to be named `feature/origintracking`.  I'm not sure if it is possible to rename the branch for an open pull request so you may need to make new ones on ableC and silver-ableC.  

While a step in the right direction, I don't think the silver-ableC fix you are proposing here is sufficient.  `ableC_* { }` code blocks can be quite complex so it is important that the error message somehow points to the location in the original concrete syntax.  In your example, what if there were a bunch of `printf` statements, only one of which contained an error?  

I think the answer here somehow involves modifying `silver:metatranslation` to generate code that attaches notes containing the original locations from the reflected abstract syntax tree to the abstract syntax tree that is built when the generated code is run.  This would be analogous to how the metatranslation library currently generates a production call with a named parameter for location when the original reflected tree has a location annotation.

602p:
> Tomorrow or Friday evening would work for me, I am free after about 6 - let me know what you prefer.
How about tomorrow at 7?

> ... would become self-hosting on Jenkins ...
OH! Duh! I misunderstood what you were suggesting re:`OVERRIDE_JARS`. I'll try that and yes, I expect that should work fine. I'll also change the downstream branches so hopefully they get picked up.

> I think the answer here somehow involves modifying `silver:metatranslation` to generate code that attaches notes containing the original locations from the reflected abstract syntax tree to the abstract syntax tree that is built when the generated code is run.
Sure, the notes attached can be totally dynamic so this should def be possible. I'll look at doing that.

grammars/core/reflect/AST_tracked.sv
```diff
@@ -0,0 +1,64 @@
+grammar core:reflect;
```
krame505, line 1: Rename back to AST.sv before merging?

grammars/core/originsimpl/Impl_real.sv
```diff
@@ -0,0 +1,27 @@
+grammar core:originsimpl;
```
krame505, line 1: Is there a reason for this to not just be in core?  Also rename before merging, I assume?

grammars/core/originsimpl/Impl_hack.sv.disabled
```diff
@@ -0,0 +1,21 @@
+grammar core:originsimpl;
```
krame505, line 1: Delete before merging?

grammars/core/reflect/AST_tracked.sv
krame505, line 1: Rename file before merging

krame505: And delete AST_untracked.sv.disabled

grammars/silver/definition/core/Expr.sv
```diff
@@ -245,6 +254,18 @@ top::Expr ::= e::Decorated Expr es::Decorated AppExprs anns::Decorated AnnoAppEx
   top.typerep = functionType(ety.outputType, es.missingTypereps ++ anns.partialAnnoTypereps, anns.missingAnnotations);
 }
 
+concrete production noteAttachment
+top::Expr ::= 'attachNote' note::Expr 'on' '{' e::Expr '}'
```
krame505, line 258: Hmm, not sure a massively like curly braces wrapping an expression.  For consistency I would maybe prefer `on ... end`, but I am open to other suggestions.

grammars/silver/definition/core/NonTerminalDcl.sv
```diff
@@ -3,15 +3,26 @@ grammar silver:definition:core;
 autocopy attribute nonterminalName :: String;
 
 concrete production nonterminalDcl
-top::AGDcl ::= cl::ClosedOrNot 'nonterminal' id::Name tl::BracketedOptTypeExprs nm::NonterminalModifiers ';'
+top::AGDcl ::= quals::NTDeclQualifiers 'nonterminal' id::Name tl::BracketedOptTypeExprs nm::NonterminalModifiers ';'
+{
+  forwards to noWrapperNonterminalDcl(quals, $2, id, tl, nm, $6, location=top.location);
+}
+
+abstract production noWrapperNonterminalDcl
```
krame505, line 11: What is the purpose of this production?  Why not just make this `nonterminalDcl`, which now has the same signature?

grammars/silver/definition/origins/BlockContext.sv
```diff
@@ -0,0 +1,56 @@
+import silver:definition:core;
+import silver:definition:env;
+import silver:definition:flow:driver;
+
+{--
+ - Does code generated in this block need to reference originsCtx, or can it
+ - get information from the frame's left-hand-side and rules? Effectively, is
+ - this code in a function or a production?
+ -}
+synthesized attribute originsContextSource :: ContextOriginInfoSource occurs on BlockContext;
```
krame505, line 10: Nore sure what this is for, but seems like there are some syn equations missing here?  I know the silver compiler is already really bad w.r.t the MWDA, but we don't want to introduce any new flow errors if possible.

grammars/silver/langutil/Message_tracked.sv
```diff
@@ -0,0 +1,142 @@
+{- A Universal error/warning message data structure -}
```
krame505, line 1: Rename file and delete disabled version before merging

grammars/silver/modification/impide/IdeDecl.sv
```diff
@@ -169,11 +169,11 @@ top::IdeStmt ::=
 }
```
krame505, line 170: Have you tested whether `./make-ide` still works?  Some of that code doesn't get rebuilt in normal deep rebuilds.

grammars/silver/reflect/Util_new.sv
```diff
@@ -0,0 +1,60 @@
+grammar silver:reflect;
```
krame505, line 1: Reminder to rename and delete disable version

grammars/silver/rewrite/ASTExpr.sv
```diff
@@ -249,6 +249,13 @@ top::ASTExpr ::= c::ASTExpr t::ASTExpr e::ASTExpr
     end;
 }
 
+abstract production noteAttachmentASTExpr
+top::ASTExpr ::= a::ASTExpr b::ASTExpr
+{
+  top.pp = pp"(${a.pp} + ${b.pp})";
+  top.value = b.value; -- ORIGINS TODO: ???
```
krame505, line 256: Still TODO?  Not sure if this is easy/possible to implement.

rebuild-for-child.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
```
krame505, line 1: Goes away once we get this to be self-hosting on Jenkins and are almost ready to merge this branch, right?

runtime/java/build.xml
```diff
@@ -28,7 +28,7 @@
     <javac debug='on' classpathref='compile.classpath' srcdir='${src}:${grammars}' destdir='${bin}' includeantruntime='false'>
       <include name='common/**/*.java' />
       <include name='core/*.java' />
-      <include name='monto/*.java' />
+      <!-- <include name='monto/*.java' /> -->
```
krame505, line 31: We should maybe just remove this code (in a separate PR) rather than just letting it sit around and not be built.  @ericvanwyk?

runtime/java/src/common/Node.java
```diff
@@ -121,6 +137,28 @@ public DecoratedNode decorate() {
 	 */
 	public abstract String getName();
 
+	public Object duplicate(Object redex, Object rule){
```
krame505, line 140: What is the purpose of these methods?

runtime/java/src/common/OriginContext.java
```diff
@@ -0,0 +1,136 @@
+package common;
+
+import core.*;
+import java.util.*;
+import common.exceptions.*;
+
+
+/**
+ * Implementation of "the stuff on the left of the turnstile" that needs to be
+ *  passed along for origin tracking. This would be a value type if Java
```
krame505, line 10: Might be good to open a bug about this - things we would change if/when Java gets value types (https://openjdk.java.net/jeps/169.)  Not urgent.

support/origintracking/svdraw2/svdraw2.py
```diff
@@ -0,0 +1,258 @@
+import os, sys
```
krame505, line 1: Not going to try reviewing this, but I trust that you (mostly) know what you are doing :-)

Also why svdraw2 and not svdraw?

tutorials/simple/with_all/silver-compile-custom
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
krame505, line 1: What is the purpose of this script?

grammars/core/originsimpl/Impl_real.sv
602p: No, I don't think there is. Wonder why I did that... and yes, rename before merging.

grammars/silver/definition/core/NonTerminalDcl.sv
602p: Uh, no good reason for this now I think either. This was a holdover from when origin tracking was a modification instead of core I believe.

runtime/java/src/common/Node.java
602p: These are the noop implementations for origin tracking helper methods. These could probably be changed to panic since they should never be called since now the trackedness information is precise after moving it into `nonterminalType`. The semantics/use of (the specializations of) these helpers is discussed in https://github.com/melt-umn/melt-website/blob/feature/origintracking/silver/concepts/origins/index.md .

support/origintracking/svdraw2/svdraw2.py
602p: Oh yeah, this script is a total disgusting hack. I offer no real defense other than that it's very very useful (see the docs PR for examples of the diagrams it makes.)

It'd be nice to have a not-horrible version of a viz tool like this. It could probably be built entirely in silver using reflection, although it might be kind of a PITA to do the graph processing in silver. It might make more sense to just rewrite a not-horrible python version.

grammars/silver/rewrite/ASTExpr.sv
602p: To un-TODO-ify this we need to think more deeply about handling reflection + origins. As is the origin of an AST is the thing it was reflected from. If manipulated/transformed it's origin would be the AST it was transformed from (or whatever) and that's origin would transiently be what it was reflected from. Then when something is reified the reified nodes have the ASTs as their origins. It's conceptually very neat, but doesn't allow reflective access to/manipulation of the origins information.

If we maintain the current state here and decide we don't want to enable reflective manipulation of origins (because we think about origins as properties of silver/java objects and not of silver values) we could implement this as `top.value = attachNote reify(a) on {b.value}` and that might be considered reasonable.

On the other hand we could decide we consider origins as part of silver values. To change this we would need to augment `nonterminalAST` to take a `Maybe<AST>` to represent it's origin info, and decide on some semantics for how this is allowed to be manipulated/how reification works (most importantly: if it's not an `anyAST` when it gets reified it won't actually point to the thing they want it to (what the reflected from) but instead to the newly reified object.) Then we would also have to augment `ASTExpr` with some kind of `originRulesEnv` to track the new rules for the rewriting evaluation.

Let's talk about this tomorrow evening in more detail. Either approach sounds reasonable to me... the truth is probably that if you hit this corner case you're trying to do something evil™ anyway :P.

runtime/java/src/common/Node.java
krame505: So these are always overridden in the nonterminal/production subclasses?  Could these just be made abstract?

support/origintracking/svdraw2/svdraw2.py
krame505: Hmm, maybe.  Silver does have a native graph library (under `silver:util:raw:graph`) by the way.

runtime/java/src/common/Node.java
602p: Moved them to `TrackedNode` and made abstract.

grammars/silver/rewrite/ASTExpr.sv
602p: This has been implemented as discussed (attaching the reified note to the AST.)

krame505:
Another thought that occurred to me on the reporting errors in generated code business: perhaps it shouldn't always be a fatal internal error when errors occur in generated code.  For example in the following production that provides an alternative syntax for addition:
```
abstract production myAdd
top::Expr ::= e1::Expr 'plus' e2::Expr
{
  top.pp = pp"${e1.pp} plus ${e2.pp}";
  forwards to ableC_Expr { ${e1} + ${e2} };
}
```
Here we don't do any explicit error checking, but the errors that would occur on the generated code would be acceptable.  However on other extension productions we do wish to flag errors in generated code, complete with information about the original source location:
```
abstract production myAdd2
top::Expr ::= e1::Expr 'plus2' e2::Expr
{
  top.pp = pp"${e1.pp} plus2 ${e2.pp}";
  local localErrors::[Message] = e1.errors ++ e2.errors ++
    if !someManualTypeChecking(e1.typerep, e2.typerep)
    then [err(...)]
    else [];
  forwards to
    if null(localErrors) then ableC_Expr { ${e1} + ${e2} } else errorExpr(localErrors);
}
```

So in addition to attaching notes on trees constructed with concrete syntax literals, we may consider another type of note that can be attached in the second production to indicate that trees constructed here should not contain errors.  The internal error check added to ableC in feature/nolocation that looks for errors in generated code would first look for this note to determine whether errors should be considered internal.  

By the way @602p did you ever get around to reading that paper I sent which explains what is going on in `silver:metatranslation`?  I would be happy to set up a meeting some time to go over this in more detail.

602p:
My intent in introducing the `extensionGenerated` note was that it would be used for the second case (i.e. where there should be no errors in the forward because they should have been caught and reported in the extension production.) Not having the internal error behavior (as in your first case) is as simple as not attaching that note. If we want to track any-and-all extension-generated code (not just that which shouldn't have errors) we could refactor `extensionGenerated` to take a boolean `treatErrorsAsInternalErrors` or split it into `extensionGenerated` and `shouldBeErrorFree`. 

With respect to having access to the source location of the `ableC_Expr{...}` as we did previously, I have a working local patch I'm cleaning up that will attach a `silverAbleCTemplateLocation(...)` note to trees constructed from `ableC_*` blocks. That should be the best of both worlds, in that we can see (following the origins chain) the source location in the user AbleC code that triggered the error (I have marked `Stmt` tracked) and the location in the AbleC extension that failed to handle it itself. This means the internal error message from my example (the `say_hi(` stmt) looks like this:

```
INTERNAL ERROR: The following error message originated in extension-generated code.
This is probably indicative of a bug in the extension as opposed to your code.
The offending extension was: 'edu:umn:cs:melt:exts:ableC:demo' - please report this to it's developers.
The error was: error: Undeclared value printf.
Origins reports the following source location: test.xc:4:1.
Origins chain follows:
  -> core:originOriginInfo(core:setAtConstructionOIT(), edu:umn:cs:melt:ableC:abstractsyntax:host:name("printf"), [], true)
  -> core:originOriginInfo(core:setAtConstructionOIT(), edu:umn:cs:melt:exts:ableC:demo:abstractsyntax:sayHiStmt(edu:umn:cs:melt:ableC:abstractsyntax:host:directRefExpr(edu:umn:cs:melt:ableC:abstractsyntax:host:name("foo"))), [edu:umn:cs:melt:ableC:extutil:extensionGenerated("edu:umn:cs:melt:exts:ableC:demo"), edu:umn:cs:melt:ableC:extutil:silverAbleCTemplateLocation(core:txtLoc("SayHi.sv:19:14"))], true)
...SNIP...
  -> core:parsedOriginInfo(core:setFromParserOIT(), core:loc("test.xc", 4, 1, 4, 12, 144, 155), [])
```
We can see in the above the `silverAbleCTemplateLocation(core:txtLoc("SayHi.sv:19:14"))` gives us the location of the `ableC_Stmt` block that generated the erroring tree.

(Relatedly: The case of wanting to use origins to debug the compiler as opposed to the program being compiled is the use-case for the `--tracing-origins` flag. If you build something with the `--tracing-origins` flag it will automatically add a note to each expression (actually every time the `contextRef` changes, but pretty close) of the source location for that expression. Building the AbleC compiler with that means that the relevant line of the error message looks like this instead:
```
  -> core:originOriginInfo(core:setAtConstructionOIT(), edu:umn:cs:melt:exts:ableC:demo:abstractsyntax:sayHiStmt(edu:umn:cs:melt:ableC:abstractsyntax:host:directRefExpr(edu:umn:cs:melt:ableC:abstractsyntax:host:name("foo"))), [core:originDbgNote("core:loc("SayHi.sv", 19, 14, 19, 58, 602, 646)"), edu:umn:cs:melt:ableC:extutil:extensionGenerated("edu:umn:cs:melt:exts:ableC:demo"), edu:umn:cs:melt:ableC:extutil:silverAbleCTemplateLocation(core:txtLoc("SayHi.sv:19:14"))], true)
```
See the `core:originDbgNote("core:loc("SayHi.sv", 19, 14, 19, 58, 602, 646)")` that reports the same information we want. So one option is to avoid adding note-attaching complexity to silver-AbleC and just use a `--tracing-origins` build when we need information on where in an extension generated bad code.)

rebuild-for-child.sh
krame505: @602p we do want to delete this, right?  Otherwise is this branch ready to merge?

krame505 (approved):

PR #340, opened 2020-01-15
Bug fix to lexical precedence relations involving subclasses

krame505:
This fixes a bug with the implementation of lexer class `extends`.  In addition to terminals in a subclass automatically becoming members on a superclass, lexical precedence relations defined for other terminals/lexer classes on the superclass should also apply to subclasses.


krame505:
This also includes the changes from #338 - that should be merged first.

ericvanwyk:
Looks good to me.

PR #342, opened 2020-01-21
Implement strategic term rewriting library and extension

krame505:
This is pretty big,  But in summary:
* A library (`silver:rewrite`) contains defines the `Strategy` nonterminal, with productions for various strategy combinators such as choice, sequence, fail, id, all, one, etc.  This is heavily inspired by Stratego, TOM, and other rewriting systems.  Attributes on `Strategy` thread through an `AST` to be rewritten.  The overall result of applying a strategy to an `AST` is success or failure.  
* The `ASTExpr` nonterminal (also in `silver:rewrite`) represents a pattern or expression to be evaluated on an `AST`.  The `rewriteRule :: (Strategy ::= ASTExpr ASTExpr)` constructor consists of a pattern `ASTExpr`, to be unified with the incoming `AST` and failing or succeeding with a substitution, and a result `ASTExpr` to be evaluated to a new `ASTExpr` resulting from the substitution.  A programmer could construct a `rewriteRule` strategy manually, but this would be tedious and not type safe.  
* The one wrinkle here is handling general function application (besides direct production calls.)  Function values are stored as `anyAST`s,  but we can't statically know the type of function into which we wish to reify, so I added a utility to the runtime that applies a function wrapped as `AST` to `AST` arguments, resulting in an `AST`.  This is possible because we can inspect the function value wrapped in `anyAST` to determine the type to which we must reify the arguments.  Operations on decorated trees (e.g. attribute access or `new`) can be done by applying a function literal that is a custom lambda.  
* An extension to Silver (`silver:extension:rewrite`) provide new syntax for applying a strategy (`rewriteWith(strategy, expr)` - this is not a function because of the result type problem with `reify`), new infix operators for choice (`<+` - same as Stratego) and sequence (`<*` - by analogy with `<+`, Stratego uses `;` which isn't an option here), and most importantly, new syntax for constructing a rule:
```
global evalStep::Strategy =
  rule on Expr of
  | add(const(a), const(b)) -> const(a + b)
  | sub(const(a), const(b)) -> const(a - b)
  | mul(const(a), const(b)) -> const(a * b)
  | div(const(a), const(b)) when b != 0 && a % b == 0 -> const(a / b)
  | div(const(a), const(b)) when b != 0 && gcd(a, b) > 1 ->
     let g::Integer = gcd(a, b) in div(const(a / g), const(b / g)) end
  end;
```
This reuses pattern matching syntax, but defines attributes on patterns and expressions to translate them into `ASTExpr`s (this required an awful hack to obtain the decorated RHS `Expr`s to translate, due to the way the pattern matching extension defers most typing to a low-level matching modification.)  Thus, theoretically, one could use concrete object language syntax on either side of the rule.  Every expression in Silver has a translation into `ASTExpr`, thus I sort of accidentally built a Silver interpreter.   This translation code incidentally reuses the Silver host embedding library to meta-explode the `ASTExpr`s built at compile time into the Silver code needed to construct them at runtime.  

### Note to self of stuff still left to do
* [x] More thorough tests of using all potential expressions on a rule RHS.  Particularly weird cases involving decoration and attribute accesses (there are probably some bugs hiding here.)
* [x] Come up with an example extension to ableC that uses rewriting with concrete object syntax.
* [ ] Implement a standalone example (interpreted language? Maybe adapt a Stratego tutorial?) and add a section to the SIlver tutorials.

### DIFF
--- a/grammars/core/Math.sv
+++ b/grammars/core/Math.sv
@@ -18,3 +18,12 @@ Integer ::= l::Integer r::Integer
   return if l < r then l else r;
 }
 
+{--
+ - Returns the greatest common divisor of two numbers.
+ -}
+function gcd
+Integer ::= l::Integer r::Integer
+{
+  return if r == 0 then l else gcd(r, l % r);
+}
+
--- a/grammars/silver/definition/core/Expr.sv
+++ b/grammars/silver/definition/core/Expr.sv
@@ -305,10 +305,11 @@ top::Expr ::= e::Expr '.' q::QNameAttrOccur
   -- 'q' shouldn't actually resolve to a name!
   top.errors := e.errors ++ forward.errors;
   
-  q.attrFor = performSubstitution(e.typerep, e.upSubst);
+  local eTy::Type = performSubstitution(e.typerep, e.upSubst);
+  q.attrFor = if eTy.isDecorated then eTy.decoratedType else eTy;
   
   -- Note: we're first consulting the TYPE of the LHS.
-  forwards to q.attrFor.accessHandler(e, q, top.location);
+  forwards to eTy.accessHandler(e, q, top.location);
   -- This jumps to:
   -- errorAccessHandler  (e.g. 1.unparse)
   -- undecoratedAccessHandler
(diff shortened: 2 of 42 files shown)


krame505:
@ericvanwyk I think that this is basically feature-complete and ready to merge.  Go ahead and approve whenever you are done looking over the changes here.

PR #345, opened 2020-02-04
Make prefix separators less brittle

### LINKED ISSUE #85: Specification of prefix separator in parser spec doesn't override one in host.
Specifying a prefix separator inside a parser specification doesn't override the one specified in the host language.  Instead an error is raised stating that there are two prefix separators specified.

I think a nicer behavior would be to have the parser spec one take precedence.

### LINKED ISSUE #148: 'prefix separators' are fragile and easily broken.
We have a feature that allows a separator to be declared for transparent prefixes.

The relevant code is here:

https://github.com/melt-umn/silver/blob/develop/grammars/silver/modification/copper/Prefix.sv
https://github.com/melt-umn/silver/blob/develop/grammars/silver/modification/copper/ParserDcl.sv

The problem is that the discovery of a prefix separator works like this:

https://github.com/melt-umn/silver/blob/develop/grammars/silver/modification/copper/Prefix.sv#L67

```
getValueDcl("_prefix_seperator", top.env)
```

Presently, this works in toy examples because either:

1. There is only one grammar involves that declares a prefix separator.
2. The prefix separator is specified as part of the parser declaration, which (erroneously) suppresses new declarations defs from propagating outward.

https://github.com/melt-umn/silver/blob/develop/grammars/silver/modification/copper/ParserDcl.sv#L20

```
top.defs = [funDef(top.grammarName, n.location, namedSig)];
m.env = appendEnv(toEnv(liftedAGDcls.defs ++ med.defs), top.env);
```

Here, `liftedAGDcls` gets sent back down the parser declaration list (`m`) and is not included in the general declarations (really, as it should be.)

Some other, more robust, mechanism is necessary for specifying prefix separators.

krame505:
Fixes #148.

This moves prefix separators from being a standalone global definition (that can only appear once) to being specified on lexer classes.  This means that the prefix separator is now a property of individual terminals, such that terminals with different prefix separators may be included in the same parser, and specifying string prefixes will automatically choose the appropriate separator.  Note that this means it is no longer makes sense to allow the prefix separator to be globally overridden in a parser specification (resolves #85); in such an (unusual) case users may still specify a regex or single-quoted literal to define a prefix that includes the desired separator, if any.

This change is particularly helpful in dealing with the composition of multiple host languages (e.g. silver-ableC), as included extensions to ableC that require transparent prefixes should use ableC's prefix separator and not Silver's.  More possibilities also exist: for example an extension that defines a DSL mimicking another language may wish to use prefix separators that mimic the syntax of that language, or a host language might choose different prefix separators for keywords and operators.  However the most common case (as with ableC) is that a host language wishes to use the same prefix separators for all terminals in its lexer classes; this may be achieved elegantly through the use of a common lexer superclass:
```
lexer class AbleC  prefix separator "::";

lexer class Comment extends AbleC;
lexer class Identifier extends AbleC;
lexer class Keyword extends AbleC;
...
```
A default prefix separator of ":" for Silver was also added in a similar fashion.  

One annoying implementation detail is that the creation of new terminals to serve as prefixes can't be moved entirely into the CstAst code, as Java class providing the Silver representation of the terminals must still be generated.  Thus writing `foo prefix Foo_t with "bar";` in a parser spec must still declare the prefix terminal, even though we don't know `Foo_t`'s prefix separator until we build the CstAst.  
This problem is solved by introducing an (abstract) terminal modifier for the generated prefix terminal, specifying which terminals are being prefixed.  During CstAst normalization, the regex of any (prefix) terminals with this modifier are updated to append the prefix separator for the prefixed terminals.  This does mean that all terminals in each prefix specification must have the same prefix separator as a single terminal is generated for each prefix specification (to avoid lexical ambiguities between prefix terminals.)  The only major consequence of this restriction is that when prefixing a grammar that exports multiple marking terminals with different prefix sepererators, the terminals being prefixed must be listed explicitly for every terminal in a prefix specification for each prefix separator (example: https://github.com/melt-umn/ableC-sample-projects/blob/feature/prefixSeperator/using_transparent_prefixes/alternate_separator/CompilerSpec.sv#L26.)

### DIFF
--- a/grammars/silver/definition/concrete_syntax/ParserSpec.sv
+++ b/grammars/silver/definition/concrete_syntax/ParserSpec.sv
@@ -42,8 +42,5 @@ top::ParserSpec ::= sl::Location  sg::String  fn::String  snt::String  grams::[S
     moduleExportedDefs(sl, top.compiledGrammars, computeDependencies(grams, top.compiledGrammars), grams, []);
 
   top.cstAst = cstRoot(fn, snt, foldr(consSyntax, nilSyntax(), addedDcls ++ med.syntaxAst), terminalPrefixes);
-  
-  local decomposedTerminalPrefixes :: Pair<[String] [String]> =
-    unzipPairs(terminalPrefixes);
 }
 
--- a/grammars/silver/definition/concrete_syntax/TerminalDcl.sv
+++ b/grammars/silver/definition/concrete_syntax/TerminalDcl.sv
@@ -101,7 +101,7 @@ top::TerminalKeywordModifier ::=
 
 
 nonterminal TerminalModifiers with config, location, unparse, terminalModifiers, errors, env, grammarName, compiledGrammars, flowEnv;
-nonterminal TerminalModifier with config, location, unparse, terminalModifiers, errors, env, grammarName, compiledGrammars, flowEnv;
+closed nonterminal TerminalModifier with config, location, unparse, terminalModifiers, errors, env, grammarName, compiledGrammars, flowEnv;
 
 synthesized attribute terminalModifiers :: [SyntaxTerminalModifier];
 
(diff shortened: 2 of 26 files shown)


krame505:
Note - this contains the changes from #342 and #344, merge this after those are merged.

PR #346, opened 2020-02-06
Bug/install

ericvanwyk:
This fixes a small bug in the Silver install script.  The '-a' is a deprecated test for file existence.  The bug is that if the symbolic link but not file it pointed to exists, then the test failed and the link was not removed.  Then the following ``ln -s`` command would fail.  The test now check if the pointed to file exists (``-e``) or (``-o``) if the symbolic link exists (``-h``).

This also removes the useless check on the return code of this command.  Since the ``-eu`` flag is set the script aborts on a non-zero return code and this code is not reached.

### DIFF
--- a/grammars/silver/composed/Default/Main.sv
+++ b/grammars/silver/composed/Default/Main.sv
@@ -1,47 +1,9 @@
 grammar silver:composed:Default;
 
 import silver:host;
-import silver:translation:java;
-import silver:driver;
-
-import silver:extension:doc;
-import silver:analysis:warnings:flow;
-import silver:analysis:warnings:exporting;
 
 parser svParse::Root {
   silver:host;
-
-  silver:extension:convenience;
-  silver:extension:list;
-  silver:extension:easyterminal;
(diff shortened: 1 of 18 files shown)


ericvanwyk:
OK, I'm confused.  Was this already merged into ``develop`` earlier?  When I look at ``develop`` now I see the changes that were introduced here.  But when I hit the "Compare"  button earlier today it showed the expected difference between this branch and ``develop``.

ericvanwyk:
I've since merged ``develop`` into this ``bug/install`` branch and now they are the same.  So I'm closing this pull request.

krame505:
I think what happened is you merged develop into this branch, then I merged your pull request locally on my machine without first pulling your merge commit.  So the extra merge commit on this branch is the only difference.

ericvanwyk:
I'm not so sure.  I did merge develop into this branch, but this was after I'd clicked the "compare" button and saw the differences.  Who knows...  Thanks for deleting the branch.

PR #354, opened 2020-03-07
Re-implement functor attributes/propagate in an extensible way, implement monoid attributes

krame505:
I've started doing a bit of cleanup and refactoring work that needs to happen before I can work on strategy attributes (and other kinds propagated attributes we may want to add such as monoid.  The current propagate extension only works for functor attributes and does nothing with regards to attribute declaration/occurrence.)

I have refactored `propagate` into something that is overloadable through the environment, such that different kinds of attributes can introduce their own `propagate` behavior.

We also want to make `occurs on` (production `attributionDcl`) overloadable - for example if writing
`functor attribute host;` 
turns into
`synthesized attribute host<a>::a;`
then
`attribute host occurs on Expr;`
 should become
`attribute host<Expr> occurs on Expr;`.
Strategy attributes will require a single attribute occurrence becoming occurence declarations for multiple attributes.
Making this possible required some annoying refactoring, due to the mutually-recursive nature of Silver's environment.  Since what `attributionDcl` now forwards to depends on the `env`, and `env` depends on `defs` introduced by the `attributionDcl`, we have a circular dependency.  This can be fixed by overriding the `defs` equation on `attributionDcl`, however we do want the attribute-occurrence defs that come from what `attributionDcl` forwards to.  This was fixed by splitting attribute occurrences out of `defs` into a seperate `occursDefs` attribute, parallel to `defs` at the top levels of the abstract syntax.

The functor attribute extension has been refactored to make use of these features.  Additionally there is a bunch of minimally-related cleanup and refactoring here.


krame505:
I did do a bit more here, monoid attributes are also implemented, such as 
```
monoid attribute errors::[Message] with [], ++;
```
Note this syntax is inspired by collection attributes.  For now all monoid attributes are automatically collection attributes using the same operator, which is quite useful.  One can globally write `propagate errors on Expr;` but then also contribute additional values on the various `Expr` productions without needing to override the base equation.

krame505:
@ericvanwyk I think this is ready to merge (pending us being happy with the in-progress Silver release.)  Could you take a look?

krame505:
@ericvanwyk Now that the releases are done, I think this is ready to merge.  Can you approve this and #355 when you get a chance?

krame505:
@ericvanwyk I see that you approved #355.  Can you approve this pull request (#354), which needs to be merged first?

PR #355, opened 2020-03-18
 Use monoid attributes as appropriate in the Silver compiler

krame505:
Based on #354, review/merge that one first.


krame505:
@ericvanwyk do we want to try and do a release of Silver + extensions for the reflection paper before merging this branch?  Or is it OK if these changes make it in there too?

krame505:
`excluding` syntax was added in this PR (#355), because, well, that was easier than adding it in #354 and merging it here.  This can be merged as soon as that PR is merged, @ericvanwyk can you take a look?

PR #359, opened 2020-04-09
Add more utility traversal strategies from Stratego's standard library

krame505:
This includes the `allTopDown` strategy used in the paper.

This should get merged before the other outstanding pull requests and included in the forthcoming 0.4.2 release of Silver.


krame505:
Yes, the stratego standard library is massive, see https://github.com/metaborg/strategoxt/tree/master/strategoxt/stratego-libraries/lib/spec/strategy

But almost any of these can be specified by programmers as needed using the built-in basic combinators.

Issue #368, opened 2020-06-12
Reify can't handle Skolem types


krame505:
I thought there was an issue for this, but I guess not.  

The result type of `reify` must be inferred statically, meaning that `reify`ing into a Skolem type will always fail.  For example,
```haskell
function reify2
Either<String b> ::= x::AST
{
  return reify(x);
}
```
will always return `left(...)` because `b` is an opaque Skolem constant that unifies with nothing.  

The type of `reify` is static because while we can ask for the runtime type representation of a particular value, there is no general way of knowing at runtime the concrete type that an expression should return.  There is an incomplete attempt at passing around this extra type information everywhere at [feature/expectedType](https://github.com/melt-umn/silver/tree/feature/expectedType), but this was abandoned due to the resulting significant increases in overhead and complexity.  

The suggested alternatives (for now) are
* Make your function that calls `reify` a forwarding language extension to Silver instead
* Use `applyAST` (only possible in some cases)
* Do something with type classes (not yet implemented, coming soonish)

tedinski:
Might be a good opportunity for a new warning or even error.

krame505:
Yeah, @602p (Louis) is on it

krame505:
Some thoughts on how we might actually fix this with type classes:
Essentially, we can introduce a type class `Typed` (or something) for things that have a run-time type representation.  This type class would have a method to get a run-time representation of this type.  `reify` would require there to be an instance of this type for its result; it would be an error to attempt to reify into a Skolem type lacking this constraint.  
Somebody could instead write the above function `reify2` as
```haskell
function reify2
Typed b => Either<String b> ::= x::AST
{
  return reify(x);
}
```
This would be quite similar in concept to passing around an expected type everywhere, except we would only pass it in the required places.  

There are 2 directions that we could go with implementing `Typed`.  
* We could make it a regular type class in the standard library, with special compiler support for deriving instances.  This means that we would need to bring in the run-time `TypeRep` representation of types as a foreign type, which kind of suggests that this type should be a regular nonterminal instead of a bunch of Java classes in the runtime.  This could be useful in other ways, but it would complicate the runtime and would probably have performance downsides.  The current mutation-based run-time unification approach probably wouldn't work with this change, either.
[...]

PR #369, opened 2020-06-17
Error/warn on impossible reifications

602p:
This introduces errors and warnings per the discussion in #368. In the case that we are attempting to reify with a skolem type directly, throw an error since this is impossible, e.g.:

```haskell
function noop
Either<String a> ::=
{
  return reify(reflect(3));
}
```

now yields `main.sv:8:9: error: reify invocation attempts to reify to a skolem type - this will never succeed, see https://github.com/melt-umn/silver/issues/368`.

In the case that we are attempting to reify with a type containing a skolem type, throw a warning since this is technically possible, e.g.:

```haskell
function noop
Either<String Maybe<a>> ::=
{
  return reify(reflect(nothing()));
}
```

now yields `main.sv:8:9: warning: reify invocation attempts to reify to a type containing a skolem - this will only succeed in the case that the value does not actually contain an instance of the skolem type, see https://github.com/melt-umn/silver/issues/368`.

### DIFF
--- a/grammars/silver/analysis/typechecking/core/BuiltInFunctions.sv
+++ b/grammars/silver/analysis/typechecking/core/BuiltInFunctions.sv
@@ -69,12 +69,12 @@ function containsSkolem
 Boolean ::= ty::Type
 {
   return case ty of
-    | skolemType(_) -> true
-    | nonterminalType(_, args) -> any(map(containsSkolem, args))
-    | decoratedType(ty) -> containsSkolem(ty)
-    | functionType(out, params, namedParams) -> containsSkolem(out) || any(map(containsSkolem, params)) || any(map((\x::NamedArgType -> containsSkolem(x.argType)), namedParams))
-    | _ -> false
-  end;
+         | skolemType(_) -> true
+         | nonterminalType(_, args) -> any(map(containsSkolem, args))
+         | decoratedType(ty) -> containsSkolem(ty)
+         | functionType(out, params, namedParams) -> containsSkolem(out) || any(map(containsSkolem, params)) || any(map((\x::NamedArgType -> containsSkolem(x.argType)), namedParams))
+         | _ -> false
+        end;
 }
 
 aspect production reifyFunctionLiteral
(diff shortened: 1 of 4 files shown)


grammars/silver/analysis/typechecking/core/BuiltInFunctions.sv
```diff
@@ -65,10 +65,32 @@ top::Expr ::= 'toString' '(' e1::Expr ')'
...
 top::Expr ::= 'reify'
 {
   top.upSubst = top.downSubst;
+
+  top.errors <-
+    case performSubstitution(top.typerep, top.finalSubst) of
+    | functionType(nonterminalType("core:Either", [stringType(), resultType]), [nonterminalType("core:reflect:AST", [])], []) ->
+      case resultType of
+        | skolemType(_) -> [err(top.location, "reify invocation attempts to reify to a skolem type - this will never succeed, see https://github.com/melt-umn/silver/issues/368")]
+        | ty -> if containsSkolem(ty) then [wrn(top.location, "reify invocation attempts to reify to a type containing a skolem - this will only succeed in the case that the value does not actually contain an instance of the skolem type, see https://github.com/melt-umn/silver/issues/368")] else []
```
krame505, line 90: (Minor) style note - `case`, `end` and all clauses should be indented the same. 
Also this could be a `when` pattern, would make this line slightly less long.

krame505:
Just checked the build log and it is indeed the test suite now failing to compile
```
Reflection.sv:179:9: error: reify invocation attempts to reify to a skolem type - this will never succeed, see https://github.com/melt-umn/silver/issues/368
Reflection.sv:189:9: warning: reify invocation attempts to reify to a type containing a skolem - this will only succeed in the case that the value does not actually contain an instance of the skolem type, see https://github.com/melt-umn/silver/issues/368
```

602p:
I ended up making some tweaks to the testing stuff to (1) make it so that `wrongCode` checks that the thing containing the substring is an error as opposed to another type of message and (2) support checking for warnings in such a way that the `AGDcl`s producing them still compile so they can be used in other tests. Assuming that looks good to you @krame505 this should be good to merge, assuming Jenkins is as happy with the test suite as my laptop is.

krame505:
Looks good to me.

The Jenkins webhooks seem to be busted again so it didn't trigger automatically.  I started a build manually, you should be able to merge as soon as it passes.

602p:
Looks like more than the webhooks might be busted. Integration failed in AbleC - http://coldpress.cs.umn.edu:8080/job/melt-umn/job/ableC/job/develop/590/console - which shouldn't be possible since it built successfully and I didn't touch codegen at all. I see the following in the output
```
+ ./runTests
sv-nailgun: couldn't find nailgun jar under /usr/share/java
./runTests: line 9: sv-serve: command not found
```

which might be relevant. @krame505 can you take a look at this or let me into Jenkins to poke at it myself?

krame505:
That... seems to be an unrelated issue.  Seems the nailgun jars went missing somehow?  

I'll get you set up with an account on Jenkins.  FYI you either need to be on the UMN network or set up a proxy over SSH.

602p:
Yep, the jars for nailgun are missing. I assume they came from the `nailgun` Ubuntu package which is no longer installed. How can we get that back? Complain to operator, or is this something @tedinski set up?

krame505:
I opened a separate issue on this over on the jenkins-lib repo: https://github.com/melt-umn/jenkins-lib/issues/17

krame505:
Merging this, since Jenkins is finally working again.

PR #376, opened 2020-09-04
Implement additional types of automatic attributes

### LINKED ISSUE #280: Better autocopy implementation
## The basic problem

Autocopy equations can't be statically generated. This simplest scenario to see where statically generating autocopy equations would fail is the following:

* Host language.
* Extension 1 with new autocopy attribute.
* Extension 2 with new forwarding production.

There is no (modular) place to generate the autocopy equations for the children of that production. This example is slightly contrived, however, because the MWDA would also mean the autocopy attribute would never be _demanded_ from that production's children!

But we could always add a forth piece:

* Extension 3 using 1&2, aspects that production, and causes demands for that autocopy attribute on the children.
* (Likewise, more extensions could do the same thing.)

> _Edit: Complete aside: Lucas points out that this makes autocopy able to do ever so slightly more than we would be able to do with just ordinary inherited attributes under the MWDA. Ext3 with normal inherited attributes would be disallowed, because the inherited equations do not exist, and adding them in that extension would be "orphaned." But autocopy doesn't mind adding them, even if orphaned, because we're in the seemingly unique situation that the missing equation could only take one form: autocopying._

## The problem with dynamic generation

Silver startup time is slow, and a part of the reason why is that we iterate over every single production looking for missing autocopy equations, before we start running any Silver code at all. We want to speed this up.

Bonus points for also removing unwanted information we currently must jam into the runtime. For instance, we currently generate a list of types of children specifically for autocopy startup to look at to determine whether to insert equations. So getting rid of this dynamic generation could also allow us to remove filling this information in at startup. Synergy!

## There is an extremely common case

Almost always, autocopy equations can be statically generated! Most autocopy attributes are in the host language, and so could be statically generated for every production (host or extension).

However, not all are. A good canonical example is `substituted` in AbleC, which uses a `substitutions` autocopy attribute. (Although this is an `option` from the host, that still doesn't mean we can statically generate equations for this attribute, since if the grammar is never included, this attribute doesn't exist!)

## Potential solutions

1. Dynamic generation at startup, as we do now.
2. Dynamic generation, but at runtime when we discover missing equations.
3. Hybrid: statically generate the common case, dynamically generate at runtime for missing equations.
    * Possible downside: we'd still need sufficient information around at runtime to be able to dynamically generate them. So we wouldn't get that synergistic effect.
4. Restrictions: autocopy could be host language only.
    * Serious problem for things like `substituted`
5. Static re-generation: Statically generate host language case, but add on a "transitive dependencies" static generator that fills in all potentially non-existant autocopy equations. This could cause "duplicate equations" to be filled in, but that's hardly an issue for autocopy.
    * Drawback: requires all grammars to pass MWDA.

## Current thoughts

We should see if we can make the Hybrid3 approach work well. Perhaps we can reduce the amount of information the runtime would need to dynamically generate missing equations to something trivial.

The basic problem at runtime generation is that a `DecoratedNode` gets demands for `inherited(X)` where `X` is the integer index of an inherited attribute. It's missing. So we have to figure out that `X` means attribute `A`, that `A` is autocopy, then look at the parent nonterminal, see if it has `A` and if so at which index `Y`, then fill in the Lazy that just accesses `Y`.

(Neat trick: because we just pass the same array of Lazys around when decorating children, we can mutate that array once, and it'll update everywhere it's been used.)

On possible solution here is to store just two arrays on each nonterminal: array of autocopy attribute names, array of autocopy attribute indexes. Then a `DecoratedNode` with a missing equation could start by seeing if it's index is in its own nonterminal's array, then it could look for that same string in its parent's nonterminal's array, then get the index. This would be a pretty small amount of information, and has the advantage that it's per-nonterminal, instead of per-production, like a lot of the metadata we have to generate right now.

krame505:
This implements several new kinds of automatic attributes that may be propagated:
* Ordinary inherited attributes may now be propagated, with the behavior of autocopy.  I believe this fixes #280.  
* Threaded attributes, e.g. `threaded attribute subsIn, subsOut :: Substitution;` "thread" a pair of synthesized and inherited attributes through a tree.  New syntax is introduced to allow more fine-grained control of this; for example writing
```
abstract production add
top::Expr ::= a::Expr b::Expr
{
  local tCheck  = ...;
  thread subsIn, subsOut on top, a, b, tCheck, top;
}
```
would generate the equations
```
a.subsIn = top.subsIn;
b.subsIn = a.subsOut;
tCheck.subsIn = b.subsOut;
top.subsOut = tCheck.subsOut;
```
Propagating threaded attributes on a production would simply thread the attributes on all children on which they occur.

* Equality attributes, e.g. `equality attribute isEqualTo, isEqual;` for checking equality between trees.  For example
```
equality attribute isEqualTo, isEqual;
nonterminal Term with isEqualTo, isEqual;
abstract production var
top::Term ::= n::String
{}
abstract production app
top::Term ::= a::Term b::Term
{}
abstract production var
top::Term ::= n::String a::Term
{}
propagate isEqualTo, isEqual on Term;
```
would translate to
```
inherited attribute isEqualTo<a>::a;
synthesized attribute isEqual::Boolean;
nonterminal Term with isEqualTo, isEqual;
abstract production var
top::Term ::= n::String
{
  top.isEqual =
    case top.isEqualTo of
    | var(n1) -> n == n1
    | _ -> false
    end;
}
abstract production app
top::Term ::= a::Term b::Term
{
  a.isEqualTo =
    case top.isEqualTo of
    | app(a1, b1) -> a1
    end;
  b.isEqualTo =
    case top.isEqualTo of
    | app(a1, b1) -> b1
    end;
  top.isEqual =
    case top.isEqualTo of
    | app(a1, b1) -> a.isEqual && b.isEqual
    | _ -> false
    end;
}
abstract production abs
top::Term ::= n::String a::Term
{
  a.isEqualTo =
    case top.isEqualTo of
    | abs(n1, a1) -> a1
    end;
  top.isEqual =
    case top.isEqualTo of
    | abs(n1, a1) -> n == n1 && a.isEqual
    | _ -> false
    end;
}
```
* Unification attributes, for equality checks when certain productions have special behavior (e.g. type variables.)  These are similar to equality attributes, except that each tree is provided to the other as a decorated reference, and the Boolean synthesized attribute is split into a "partial" attribute (true if there is equality in one direction, this may be overridden) and a "total" attribute (true if there is equality in both directions.)  See https://github.com/melt-umn/meta-ocaml-lite/blob/feature/autoattr/grammars/edu.umn.cs.melt.metaocaml/abstractsyntax/Type.sv for an example.

This shouldn't be merged until #375 and #333 have been merged.

The next step here is to use these new types of attributes in the Silver compiler, however we should hold off on introducing a circular dependency with this branch until the preceding pull requests have been merged as well.

### DIFF
--- a/grammars/silver/extension/autoattr/DclInfo.sv
+++ b/grammars/silver/extension/autoattr/DclInfo.sv
@@ -95,12 +95,66 @@ top::DclInfo ::= sg::String sl::Location inh::String syn::String
   top.propagateDispatcher = propagateEqualitySyn(inh, _, location=_);
 }
 
+abstract production unificationInhDcl
+top::DclInfo ::= sg::String sl::Location fn::String tyVar::TyVar
+{
+  top.sourceGrammar = sg;
+  top.sourceLocation = sl;
+  top.fullName = fn;
+
+  top.typerep = varType(tyVar);
+  top.dclBoundVars = [tyVar];
+  top.isInherited = true;
+  
+  top.decoratedAccessHandler = inhDecoratedAccessHandler(_, _, location=_);
+  top.undecoratedAccessHandler = accessBounceDecorate(inhDecoratedAccessHandler(_, _, location=_), _, _, _); -- TODO: should probably be an error handler! access inh from undecorated?
+  top.attrDefDispatcher = inheritedAttributeDef(_, _, _, location=_); -- Allow normal inh equations
+  top.attributionDispatcher = unificationInhAttributionDcl(_, _, _, _, location=_); -- Same as functor, except decorated
+  top.propagateDispatcher = propagateEqualityInh(_, location=_);
+}
+
+abstract production unificationSynPartialDcl
+top::DclInfo ::= sg::String sl::Location inh::String synPartial::String syn::String
+{
+  top.sourceGrammar = sg;
+  top.sourceLocation = sl;
+  top.fullName = synPartial;
+
+  top.typerep = boolType();
+  top.dclBoundVars = [];
+  top.isSynthesized = true;
+  
+  top.decoratedAccessHandler = synDecoratedAccessHandler(_, _, location=_);
+  top.undecoratedAccessHandler = accessBounceDecorate(synDecoratedAccessHandler(_, _, location=_), _, _, _);
+  top.attrDefDispatcher = synthesizedAttributeDef(_, _, _, location=_); -- Allow normal syn equations
+  top.attributionDispatcher = defaultAttributionDcl(_, _, _, _, location=_);
+  top.propagateDispatcher = propagateUnificationSynPartial(inh, _, syn, location=_);
+}
+
+abstract production unificationSynDcl
+top::DclInfo ::= sg::String sl::Location inh::String synPartial::String syn::String
+{
+  top.sourceGrammar = sg;
+  top.sourceLocation = sl;
+  top.fullName = syn;
+
+  top.typerep = boolType();
+  top.dclBoundVars = [];
+  top.isSynthesized = true;
+  
+  top.decoratedAccessHandler = synDecoratedAccessHandler(_, _, location=_);
+  top.undecoratedAccessHandler = accessBounceDecorate(synDecoratedAccessHandler(_, _, location=_), _, _, _);
+  top.attrDefDispatcher = synthesizedAttributeDef(_, _, _, location=_); -- Allow normal syn equations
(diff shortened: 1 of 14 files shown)


krame505:
Since this doesn't interfere with origin tracking and I don't anticipate further work on this soon, I am going to merge this today as we discussed.

I added @ericvanwyk as a reviewer, but I will go ahead and merge as soon as the build server is free.

ericvanwyk (approved):

PR #379, opened 2020-09-23
Bugfix/incorrect matching order

### LINKED ISSUE #371: Incorrect matching
Consider the following match:
```
case left(3), left(4) of
| _, left(_) -> "right"
| left(_), _ -> "wrong"
end
```
The result of this should be `"right"`, but the result I am getting is `"wrong"`.

It appears that the clauses in the pattern matching are being compiled into a match in the wrong order.  I can look at this in the future, but I don't have the time right now.

RandomActsOfGrammar:
This fixes #371, that pattern matching was working left-to-right rather than top-to-bottom, as pattern matching works in other programming languages.

Pattern matching was originally compiling cases where the first match mixed both variable and constructor cases by putting all the constructor cases at the beginning, only considering the variable cases when no constructor case matched.  This meant in
```
case left(3), left(4) of
| _, left(_) -> "right"
| left(_), _ -> "wrong"
end
```
the `| left(_), _ -> "wrong"` match rule was pulled to the beginning of the compiled match, with the other rule put at the end because it starts with matching on a variable.

The solution was to follow Section 5.2.6 in [The Implementation of Functional Programming Languages](https://www.microsoft.com/en-us/research/uploads/prod/1987/01/slpj-book-1987-r90.pdf).  This compiles the mixed case by ordered segments of the match rules, with each segment having the same type of pattern (variable or constructor).  The `case` expression above would yield two segments, one with each match rule, with the segments in the same order as the rules occur in the written code.  Since compilation goes by segments of the match rules in order, rather than partitioning all match rules into lists with the same type of pattern, it respects the top-to-bottom order of the written pattern matching.  It builds each segment into its own `case` with the match failure being the `case`s for the rest of the segments.  The above `case` would take a step toward compilation resulting in (taking notational liberties) something like
```
case left(3), left(4) of
| _, left(_) -> "right"
else case left(3), left(4) of
     | left(_), _ -> "wrong"
     end
end
```
which respects the order in which the match rules were written.

One concern brought up in #371 was that fixing this might cause a blow-up in the size of the generated code.  There is some increase in size because it can build quite a few `case` constructs (in most code people write, this will probably be ~2),  but I don't think there is a large blow-up.  It doesn't duplicate any of the RHS expressions.

### DIFF
--- a/grammars/silver/extension/patternmatching/Case.sv
+++ b/grammars/silver/extension/patternmatching/Case.sv
@@ -160,7 +160,7 @@ top::Expr ::= es::[Expr] ml::[AbstractMatchRule] failExpr::Expr retType::Type
 function initialSegmentPatternType
 Pair<[AbstractMatchRule] [AbstractMatchRule]> ::= lst::[AbstractMatchRule]
 {
-  {-return case lst of
+  return case lst of
            --this probably shouldn't be called with an empty list, but catch it anyway
          | [] -> pair([], [])
          | [mr] -> pair([mr], [])
@@ -171,17 +171,7 @@ Pair<[AbstractMatchRule] [AbstractMatchRule]> ::= lst::[AbstractMatchRule]
               in pair(mr1::sub.fst, sub.snd) end
            else --the first has a different type of pattern than the second
               pair([mr1], mr2::rest)
-         end;-}
-  return if null(lst) --this probably shouldn't be called with an empty list, but catch it anyway
-         then pair([], [])
-         else if null(tail(lst))
-              then pair(lst, [])
-              else if head(lst).isVarMatchRule == head(tail(lst)).isVarMatchRule
-                   then --both have the same type of pattern
-                      let rest::Pair<[AbstractMatchRule] [AbstractMatchRule]> = initialSegmentPatternType(tail(lst))
-                      in pair(head(lst)::rest.fst, rest.snd) end
-                   else --the first has a different type of pattern than the second
-                      pair([head(lst)], tail(lst));
+         end;
 }
 
 {-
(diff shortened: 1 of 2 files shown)


grammars/silver/extension/patternmatching/Case.sv
```diff
@@ -147,17 +147,55 @@ top::Expr ::= es::[Expr] ml::[AbstractMatchRule] failExpr::Expr retType::Type
...
+  local mixedCase :: Expr = buildMixedCaseMatches(es, ml, failExpr, retType, top.location);
+}
+
+
+--Get the initial segment of the match rules which all have the same
+--pattern type (constructor or var) and the rest of the rules
+function initialSegmentPatternType
+Pair<[AbstractMatchRule] [AbstractMatchRule]> ::= lst::[AbstractMatchRule]
+{
+  return if null(lst) --this probably shouldn't be called with an empty list, but catch it anyway
```
krame505, line 163: Style nit: this could probably be much cleaner as a pattern match

krame505 (approved):
Looks good to me, pending Jenkins passing.  I can't access coldpress right now though for some reason.

grammars/silver/extension/patternmatching/Case.sv
RandomActsOfGrammar: I was trying to avoid using pattern matching in defining pattern matching.  I'm not sure if that is necessary, but it feels strange enough I didn't want to do it.

krame505: Eh, I don't see a strong reason to do so - Silver is already self-bootstrapped in lots of ways.

RandomActsOfGrammar: I just pushed a commit changing this to pattern matching.

krame505: Looks good to me.

PR #382, opened 2020-10-21
Remove an unsound strategy expression optimization

krame505:
I just realized while working on the paper that one of the rules for optimizing congruences is actually invalid - while a congruence containing a strategy that always fails will always fail, a strategy containing all identity strategies is not equivalent to identity, as it will still fail when applied to any non-matching production.  Fortunately I don't think this actually made a difference in any of our examples.


krame505:
@ericvanwyk could you approve this when you get a chance?

ericvanwyk (approved):

ericvanwyk:
@krame505 Done!

Issue #76, opened 2016-06-06
Improved support for pattern matching on primitives


tedinski:
I was going through some old starred emails, cleaning out the junk. Here's one we never got around to:

1) We should be able to pattern match on terminals. Possibly something like `case '+' of terminal("+") -> ... end`.

2) Right now, we have primitives (sting, bool, list) polluting the *modification* primitive match. These should probably be 100% taken care of by the *extension* implementing nested pattern matching. That way, primitive match is only about productions.

3) We lack to ability to pattern match on annotation values. This should be fixed. In the extension module as well, despite it appearing to be about productions. That is `case t of foo(x, y, location=l) -> ...` should probably translate to a primitive match on foo with `let l = t.location in ...` added to the RHS expression. This enhancement is related to issue #32, about enhancements to annotations.

N.B. if we choose to eventually allow named parameters besides annotations on productions, then we have to figure out how to implement that w.r.t. patterns because the above approach won't work. (Because things are now per-production instead of nonterminal-wide.) Perhaps if we have that feature, we should push annotation matching down to the primitive match modification. One possibility: don't allow named parameters on productions! Instead, just allow positional parameters but we can supply them via their names. Then we would just be pattern matching on them via their names as well. e.g. `case t of foo(left=x, right=y) -> ... x ... y ... end`.  (However, this has the downside of introducing the names as part of the public interface. But maybe that's okay. It's just weird in conjunction with aspects allowing renaming. But perhaps we should warn if aspect try to rename. :))

krame505:
Terminal patterns would be nice to have for #321 - see Ted's comment there.  The terminal pattern syntax should probably also include the location, but I don't think the type is actually needed.

counc009:
@krame505 Just want to check my understanding of this issue. If I understand what it's saying correct, we want to not use the primitive "patterns" in PrimitiveMatch.sv which directly generate Java translation and instead replace them with appropriate Silver expressions, so for instance if x is an `Integer` then
```
  case x of
  | 4 -> E0
  | 5 -> E1
  | _ -> E2
  end;
```
would become
```
let genName :: Integer = x in -- some generated name
if genName == 4 then E0
else if genName == 5 then E1
else E2
end
```
(obviously this is roughly speaking, what I mean is more that the `caseExpr` abstract production that is equivalent to the case statement above should produce the code I gave, roughly speaking the production invoked would be `caseExpr([x], [4 -> E0, 5 -> E1], E2, Type)`) 

[...]

PR #383, opened 2020-11-02
Fix a bug with match rule translation causing a crash when the last rule had too few patterns

krame505:


RandomActsOfGrammar:
Can you explain what problem this is supposed to fix?  The only production for `AbstractMatchRule` is `matchRule`, so adding a default pattern where you did shouldn't change anything.

krame505:
Yes, but it is matching on `headPat :: restPat` as its first argument which only matches a non-empty list.  It seems that if too few patterns are provided in a rule, this pattern can fail to match.  Adding this case is a workaround that lets the error message be reported.

RandomActsOfGrammar:
I see; it was an incomplete match before you added that.  However, my (admittedly few) attempts to trigger the original code into crashing instead of giving an error message aren't working.  Do you have a small crashing example?

krame505:
Umm no, this was something I ran into when experimenting in ableC.  I don't remember exactly what the code was that triggered this, but I did test that this change fixed the crash.

krame505:
I think it was a case expression matching on two values, where the last case only had one pattern in the pattern list.

RandomActsOfGrammar:
The number of patterns is getting checked on the concrete productions of type `MatchRule`.  For this to cause an error, your example must be bypassing the concrete productions.  That should be fixed as well.

What error message do you get with this fix?  I'm not sure where that would be coming from, since the code looks like it just assumes there will be enough patterns after the concrete productions are gone.

I want to be sure this is the right fix, rather than adding some other check elsewhere.

krame505:
I am getting the expected error message from the concrete `MatchRule`, now.  The issue is that translation is still being demanded even when there are errors in the concrete `MRuleList` due to how pattern error checking is structured.  This was resulting in a crash in computing the translation, before the concrete MatchRule error message could be reported.

RandomActsOfGrammar:
That seems like something that should be fixed, but that is another issue.

RandomActsOfGrammar (approved):

krame505:
The issue stems from how pattern matching is poorly divided into an extension and a modification, such that most (but not all) of the checking happens on what the extension forwards to.  Really the extension should handle all of the checking and the modification should just be there providing primitives to deal with productions, probably.  This is complicated a bit by needing to support GADT productions, I think.  

We already have a few open issues that touch on this (e.g. #76.)

PR #384, opened 2020-11-05
Implicit Monads

RandomActsOfGrammar:
This allows monads to be used implicitly in equations.

It also adds an example of the simply-typed lambda calculus to the tutorials which makes use of implicit monads.


grammars/core/monad/Either.sv
```diff
@@ -16,3 +16,20 @@ Either<a b> ::= x::b
 {
   return right(x);
 }
+
```
krame505, line 19: These type-specific functions will probably all go away once we have type classes.  But probably ok to add them for now.

grammars/core/monad/State.sv
```diff
@@ -38,6 +38,13 @@ top::State<s Unit> ::= newState::s
   top.stateVal = unit();
 }
 
+abstract production modifyState
```
krame505, line 41: Could forward to `bindState(getState(), \ x::s -> setState(fun(x)))`, I suppose.  But as-is is probably a little more efficient, IDK.

grammars/silver/composed/Default/Main.sv
```diff
@@ -4,6 +4,39 @@ import silver:host;
 
 parser svParse::Root {
   silver:host;
+
```
krame505, line 7: Why are you explicitly re-including these?  All of them should be exported by `silver:host` (`silver:host:core` is just the core language without any extensions or modifications enabled.)  The point is that we don't want to keep this list in more than one place.

grammars/silver/definition/core/GlobalDcl.sv
```diff
@@ -5,7 +5,8 @@ import silver:definition:flow:driver only ProductionGraph, FlowType, constructAn
 concrete production globalValueDclConcrete
 top::AGDcl ::= 'global' id::Name '::' t::TypeExpr '=' e::Expr ';'
 {
-  top.unparse = "global " ++ id.unparse ++ " :: " ++ t.unparse ++ " = " ++ e.unparse ++ "\n"; 
+  top.unparse = "global " ++ id.unparse ++ " :: " ++ t.unparse ++ " = " ++ e.unparse ++ ";\n";
+  top.errors := t.errors ++ e.errors;
```
krame505, line 9: Why the explicit equation for `errors` here?  This is auto-computed as a monoid attribute.

grammars/silver/definition/flow/env/ProductionDcl.sv
```diff
@@ -4,6 +4,23 @@ import silver:modification:defaultattr;
 import silver:definition:flow:driver only ProductionGraph, findProductionGraph;
 import silver:driver:util; -- only for productionFlowGraphs occurrence?
 
+{-
```
krame505, line 7: Delete?  We generally don't want to keep commented out chunks of obsolete code sitting around.

grammars/silver/extension/implicit_monads/BuiltInFunctions.sv
```diff
@@ -0,0 +1,240 @@
...
+  e.mDownSubst = top.mDownSubst;
+  top.mUpSubst = e.mUpSubst;
+  top.mtyperep = if !isList && isMonad(e.mtyperep)
+                 then monadOfType(e.mtyperep, intType())
+                 else intType();
+
+  e.monadicallyUsed = !isList && isMonad(e.mtyperep);
+  top.monadicNames = e.monadicNames;
+
+  top.monadRewritten = if !isList && isMonad(e.mtyperep)
```
krame505, line 24: A bit of a nit-pick, but stylistically we generally always line break and indent after `=` when there are long equations, rather than aligning the whole body after the `=`.  Don't feel obligated to change it though.

grammars/silver/extension/implicit_monads/BuiltInFunctions.sv
```diff
@@ -0,0 +1,240 @@
...
+  e.mDownSubst = top.mDownSubst;
+  top.mUpSubst = e.mUpSubst;
+  top.mtyperep = e.mtyperep;
+  e.monadicallyUsed = false;
+  top.monadicNames = e.monadicNames;
+  top.monadRewritten = newFunction('new', '(', e.monadRewritten, ')', location=top.location);
+}
+
+{-
+  TODO I don't know what these are supposed to be, really, but I suppose they should be bound in if they're monads?
```
krame505, line 193: This is the explicit terminal value constructor (e.g. `terminal(Identifier_t, "foo", bogusLoc)`) - this should be handled just like any other operator or literal, I suppose.

grammars/silver/extension/implicit_monads/BuiltInFunctions.sv
```diff
@@ -0,0 +1,240 @@
...
+  top.mUpSubst = top.mDownSubst;
+  top.mtyperep = t.typerep;
+  top.monadicNames = [];
+  top.monadRewritten = error("terminalConstructor monadRewritten not defined");
+}
+
+
+--------------------------------------------------------------------------------
+-- Deprecated variants of built-in functions
+
```
krame505, line 208: No need to include aspect equations for forwarding productions?

grammars/silver/extension/implicit_monads/Case.sv
```diff
@@ -0,0 +1,399 @@
...
+{
+  ml.mDownSubst = top.mDownSubst;
+  local monadInExprs::Boolean =
+    monadicallyUsedExpr(es.rawExprs, top.env, ml.mUpSubst, top.frame,
+                        top.grammarName, top.compiledGrammars, top.config, top.flowEnv,
+                        top.expectedMonad);
+  local monadInClauses::Boolean =
+    foldl((\b::Boolean a::AbstractMatchRule ->
+            b ||
+            let ty::Type = decorate a with {env=top.env; mDownSubst=ml.mUpSubst; frame=top.frame;
```
krame505, line 26: Why not just extract the match rule expression here and decorate that instead of decorating the entire match rule?

grammars/silver/extension/implicit_monads/Case.sv
```diff
@@ -0,0 +1,399 @@
...
+
+  ml.mDownSubst = top.mDownSubst;
+  local monadInExprs::Boolean =
+    monadicallyUsedExpr(es.rawExprs, top.env, ml.mUpSubst, top.frame,
+                        top.grammarName, top.compiledGrammars, top.config, top.flowEnv,
+                        top.expectedMonad);
+  local monadInClauses::Boolean =
+    foldl((\b::Boolean a::AbstractMatchRule ->
+            b ||
+            let ty::Type = decorate a with {env=top.env; mDownSubst=ml.mUpSubst; frame=top.frame;
```
krame505, line 277: Ditto.

grammars/silver/extension/implicit_monads/Case.sv
```diff
@@ -0,0 +1,399 @@
...
+top::PatternList ::=
+{
+--  top.errors := [];
+
+  top.patternTypeList = [];
+}
+
+
+
+attribute env, mDownSubst, merrors, mtyperep, frame, grammarName, compiledGrammars,
```
krame505, line 380: These would all cause orphaned-occurs flow error - technically I think you should define all new attributes with the same types and use them here.  But as mentioned above, I think you can just add an attribute to extract the match rule expression and decorate that instead, since these are all just copied down anyway?

grammars/silver/extension/implicit_monads/Case.sv
```diff
@@ -0,0 +1,399 @@
...
+aspect production matchRuleWhen_c
+top::MatchRule ::= pt::PatternList 'when' cond::Expr arr::Arrow_kwd e::Expr
+{
+  top.patternTypeList = pt.patternTypeList;
+}
+
+aspect production patternList_one
+top::PatternList ::= p::Pattern
+{
+--  top.errors := p.errors;
```
krame505, line 358: Generally we would like to avoid leaving lots of commented out equations, unless there is a reason we would like to include them but can't, in which case there should be a justifying comment.

grammars/silver/extension/implicit_monads/CopperExpr.sv
```diff
@@ -0,0 +1,68 @@
+grammar silver:extension:implicit_monads;
+
+{-
+  I honestly have no idea if these can occur in a production.  I kind of doubt
```
krame505, line 4: These can only occur in copper action blocks, which reuse the same syntax as production bodies and expressions.  I don't think you would be applying any rewriting inside action blocks, but I guess it's still good to include these for completeness.

grammars/silver/extension/implicit_monads/DclInfo.sv
```diff
@@ -0,0 +1,78 @@
+grammar silver:extension:implicit_monads;
+
+
+abstract production restrictedSynDcl
```
krame505, line 4: Technically these should all be non-forwarding since `DclInfo` is a closed nonterminal - doing so avoids interfering equations, too.  See https://github.com/melt-umn/silver/blob/develop/grammars/silver/extension/autoattr/DclInfo.sv for example.

grammars/silver/extension/implicit_monads/Expr.sv
```diff
@@ -0,0 +1,2491 @@
...
+
+  local mty::Type = head(nes.monadTypesLocations).fst;
+  --need to check that all our monads match
+  top.merrors <- if null(nes.monadTypesLocations) ||
+                   foldr(\x::Pair<Type Integer> b::Boolean -> b && monadsMatch(mty, x.fst, ne.mUpSubst).fst, 
+                         true, tail(nes.monadTypesLocations))
+                then []
+                else [err(top.location,
+                      "All monad types used monadically in a function application must match")];
+  --need to check it is compatible with the function return type
```
krame505, line 183: Commented out code - is this important?

grammars/silver/extension/implicit_monads/Expr.sv
```diff
@@ -0,0 +1,2491 @@
...
+                       then if isMonad(e2.mtyperep)
+                            then bindBoth
+                            else bind1
+                       else if isMonad(e2.mtyperep)
+                            then bind2
+                            else neq(e1.monadRewritten, '!=', e2.monadRewritten, location=top.location);
+}
+
+concrete production ifThen
+top::Expr ::= 'if' e1::Expr 'then' e2::Expr 'end' --this is easier than anything else to do
```
krame505, line 1568: Seems like a reasonable syntactic decision.

grammars/silver/extension/implicit_monads/Lambda.sv
```diff
@@ -0,0 +1,37 @@
+grammar silver:extension:implicit_monads;
+
+--import silver:definition:flow:ast only ExprVertexInfo, FlowVertex;
+
+--- Concrete Syntax for lambdas
```
krame505, line 5: Uhh abstract syntax for lambdas?

grammars/silver/extension/implicit_monads/Let.sv
```diff
@@ -0,0 +1,166 @@
+grammar silver:extension:implicit_monads;
+
+--import silver:definition:flow:ast only ExprVertexInfo, FlowVertex;
+
+--- Concrete Syntax for lets
```
krame505, line 5: Ditto

grammars/silver/extension/implicit_monads/PrimitiveMatch.sv
```diff
@@ -0,0 +1,478 @@
...
+          returnFun, returnify<PrimPattern> occurs on PrimPattern;
+
+--returnFun is the monad's defined Return for returnify
+inherited attribute returnFun::Expr;
+synthesized attribute returnify<a>::a;
+--type matched by patterns; provided by Case.sv
+--synthesized attribute patternType::Type;
+
+
+aspect production matchPrimitiveConcrete
```
krame505, line 17: Remove empty aspects?

grammars/silver/extension/implicit_monads/ProductionBody.sv
```diff
@@ -0,0 +1,230 @@
...
+
+
+
+
+
+--take a list of unallowed attributes and generate error messages for them
+function buildExplicitAttrErrors
+[Message] ::= l::[Pair<String Location>]
+{
+  return case l of
```
krame505, line 142: Could just be a `map`, I think?  Took me a second wondering why this was recursive.

grammars/silver/extension/implicit_monads/ProductionDcl.sv
```diff
@@ -0,0 +1,51 @@
+grammar silver:extension:implicit_monads;
```
krame505, line 1: Delete?

grammars/silver/extension/implicit_monads/ProductionBody.sv
```diff
@@ -0,0 +1,230 @@
...
+  forwards to if null(merrors)
+              then attributeDef(dl, '.', attr, '=',
+                                case fail of | right(e) -> e end,
+                                ';', location=top.location)
+              else errorProductionStmt(merrors, location=top.location);
+}
+
+
+--Write a marked equation that allows implicit use of monads
+terminal Implicit_kwd    'implicit'     lexer classes {KEYWORD,RESERVED};
```
krame505, line 42: Nit: this terminal declaration should go before `emptyAttributeDef`.  

Also have you checked that these actually need to be reserved?  Generally we want to avoid introducing new reserved keywords and lean on context-aware scanning whenever possible.

grammars/silver/extension/implicit_monads/Types.sv
```diff
@@ -0,0 +1,124 @@
+grammar silver:extension:implicit_monads;
+
+
```
krame505, line 3: Hmm, I thought we decided to go with lower case for these?  Upper case is inconsistent with ...every other declaration keyword in Silver.  I realize that these are kind of like types, but these aren't actually type expressions so having them upper case would be confusing.  The convention we have adopted for `functor`, `monoid`, etc. is lower case too, despite these being "sort of" types.

grammars/silver/extension/implicit_monads/Types.sv
```diff
@@ -0,0 +1,124 @@
...
+terminal Explicit_t 'Restricted';
+terminal Implicit_t 'Implicit';
+terminal Unrestricted_t 'Unrestricted';
+
+
+abstract production explicitType
```
krame505, line 9: Is the purpose of these types just to give nicer error messages?  Neat idea.  I would maybe just change the pp to `"restricted " ++ t.typepp` or `"implicit " ++ t.typepp` though, having comment symbols show up in error messages seems potentially confusing (and not especially helpful.)

grammars/silver/extension/implicit_monads/Types.sv
```diff
@@ -0,0 +1,124 @@
...
+}
+
+
+
+
+
+
+
+
+
```
krame505, line 42: These productions don't actually need to have interfering equations for `defs`, you can just forward to `defsAGDcl` - see https://github.com/melt-umn/silver/blob/develop/grammars/silver/extension/autoattr/Functor.sv#L3 for example.

grammars/silver/extension/implicit_monads/Util.sv
```diff
@@ -0,0 +1,325 @@
...
+           right(Silver_Expr { core:monad:failList($Expr{string}) })
+           --baseExpr(qNameId(name("failList", l), location=l), location=l)
+         | decoratedType(t) -> monadFail(t, l)
+         | _ ->
+           error("Tried to get the fail for a non-monadic type at " ++ l.unparse)
+         end;
+}
+
+
+function monadPlus
```
krame505, line 279: Shouldn't these functions be *monoid* plus and zero?  These operations aren't really connected directly to monads AFAIK, they just are available for monads that also happen to be monoids.

krame505 (changes requested):
Wow, there's a lot more here that I thought!  (Having not really looked at what you were working on in this branch too closely.)

Just a (relatively) few things to fix/discuss.

grammars/silver/composed/Default/Main.sv
RandomActsOfGrammar: This is probably from when I started last year and didn't understand as much about how it worked.  I'll change this file to the `develop` version.

grammars/silver/definition/core/GlobalDcl.sv
RandomActsOfGrammar: I'll get rid of it.  This is leftover from my original attempt.

grammars/silver/definition/flow/env/ProductionDcl.sv
RandomActsOfGrammar: Agreed.

grammars/silver/extension/implicit_monads/BuiltInFunctions.sv
RandomActsOfGrammar: I have those all over the place, so I won't change it.  I will try to keep it in mind for the future.

grammars/silver/extension/implicit_monads/BuiltInFunctions.sv
RandomActsOfGrammar: I'll remove them.

grammars/silver/extension/implicit_monads/BuiltInFunctions.sv
RandomActsOfGrammar: I added actual rewriting for it.

grammars/silver/extension/implicit_monads/DclInfo.sv
RandomActsOfGrammar: I changed them to be non-forwarding and copied the equations for the other attributes from the existing `DclInfo` productions.

grammars/silver/extension/implicit_monads/Lambda.sv
RandomActsOfGrammar: Removed---I'm guessing it came from copying the original file.

grammars/silver/extension/implicit_monads/Let.sv
RandomActsOfGrammar: Ditto.

grammars/silver/extension/implicit_monads/ProductionBody.sv
RandomActsOfGrammar: It could be a `map`.

grammars/silver/extension/implicit_monads/ProductionDcl.sv
RandomActsOfGrammar: Yes.  This is from my initial attempt, which rewrote in monadic productions, rather than particular equations.

grammars/silver/extension/implicit_monads/Util.sv
RandomActsOfGrammar: We could change them to that.  I don't see any point to changing them; we shouldn't need them once we have type classes, and they are for monadification right now.

grammars/silver/extension/implicit_monads/Types.sv
RandomActsOfGrammar: These types are how implicit and restricted attributes are actually tracked to ensure attribute access restrictions are followed.  For example, implicit equations check that any attributes accessed inside them have an `implicit` type or a `restricted` type.

The comment in the `typepp` was for debugging purposes when I couldn't figure out why it was doing what it was doing.  I'll remove it.

grammars/silver/extension/implicit_monads/Types.sv
RandomActsOfGrammar: I removed the equations for `defs`.

grammars/silver/extension/implicit_monads/Types.sv
krame505: Hmm, that is potentially brittle- generally pattern matching on a forwarding production is the same as writing an interfering equation. 
For attribute accesses, couldn't you just check the `DclInfo` that you get from looking up the attribute to see if it is restricted or implicit? You can define new attributes/match on your new non-forwarding `DclInfo` productions without issue since that nonterminal is closed and you have a default (i.e. unrestricted.)

grammars/silver/extension/implicit_monads/ProductionBody.sv
RandomActsOfGrammar: If I remove `Reserved`, I get the following Copper error for all three:
```
Unresolvable lexical ambiguity at parser states   [191,639,769,906,1091,1461] between terminals: 
       [silver:definition:core:IdLower_t,'implicit']
```

grammars/silver/extension/implicit_monads/Types.sv
RandomActsOfGrammar: I'm changing it to use the lowercase terminals.

grammars/silver/extension/implicit_monads/ProductionBody.sv
krame505: Hmm, nevermind - I forgot that these are used inside production bodies (where there would be an ambiguity with equation LHS names.)

RandomActsOfGrammar:
I think I addressed all your comments, Lucas.  However, I'm having a problem with marked equations, so this shouldn't be merged.

When I try to write marked equations for `restricted` or `implicit` attributes, I end up with an infinite computation.  For example, the following grammar does not work:
```
imports core:monad;

restricted synthesized attribute res1::Integer;

nonterminal Nt1 with res1;

abstract production p
top::Nt1 ::=
{
  restricted top.res1 = 5;
}
```
There isn't any reason I can see why this shouldn't work.  The affected productions are `implicitAttributeDef` and `restrictedAttributeDef` in `ProductionBody.sv`.  I thought the problem was with accessing `attr.attrDcl`, but empty equations (`implicit t.a = ;` from production `emptyAttributeDef`) work just fine, and they are essentially the same as other implicit equations.

I have no idea what else to look at to figure out what the problem is.  Lucas, do you have any ideas?

krame505:
Wait, I thought you were getting rid of the various explicitly-marked concrete equation productions, now that equations are overloaded?  Keeping them seems overly complicated, since one would presumably never e.g. have an `implicit` equation for an attribute declared as `restricted`.  Removing these could also allow for the `restricted` and `unrestricted` terminals to not be reserved.  

Also it seems you are doing error checking on the concrete productions - what is the reason for this?  I would think that you would want to perform these checks on the abstract productions to which the overloaded equations would dispatch.  I haven't looked too closely, but this may fix the infinite recursion issues you are seeing.

RandomActsOfGrammar:
By keeping the markers in addition to overloading equations, we can maintain consistency with the paper.

The only error checking I'm doing on the concrete productions is that they're being used for the correct mode of attribute, such as checking that an implicit equation is defining an implicit attribute.  The rest of the checking is being done on the abstract productions.

krame505:
Ah, that seems reasonable.  

The problem seems to be that you are forwarding based on the result of looking up the attribute in the environment, which eventually depends on `defs` for all `ProductionStmt`s in the production body (due to the mutually recursive nature of production bodies.)  This introduces a circularity since the `defs` are attempted to be computed via forwarding in `implicitAttributeDef`, etc.  See the comment on `attributeDef` in `silver:definition:core:ProductionBody.sv`.  You should be able to fix this by providing explicit empty equations for `defs` (and also `productionAttributes`, for similar reasons) as on `attributeDef`.  This isn't interfering, since these attributes should always be empty anyway based on what is forwarded to.

RandomActsOfGrammar:
Thank you.  Yes, that fixed it.  If you want to take another look, I think it should meet with your approval.

RandomActsOfGrammar:
It occurred to me that the mode types were no longer necessary.  I removed them and added a missing check for not redefining attribute names, with other error checking I found on the normal attribute-defining productions.

krame505 (approved):

Issue #185, opened 2017-07-10
Refactor Silver's type system implementation


tedinski:
I wanted to dump this out of my head.

Silver's type system implementation could use a few improvements. This is not new features (yet) but might be necessary to do so.

First, a number of things are misnamed! We should get a renaming pass done:

```
Type -> TypeExpr
TypeList -> TypeExprs
TypeExp -> Type
```

Second, a number of type are poorly implemented. We should embraced "foreign types" for Lists, instead of the current... awfulness. (Seriously, don't look at the list extension in silver unless you want to immediate destroy it and re-do it.) This may be difficult without first solving https://github.com/melt-umn/silver/issues/25 ? Not sure.

Third, *probably* a big performance problem with the Silver compiler is that we do Unification... poorly. Right now there's a `Substitutions` list, which just screams O(N^2). We could consider alternatives. I've thought about a magically mutating `UnificationContext` type...

Finally, the major improvement is to introduce `PolyType` in addition to `Type` (which prior to the rename is currently called `TypeExp`). This does what its name suggests: represents a type that quantifies over some type variables. Then we need to actually use `PolyType` in the appropriate places. Figuring out how to do this correctly will require some good knowledge of how type systems work.

But presently, there's a few obvious places where we should be using a `PolyType`:

https://github.com/melt-umn/silver/blob/develop/grammars/silver/definition/env/DclInfo.sv#L198

```
abstract production ntDcl
top::DclInfo ::= sg::String sl::Location fn::String bound::[TyVar] ty::TypeExp ...
```

Here, `bound` are the bound variables in `ty`, so, uh, obvious change here. However, a little further up we have:

```
abstract production prodDcl
top::DclInfo ::= sg::String sl::Location ns::NamedSignature
```

This ought to be a PolyType... except that we're implicitly quantifying over all free variables of `ns`. This shouldn't be implicit. Making this non-implicit is part of what `PolyType` would be supposed to accomplish. Right now, because of this implicitness, we have to always remember to correctly call things:

```
abstract production productionReference
top::Expr ::= q::Decorated QName
{
...
  top.typerep = freshenCompletely(q.lookupValue.typerep);
}
```

The types should enforce this instead, because a `typerep` on an `Expr` should be a monotype, while the type we get when looking things up in the environment should (usually?) be a polytype.

[...]

PR #385, opened 2020-11-10
Refactor type system implementation to introduce PolyType

krame505:
This does one of the type system refactorings mentioned in #185, introducing `PolyType` to be explicit about what type variables are being quantified over, as a prerequisite to implementing type classes.  Instead of `DclInfo` items having a `typerep`, they now have a `typeScheme` (the common term for these in Hindley-Milner type systems) from which a freshened `typerep` may be accessed.  This eliminates all uses of `freshenCompletely`, as looked-up types are now automatically freshened to the appropriate degree as determined in the environment.  Not having to keep track of this manually wherever a type is used should help keep things more maintainable; this change uncovered a (minor) bug in error checking for strategy attributes.  

I think that this is essentially what @tedinski had in mind, but he can chime in if he wants if I missed something critical here.

### DIFF
--- a/grammars/silver/definition/env/Type.sv
+++ b/grammars/silver/definition/env/Type.sv
@@ -1,5 +1,22 @@
 grammar silver:definition:env;
 
+synthesized attribute boundVars :: [TyVar] occurs on PolyType;
+attribute typerep occurs on PolyType;
+
+aspect production monoType
+top::PolyType ::= ty::Type
+{
+  top.boundVars = [];
+  top.typerep = ty;
+}
+
+aspect production polyType
+top::PolyType ::= tvs::[TyVar] ty::Type
+{
+  top.boundVars = freshTyVars(length(tvs));
+  top.typerep = freshenTypeWith(ty, tvs, top.boundVars);
+}
+
 -- Just to clarify:
 -- call prettyType to pretty print the type.
 -- get typeName to find out what nonterminal a NT or DNT is
--- a/grammars/silver/definition/env/Type.sv
+++ b/grammars/silver/definition/env/Type.sv
@@ -1,13 +1,20 @@
 grammar silver:definition:env;
 
+-- Just to clarify:
+-- call prettyType to pretty print the type.
+-- get typeName to find out what nonterminal a NT or DNT is
+synthesized attribute typeName :: String;
+
 synthesized attribute boundVars :: [TyVar] occurs on PolyType;
 attribute typerep occurs on PolyType;
+attribute typeName occurs on PolyType;
 
 aspect production monoType
 top::PolyType ::= ty::Type
 {
   top.boundVars = [];
   top.typerep = ty;
+  top.typeName = ty.typeName;
 }
 
 aspect production polyType
@@ -15,15 +22,10 @@ top::PolyType ::= tvs::[TyVar] ty::Type
 {
   top.boundVars = freshTyVars(length(tvs));
   top.typerep = freshenTypeWith(ty, tvs, top.boundVars);
+  top.typeName = ty.typeName;
 }
 
--- Just to clarify:
--- call prettyType to pretty print the type.
--- get typeName to find out what nonterminal a NT or DNT is
-
 attribute typeName occurs on Type;
-
-synthesized attribute typeName :: String;
 
 aspect default production
 top::Type ::=
--- a/grammars/silver/definition/env/Type.sv
+++ b/grammars/silver/definition/env/Type.sv
@@ -5,23 +5,17 @@ grammar silver:definition:env;
 -- get typeName to find out what nonterminal a NT or DNT is
 synthesized attribute typeName :: String;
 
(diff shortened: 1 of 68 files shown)


grammars/silver/extension/implicit_monads/PatternTypes.sv
```diff
@@ -3,9 +3,9 @@ grammar silver:extension:implicit_monads;
 aspect production prodAppPattern
 top::Pattern ::= prod::QName '(' ps::PatternList ')'
 {
-  top.patternType = case prod.lookupValue.typerep of
+  top.patternType = case prod.lookupValue.typeScheme.typerep of
                     | functionType(out, _, _) -> out
-                    | _ -> prod.lookupValue.typerep
+                    | _ -> prod.lookupValue.typeScheme.typerep
```
tedinski, line 8: Is this right?

grammars/silver/definition/env/Type.sv
```diff
@@ -1,5 +1,22 @@
...
+{
+  top.boundVars = [];
+  top.typerep = ty;
+}
+
+aspect production polyType
+top::PolyType ::= tvs::[TyVar] ty::Type
+{
+  top.boundVars = freshTyVars(length(tvs));
+  top.typerep = freshenTypeWith(ty, tvs, top.boundVars);
```
tedinski, line 17: I see `.typeScheme.typerep` get used in a lot of places where it was just `.typerep` before, which means we're now freshening this type a whole lot, and we didn't used to. I'm not sure that's great for performance.

It might be nice to promote certain operations to `PolyType`. So e.g. `.typeScheme.typerep.isDecorable` could be `.typeScheme.isDecorable` and we can skip the freshening. (Ditto at least `typeName`, `isError`)

And as long as I'm looking at this diff, seems a lot of places where we might add an attribute like `asNtOrDecType` to compute `ntOrDecType(q.lookupValue.typeScheme.typerep, freshType());` for us.

One thing I'd note though: There's a lot of places right now that look like places where we're making assumptions that there's a `monoType`. I wonder if you might actually be making things harder later by making these freshen. Maybe an awful lot of these should be `.monoType` that does not freshen and dies if it comes upon a polyType, then you'll have all these locations easily findable and audit-able when you add more to PolyType.

But I'm not 100% about the above. It's just right now we have a clear signal on where polytypes are expected, but after this commit we wouldn't see that anymore. Might be a loss.

grammars/silver/extension/treegen/Arbitrary.sv
```diff
@@ -22,12 +22,12 @@ top::AGDcl ::= 'derive' 'Arbitrary' 'on' names::QNames ';'
...
 Boolean ::= l::DclInfo  r::DclInfo
 {
-  return length(l.typerep.inputTypes) <= length(r.typerep.inputTypes);
+  return length(l.typeScheme.typerep.inputTypes) <= length(r.typeScheme.typerep.inputTypes);
 }
 function prodDclInfoNumChildEq
 Boolean ::= l::DclInfo  r::DclInfo
 {
-  return length(l.typerep.inputTypes) == length(r.typerep.inputTypes);
+  return length(l.typeScheme.typerep.inputTypes) == length(r.typeScheme.typerep.inputTypes);
```
tedinski, line 30: I've seen a lot of places where we use `inputTypes` just to get an `arity`. That might also be nice to promote to `PolyType` as an attribute.

grammars/silver/definition/env/Type.sv
krame505: Hmm, good points - thanks.

FWIW, I didn't see a significant performance hit with this change, so I guess the extra freshening can't be too bad.  But I'll still try to clean this up a bit.

grammars/silver/extension/implicit_monads/PatternTypes.sv
krame505: Just updating what was there.  @RandomActsOfGrammar?

krame505:
Ok, I think I've finished making all the tweaks/improvements suggested by @tedinski, I think this is ready to merge now.  Can someone please approve this if it looks alright?

grammars/silver/extension/implicit_monads/PatternTypes.sv
RandomActsOfGrammar: It should be right, since pattern matching works.  Maybe the second part isn't needed, since the applied name should have a function type, but we need some result there.

krame505: I don't think that this works for GADT productions though - I'm pretty sure the looked-up production type needs to get unified with the types of the children.  But this discussion should be in a seperate issue as this isn't affected by the refactor in the pull request.

krame505:
@ericvanwyk could you please review this pull request?  This needs to be merged before #390 which you just reviewed.

ericvanwyk (approved):

ericvanwyk:
Sorry I missed this one.  Again, don't hesitate to email me to prod me on these.

PR #386, opened 2020-11-13
Refactor DclInfo to use annotations

krame505:
This does a bit more refactoring on `DclInfo` - previously every production on this nonterminal had the source location and grammar name as children that are available as attributes, this makes those values into annotations.  Doing so saves a bunch of unneeded equations for these attributes, and also simplifies the `DclInfo` production signatures making them more readable.  

This is based on top of feature/polytype, so merge that before reviewing this, I guess?


krame505:
@ericvanwyk seems this PR slipped through the cracks too, please take a look.

ericvanwyk (approved):

ericvanwyk:
@krame505 thanks for the ping.  Done.

PR #388, opened 2020-11-21
Fix outdated references to 'explicit' attributes in error messages

krame505:
I believe these should be "restricted" instead?


RandomActsOfGrammar (approved):
Looks good.

PR #390, opened 2020-11-23
Refactor type checking to use threaded attributes

krame505:
This cuts out a bunch of boilerplate by using threaded attributes for `upSubst`/`downSubst` throughout the typechecking code.  

This is based on feature/polytype, so review this after #385 is merged, I guess?

### DIFF
--- a/grammars/silver/analysis/typechecking/core/AspectDcl.sv
+++ b/grammars/silver/analysis/typechecking/core/AspectDcl.sv
@@ -22,8 +22,7 @@ top::AGDcl ::= 'aspect' 'production' id::QName ns::AspectProductionSignature bod
       end;
 
   ns.downSubst = emptySubst();
-  errCheck1.downSubst = ns.upSubst;
-  body.downSubst = errCheck1.upSubst;
+  thread downSubst, upSubst on ns, errCheck1, body;
   
   ns.finalSubst = errCheck1.upSubst;
 }
@@ -48,8 +47,7 @@ top::AGDcl ::= 'aspect' 'function' id::QName ns::AspectFunctionSignature body::P
       end;
 
   ns.downSubst = emptySubst();
-  errCheck1.downSubst = ns.upSubst;
-  body.downSubst = errCheck1.upSubst;
+  thread downSubst, upSubst on ns, errCheck1, body;
   
   ns.finalSubst = errCheck1.upSubst;
 }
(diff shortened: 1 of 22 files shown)


RandomActsOfGrammar:
It looks good to me, but I don't know anything about the `polytype`/`monotype` changes, so I'm not sure I should approve it.  It might be good if someone who knows about that takes a look as well.

grammars/silver/analysis/typechecking/core/AspectDcl.sv
```diff
@@ -111,18 +103,15 @@ top::AspectRHSElem ::= id::Name t::Type
 aspect production aspectFunctionSignature
 top::AspectFunctionSignature ::= lhs::AspectFunctionLHS '::=' rhs::AspectRHS 
 {
-  lhs.downSubst = top.downSubst;
-  rhs.downSubst = lhs.upSubst;
-  top.upSubst = rhs.upSubst;
+  propagate downSubst, upSubst;
```
ericvanwyk, line 106: What is the difference between `thread downSubst, upSubst on top, errCheck1, top` and `propagate downSubst, upSubst`?

grammars/silver/analysis/typechecking/core/BuiltInFunctions.sv
```diff
@@ -3,26 +3,12 @@ grammar silver:analysis:typechecking:core;
...
-aspect production stringLength
-top::Expr ::= e::Decorated Expr
-{
-  top.upSubst = top.downSubst;
-}
-aspect production errorLength
-top::Expr ::= e::Decorated Expr
-{
-  top.upSubst = top.downSubst;
+  propagate upSubst, downSubst;
```
ericvanwyk, line 6: Where were we sending the `downSubst` attribute down the `forward` tree?  Is this no longer happening with the new `Decorated Expr`?  Why change this to `Decorated Expr`?

ericvanwyk (approved):
I left a few questions in the comments that I'm curious about but otherwise it looks good.

grammars/silver/analysis/typechecking/core/AspectDcl.sv
krame505: `propagate` only threads the value through the children of the production - if we also want to include the local `errCheck1` then we can use the `thread ... on ...` declaration for more precise control.

grammars/silver/analysis/typechecking/core/BuiltInFunctions.sv
krame505: I think the way the diff is rendering is a bit confusing in this case - the attribute equations in the body of the `lengthFunction` production are replaced by `propagate upSubst, downSubst;` and the aspects for `stringLength` and `errorLength` are deleted.  

I believe that passing `downSubst` to the forward tree is done because the type checking is not repeated on what this production forwards to (as these type-specialized productions contain decorated references), so the type context resulting from checking `e` must still be available in the forward.  

The behavior of `propagate upSubst, downSubst;` on a forwarding production is to finally thread `downSubst` to the forward rather than defining an equation for `upSubst` - so we are still maintaining the same threading behavior with this change.  Aspects for the other productions are no longer required as they are covered by a global `propagate`, however global propagation does not generate equations for forwarding productions unless specified explicitly, so the aspect on `lengthFunction` is still needed.

PR #392, opened 2020-11-30
Fix flow errors in implicit monads

RandomActsOfGrammar:
This fixes the large number of flow errors introduced when implicit monad use was originally introduced.  There are still some flow errors because new constructs forward based on attributes introduced in the extension, but these seem to be benign since these constructs should only occur in implicit equations.


grammars/silver/extension/implicit_monads/Case.sv
```diff
@@ -320,36 +322,126 @@ top::Expr ::= 'case_any' es::Exprs 'of' vbar::Opt_Vbar_t ml::MRuleList 'end'
               end;
 }
 
+
+
+--There are several thing swe need for mtyperep on e which don't occur on match rules
```
krame505, line 327: typo

krame505 (approved):

Issue #366, opened 2020-05-24
Type class implementation


krame505:
Some more thoughts on implementing type classes, separated out from design discussion in #364.

Type classes and instances would be represented in seperate environment namespaces, similarly to how attributes and occurrences are represented in the environment.

Every type class would be translated as an abstract class in Java with an abstract "get" method for every member, and type class instances would be concrete sub-classes of the corresponding abstract class.  For example,
```haskell
class Monoid m {
  mempty :: m;
  mappend :: (m ::= m m);
}

instance Monoid Maybe<a> {
  mempty = nothing();
  mappend = orElse;
}
```
would translate as
```java
abstract class CMonoid<m> {
  public abstract m mempty();
  public abstract NodeFactory<m> mappend();
}

abstract class IMonoidMaybe extends CMonoid<NMaybe> {
   public NMaybe mempty() {
    return new Pnothing();
  }
  public NodeFactory<NMaybe> mappend() {
    return PorElse.factory;
  }
}
```

When a type class member value is referenced, we would look up an instance in the environment for the inferred type.  This would yield the code for constructing the appropriate instance object, on which the referenced member can then be accessed.  

Functions with type constraints on type variables in their signature would be passed the appropriate instance object as an implicit parameter.  Looking up an instance in the environment for a type variable would result in code to access the parameter instance object.

Default implementations in type classes could be implemented simply as default implementations in the abstract class translation.

Type constraints in class and constructor signatures would simply instance objects passed as constructor parameters to the abstract class or concrete implementation class, respectively.

ericvanwyk:
You have thought about this more than I have, but don't type classes naturally lend themselves to being Java interfaces in the implementation?  Where does that not work?

krame505:
Yes, that was actually my first thought.  But consider how things work for translating a type class with multiple super-classes (e.g. `class Foo a, Bar a => Baz a`.)
These would become interfaces `CFoo`, `CBar` and `CBaz`, where `CBaz` extends `CFoo` and `CBar`.  Instance of `Foo` and `Bar` for `Integer` would become classes `IFooInteger` and `IBarInteger` that implement the appropriate interfaces.

[...]

Issue #364, opened 2020-04-30
Type class design


krame505:
I thought we had an issue for this, but I guess not.  

This has been deferred for too long, and we are at the point where a lot of things would seriously benefit from having proper support for type classes.  This includes Dawn's implicit monad work, my work with term rewriting, and potential future development of Ted's work on randomized property testing.  Also there would be general benefits to the readability of Silver code (e.g. monoid attributes and do-notation wouldn't need to be parameterized by the monoid/monad instance functions to use, some common patterns could be made into generic utilities with type support for things like "anything with attribute pp", etc.)  I am considering taking a stab at this as a side-project this summer.  

A few thoughts/notes on this:
* Keeping things simple
  * We probably want to avoid as much of the complexity of higher-kind types as possible.  Require explicit kind signatures for type variables in constraints, maybe?
  * Maybe don't support default method implementations, at least initially.
  * We probably still want to eventually support chaining type classes into hierarchies, e.g. Functor -> Applicative -> Monad.  I don't think this itself adds much complexity.
* Quick syntax proposal:
```haskell
class Monad<m<*>> {
  bind::(m<b> ::= (m<b> ::= a) m<a>);
  pure::(m<a> ::= a);
}

instance Monad<Maybe> {
  bind = \ f::(Maybe<b> ::= a) Maybe<a> ->
    case a of just(x) -> f(x) | nothing() -> nothing() end;
  pure = just;
}

function seq
Monad<m<*>> => m<b> ::= x::m<a> y::m<b>
{ return bind(\ a -> y, x); }
```
* Attributes are themselves a restricted form of type class.  Once we have type constraints it should be straightforward-ish to support attributes in type constraints.  Presumably these constraints would need to include a (maximum) flow type.  For example:
```haskell
class Eq<a> {
  eq::(Boolean ::= a a);
}

inherited attribute isEqualTo<a>::a;
synthesized attribute isEqual::Boolean;
instance isEqualTo<a> occurs on a, isEqual {isEqualTo} occurs on a => Eq<a> {
  eq = \ x::a y::a -> decorate x with { isEqualTo = y; }.isEqual;
}
```
* As another example, `pp {} occurs on a` could be used in place of a `Show` type class.
* Standard library type classes can/should probably mirror Haskell, to a large extent.  Infix operators such as `==`,  `<>` and `>>=` can be made language extensions.
* Implementation: I haven't given this much thought yet.  Presumably #185 needs to get finished first.
[...]

PR #397, opened 2020-12-20
Implement type classes

krame505:
This implements the first step of type classes as outlined in #364 and #366.  There are some test cases in https://github.com/melt-umn/silver/tree/feature/typeclass/test/silver_features/TypeClasses.sv and https://github.com/melt-umn/silver/blob/feature/typeclass/test/silver_features/myeq/MyEq.sv

This is pretty much just the basic features of type classes in Haskell, plus the `FlexibleInstances` extension which by itself isn't unsafe and may be useful for integrating type classes and attributes.  This basically lets one write instances that are not for a type applied only to type variables, e,g, `instance Foo a` or `instance Foo Pair<Integer a>`.   Note that I am not supporting `UndecidableInstances` which makes things ... undecidable.  This means that all type class constraints must still be of the form `Foo a`, and the constraints must be smaller than the instance head (i.e. no `instance Foo a => Bar a`.)  

Next up is probably support for higher-kinded type classes, e.g. `Functor` or `Monad`, which may require some deeper tinkering with how types are translated - I need to think about this first.  

With this pull request there should also be everything needed to implement `Eq` and `Monoid` in the standard library - I will start on that soon as well.


grammars/silver/definition/core/ClassDcl.sv
```diff
@@ -0,0 +1,126 @@
...
+  top.errors <-
+    if indexOf(":", id.name) == -1 then []
+    else [err(id.location, "Class name must be unqualified.")];
+
+  -- Here we ensure that the type is just a type *variable*
+  top.errors <- var.errorsTyVars;
+  
+  -- Redefinition check of the name
+  top.errors <- 
+    if length(getTypeDclAll(fName, top.env)) > 1
```
ericvanwyk, line 29: In a lazy language like Silver doesn't `length` require looking through the entire `env` to identify all matches.  Why not use a `case` with `[]` and `_ :: _` patterns.  This would then stop looking at finding the first match.

ericvanwyk (approved):

grammars/silver/definition/core/ClassDcl.sv
krame505: Yes, but it will only have to look at more than one element in the case when there actually *is* more than one in the environment - i.e. when there is an error.  Also we want to check if there is *more* than one in the env, not at least one (since the current declaration shows up in the env due to mutual recursion), so the pattern would need to be `_ :: _ :: _` - which seems considerably more awkward to be slightly more efficient in the error case.

ericvanwyk:
I agree the pattern is a bit clumsy.  But we need to compute the integer
value that is the length - so we must progress all the way to the end of
the list instead of stopping after finding the second one.  Or maybe I
don't understand what you mean by "will only have to look at more than one
element ...".   Either way, changing this may not be worth the performance
benefit but it is hard to say.

krame505:
`getTypeDclAll` returns the list of elements in the env (internally a tree-set) that match the name.  We are trying to check if this list contains more than one element - so we could only stop when we have already seen two elements (and thus already found an error.)  If there are no errors in the program, we will fully evaluate these lists (of one element each) in either approach.  

So there would only ever be a performance improvement using the pattern in programs that have more than two duplicate definitions of a name.

PR #402, opened 2020-12-29
Major refactor to support partial type application and higher-kinded polymorphism

### LINKED ISSUE #398: Higher-kinded polymorphism
I've been thinking a bit more about how we would do higher-kinded polymorphism in Silver, needed for type classes such as `Functor` and `Monoid`.  Despite what I said in previous discussions, I think that types in Silver do need to be curried after all for this to work sensibly.  However perhaps we don't need type constructor kinds such as `(* -> *) -> *` - so I think that a "kind-arity" should suffice here rather than having a `Kind` nonterminal.  

For example, consider the `Functor` type class:
```haskell
class Functor f<*> {
  map :: (f<b> ::= (b ::= a) f<a>);
}
```
I am open to suggestions for better kind signature syntax, btw. `f :: (* ::= *)` would sort of mirror Haskell, but `(* ::= * *)` seems weird if types are curried.  

`Functor` has an instance for `Either`, where the left argument is any type and the right is the one over which we can map:
```haskell
instance Functor Either<a> {
  map = \ fn::(c ::= b) e::Either<a b> ->
    case e of left(x) -> left(x) | right(x) -> right(fn(x)) end;
}
```
The reason we would want currying can be seen in how this instance would be resolved.  For example in the following code:
```haskell
local foo :: Either<String Integer> = ...;
local res :: Either<String Float> = map(toFloat, foo);
```
we would infer the type `(Either<String Float> ::= (Float ::= Integer) Either<String Integer>)` for `map`.  This type would need to be unified with `(f<b> ::= (b ::= a) f<a>)` prior to instance resolution - `f<b>` needs to unify with `Either<String Float>`, giving substitutions `f ~> Either<String>, b ~> Float`.  I don't know how this could be done sensibly without currying.  

The representation of `Type` in Silver will need to change a bit to make this work
* A new production `apType :: (Type ::= Type Type)` will represent application;
* `nonterminalType` would change from `(Type ::= String [Type])` to `(Type ::= String Integer)`, containing an arity instead of a list of children;
* `varType` and `skolemType` will also need an arity attached: `(Type ::= TyVar Integer)`

`apType` would be translated as the translation of the type being applied, and an unapplied type of kind-arity > 0 should never show up in a place where a translation is demanded.  Note that the argument type of `apType` should also always have kind-arity 0.

I'm open to suggestions if anyone sees any potential pitfalls here that I may have missed - @remexre @ericvanwyk  @tedinski?

krame505:
Fixes #398; see discussion in that issue.  With these changes we should now be able to add type classes for `Functor`, `Monad`, etc. (!)

This did require some significant changes to the list modification-pretending-to-be-an-extension that makes the situation there even slightly more awkward - we need to have `[]`, the unapplied list type constructor, itself as a type so that we can unify `[a]` with `f<a>` and write an instance for `Functor []`, etc.  This unfortunately necessitated changing the forward of `[a]` from `Decorated core:List<a>` to just `core:List<a>`, since there is no way to have `[]` be something, that when applied to `a` turns into  `Decorated core:List<a>` (and `Decorated core:List` would be a kind error.)  There is no real practical impact of this because the java translation is specialized anyway, but this is something to watch out for performance when considering alternative translations because `head`, `tail`, `length` and `null` would no longer be cached.


krame505:
@ericvanwyk please take a look when you get a chance, this is currently gating starting work on refactoring the standard library to use type classes (which would be nice to have done before the hackathon to avoid conflicts with changes being made by others.)

grammars/silver/analysis/typechecking/core/BuiltInFunctions.sv
```diff
@@ -48,7 +48,7 @@ Boolean ::= ty::Type
 {
   return case ty of
          | skolemType(_) -> true
-         | nonterminalType(_, args) -> any(map(containsSkolem, args))
+         | appType(c, a) -> containsSkolem(c) || containsSkolem(a)
```
ericvanwyk, line 51: Is there anything written up about the reasoning behind replacing `nonterminalType` with, I presume, `appType`?

ericvanwyk, line 51: Is there anything written up about `nonterminalType` being replaced by, I presume `appType`?  This seems significant, no?

grammars/silver/definition/concrete_syntax/ProductionDcl.sv
```diff
@@ -136,9 +136,9 @@ top::Type ::=
 }
 
 aspect production nonterminalType
-top::Type ::= fn::String params::[Type]
+top::Type ::= fn::String k::Integer
```
ericvanwyk, line 139: What is the reasoning here?  Do we not care about the `params` or did they move somewhere?

grammars/silver/definition/type/syntax/Terminals.sv
```diff
@@ -1,7 +1,9 @@
 grammar silver:definition:type:syntax;
 
+-- '<' has precedence 9, assoc = left
+
 terminal Boolean_tkwd    'Boolean'    lexer classes {TYPE,RESERVED};
-terminal Decorated_tkwd  'Decorated'  lexer classes {TYPE,RESERVED};
+terminal Decorated_tkwd  'Decorated'  lexer classes {TYPE,RESERVED}, precedence=1;
```
ericvanwyk, line 6: This seems unfortunate.

ericvanwyk (approved):
There are many parts of this I am unclear on, but that is just my losing touch with too much of the implementation these days.

grammars/silver/analysis/typechecking/core/BuiltInFunctions.sv
krame505: The discussion in #398, I guess?  

`nonterminalType` isn't exactly being replaced, type application is just being factored out - e.g.
```
nonterminalType("core:pair", [stringType(), integerType()])
```
would now become
```
appType(appType(nonterminalType("core:pair", 2), stringType()), integerType())
```
Hopefully this addresses some of your other comments?

ericvanwyk: That is what I ended up suspecting was the case.  The discussion in #398 helps, but isn't exactly a great resource for later.  But if that is what have it will have to do.  Thanks for this work - it does look very cool and I'm excited to have it in Silver.

grammars/silver/definition/type/syntax/Terminals.sv
krame505: Since type application is now `TypeExpr '<' TypeExprs '>'` rather than just `QName '<' TypeExprs '>'`, there is an ambiguity that didn't exist before - `Decorated Foo<Integer>` could now be parsed instead as `(Decorated Foo)<Integer>` rather than `Decorated (Foo<Integer>)` - which would always be a semantic error since `Decorated` anything has kind 0.  

`<` already has a precedence (since this is the same terminal as used in expressions), and adding a precedence here seemed simpler than a significant grammar refactoring.

PR #404, opened 2020-12-30
Removing obsolete monto code

krame505:


remexre (approved):

remexre:
I think there are some jars that can be nuked now too; gson and jeromq in particular

krame505:
I removed them from the list in `cron-job-publish-jars-from-foundry.py`, will that automatically be updated in the cron job on foundry when this is merged?

remexre:
No, because I wasn't confident that I wouldn't accidentally break something if I made the script update itself... The updated version needs to be moved to `/lhome/gitbot/cron-job-publish-jars-from-foundry.py` after the merge.

Issue #405, opened 2021-01-02
Silver concrete syntax construction syntax used in Silver extensions make post-refactor bootstrapping a nightmare


krame505:
This is basically what we outlined in the COLA reflection paper, but the practical implications are much more annoying than I previously thought.  

For example, @602p wanted to move the `end` keyword from `silver:modification:let_fix` to `silver:definition:core` to use this in host-language concrete syntax for origin tracking (where it probably belonged in the first place.  This is a reserved keyword and is used by several extensions and modifications, some of which are importing the let modification just to use this keyword.)  

However, the existing build of Silver uses `Silver_Expr { ... }` syntax literals in a number of places, many of which include `let` expressions and other syntax that uses the `end` keyword.  So if one attempts to rebuild Silver normally after making this change, the `silver:metatranslation` library reflects the abstract syntax trees of these literals and then attempts to translate them into Silver code according to the current compiler's abstract syntax - however the new abstract syntax that is being compiled doesn't match that of the current compiler (the names have changed), so the build fails during the initial pass.  

So the problem is that when bootstrapping this sort of refactor, we need to generate code for concrete syntax literals that matches the abstract syntax that we are currently compiling, rather than what we got from `reflect`.  In the case of moving the `end` keyword, the process that I ended up following was
1. Apply this patch:
```
--- a/grammars/silver/metatranslation/Translation.sv
+++ b/grammars/silver/metatranslation/Translation.sv
@@ -183,7 +183,7 @@ top::AST ::= terminalName::String lexeme::String location::Location
     terminalConstructor(
       'terminal', '(',
       nominalTypeExpr(
-        makeQNameType(terminalName, top.givenLocation),
+        makeQNameType(if terminalName == "silver:modification:let_fix:End_kwd" then "silver:definition:core:End_kwd" else terminalName, top.givenLocation),
         location=top.givenLocation),
       ',',
       stringConst(
```
2. Run `./self-compile --clean && cp build/silver.composed.Default.jar jars/`
3. Make the change moving `End_kwd` to `silver:definition:core`
4. Run `./self-compile` again to check that nothing broke
5. Revert the patch from 1
6. _Finally_ run `./deep-rebuild`
7. Commit and push the result, override the jars on Jenkins

We (I) will need to deal with this again, but on much worse scale when #152 finally gets done, moving compiler internals under `silver:` to `silver:compiler:`, since the full names of everything(!) will change at that point.  

[...]

PR #406, opened 2021-01-04
Major refactor to compiler and library grammar structure

krame505:
This (finally!) does the renaming as discussed in #152.  Restructuring the runtime and splitting things into seperate artifacts are projects left as future work.  

Note that some of the tests and downstream projects are currently failing, I will get to these soon but wanted to make this available for review.


ericvanwyk (approved):
I haven't looked terribly close but a quick glance looks good. Let's go with it for the Hackathon.  (I do wonder if `silver:compiler` should be `silver:lang` or `silver:langspec`.  But that is a minor issue.)

krame505:
Yeah, probably not `silver:lang` since that's extra confusing given that we were considering this as the grammar for the standard library.  

Also this is extremely annoying to change due to the bootstrapping issues I mentioned in #405, so I agree we should leave it for now.

ericvanwyk:
Agreed.

PR #411, opened 2021-01-05
Disambiguates pretty names for terminals

### LINKED ISSUE #335: Automatically-derived terminal names are ambiguous
The automatically-derived "human" terminal names implemented in #221 are awesome, except when Copper is reporting lexical ambiguities.  It is less than helpful to get an error message of the form
```
Unresolvable lexical ambiguity at parser states   [...] among terminals:
       ['table',edu:umn:cs:melt:ableC:concretesyntax:Identifier_t,'table']
```
as we don't know where the ambiguous 'table' keywords originated (what if 3 extensions defined a 'table' keyword?)  In this case it would be more helpful to have the full names of the terminals, as they would need to be included in the disambiguation function.  Not sure if this is something to fix in SIlver, or if we want to make Copper report both names when dealing with lexical ambiguities - e.g. something like
```
Unresolvable lexical ambiguity at parser states   [...] among terminals:
       ['table' (bogus_table:Table_t)
        edu:umn:cs:melt:ableC:concretesyntax:Identifier_t,
        'table' (edu:umn:cs:melt:exts:ableC:tables:concretesyntax:Table_t)]
```

remexre:
Resolves #335.

### DIFF
--- a/grammars/silver/compiler/definition/concrete_syntax/ast/Syntax.sv
+++ b/grammars/silver/compiler/definition/concrete_syntax/ast/Syntax.sv
@@ -1,6 +1,7 @@
 grammar silver:compiler:definition:concrete_syntax:ast;
 
 imports silver:compiler:translation:java:core only makeTerminalName;
+import silver:util:treemap as tm;
 
 -- For looking syntax elements up by name.
 monoid attribute cstDcls :: [Pair<String Decorated SyntaxDcl>] with [], ++;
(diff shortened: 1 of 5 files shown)


grammars/silver/compiler/definition/concrete_syntax/ast/Syntax.sv
```diff
@@ -41,13 +42,20 @@ autocopy attribute layoutTerms::EnvTree<String>;
 autocopy attribute prefixesForTerminals :: EnvTree<String>;
 autocopy attribute componentGrammarMarkingTerminals :: EnvTree<[String]>;
 
+-- Creating unambiguous <PP>s; this is a multiset used to accumulate all the
+-- names for terminals, and the actual name for <PP> will be modified to
+-- disambiguate if it would be ambiguous.
+monoid attribute prettyNamesAccum::[String] with [], ++;
+-- invariant: only have one value
+autocopy attribute prettyNames::tm:Map<String Integer>;
```
krame505, line 50: Took me a minute to understand this the Integer was the number of terminals with that name.  

Why not just pass down the original map?  Or even just make this a map of pretty name -> terminal name and collect up `Pair<String String>` instead of collecting just the pretty names and zipping with `Unit`?

remexre: trying to reduce memory footprint; I suppose `unit()` allocates, though... Iunno, I guess I want (in Haskell) `foldl' (Data.Map.Strict.alter ((1+) . fromMaybe 0)) Data.Map.Strict.empty prettyNamesAccum` for minimal allocation, but...

I'll do the `Pair<String String>` thing.

krame505 (approved):
Looks good, just hold off on merging for a bit until Jenkins is successfully rebuilding develop.

PR #413, opened 2021-01-05
Migrate silver:compiler:definition:core:regex into a seperate library, add Silver regex matching extension

### LINKED ISSUE #407: Move silver:compiler:definition:regex out of the compiler
This is a generally-useful library that should exist outside of the compiler as just `silver:regex`, probably (for consistency, we have `silver:json` and `silver:xml` already.)  

Then we should actually use this library - it has been copied out of the compiler to a few places, like in the ableC-regex extension.

krame505:
This fixes #407 by factoring out regexps into a separate library with abstract syntax, and also introduces a regex matching extension to silver, based on the example of Brzozowski derivatives from the strategy attribute paper.  

This means we no longer are putting regex concrete syntax in interface files, and also fixes a long-standing bug with single-quoted "easy terminal" resolution.


ericvanwyk (approved):
Looks good.

PR #414, opened 2021-01-05
Add --mwda flag to options

### LINKED ISSUE #396: Add a --mwda flag
Currently we have the flags `--warn-all` that enables all warnings related to the MWDA, and `--warn-error` that turns all warnings into errors.  We use these together in the ableC extensions when running the MWDA.  

However, this effectively prevents us from adding new kinds of warnings that aren't MWDA ones, since `--warn-error` causes these warnings to be interpreted as fatal errors (see e.g. the warning about production names being lower case in the silver regex grammar causing the `mwda` step to fail for an ableC extension that imported this grammar.)  Additionally we may also wish to include other analyses in `--warn-all` that aren't MWDA-related (e.g. #278).  

A nice fix for this would be to add `--mwda`, that enables all MWDA checks and treats only *those* failures as errors, while allowing all other warnings to remain warnings.  

The next step after implementing this is to actually use it in all the ableC extension Makefiles.

counc009:
Resolves #396 
My approach is to introduce a new --mwda flag into the command-line options and wire the resulting boolean (whether the flag is given or not) into everywhere the `wrn` for the MWDA analysis was produced. The flag is then used to enable the production of these messages (the same as if the `--warn-all` flag were used) and the value is also used to determine whether to produce a `wrn` or an `err`. The reason for this is that it causes `--warn-all` to still produce all MWDA warnings as just warnings (which can then be made errors using `--warn-error`).

Once this is merged we'll need to change the Makefiles in all the ableC extensions to use `--mwda` rather than `--warn-all --warn-error` to run MWDA analysis. Once that's done I am aware of one place in Silver (line 130 of silver/compiler/definition/core/BuiltInFunctions.sv) where we have a deprecation warning that is commented out because of its interference with the MWDA analysis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,3 +67,4 @@ Silver and Copper are distributed under the GNU Lesser General Public
 License.  See the files COPYING and COPYING.LESSER for details of
 these licenses.  More information can be found at
 http://www.gnu.org/licenses/.
+
--- a/grammars/silver/compiler/analysis/warnings/flow/FlowTypeCopyEquation.sv
+++ b/grammars/silver/compiler/analysis/warnings/flow/FlowTypeCopyEquation.sv
@@ -26,16 +26,16 @@ top::AGDcl ::= 'abstract' 'production' id::Name ns::ProductionSignature body::Pr
 
   top.errors <-
     if null(body.errors ++ ns.errors)
-    && (top.config.warnAll || top.config.warnMissingInh)
+    && (top.config.warnAll || top.config.warnMissingInh || top.config.runMwda)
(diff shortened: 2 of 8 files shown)


krame505 (approved):

ericvanwyk:
Great! Thanks @counc009 - the next step is to figure out where this gets documented.  Can you follow up on that?

counc009:
Yeah, of course. The two places on the website I see change needed is in the command-line reference page on the MWDA page. I'll make those changes and make a PR on the website. Let me know if anyone knows of other places to update.

PR #415, opened 2021-01-06
Fixes newlines in unparse.

hypnoticOxbow:
Before they were just prepended with line comments.
Without this, some elements of "raw" silver would be fed to javac on occasion.
An example of the issue I was seeing:
```
public static silver.compiler.definition.core.NAGDcl invoke(final common.OriginContext originCtx , final Object c_ty, final Object c_attr, final Object c_mRule) {
		try {
			final common.DecoratedNode context = new PgenerateAspectExprFromMatch(c_ty, c_attr, c_mRule).decorate(originCtx);
			//case mRule.headPattern of wildcPattern(_) -> Silver_AGDcl {aspect default production
top::$TypeExpr{ty} ::=

```
That `top::$TypeExpr{ty} ::=` bit would get treated as java code, and error out

### DIFF
--- a/grammars/silver/compiler/extension/silverconstruction/Syntax.sv
+++ b/grammars/silver/compiler/extension/silverconstruction/Syntax.sv
@@ -11,14 +11,14 @@ imports silver:compiler:extension:patternmatching;
 concrete production quoteAGDcl
 top::Expr ::= 'Silver_AGDcl' '{' ast::AGDcl '}'
 {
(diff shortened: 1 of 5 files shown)


hypnoticOxbow:
It turns out that because java comments can't nest, the approach of surrounding unparse with block comments in java doesn't help us here. Instead, I stripped out new lines from the relevant unparse attributes that were causing me trouble.

krame505 (commented):
For readability, maybe replace newlines with spaces instead of just removing them? But otherwise looks good.

krame505 (approved):

PR #416, opened 2021-01-06
Improve type errors on functions with incorrect number of arguments

### LINKED ISSUE #33: Function type errors in arguments are wrong and confusing

counc009:
Resolve #33 (I think, still can't replicate the "too many arguments" errors when the function is not found).

My approach to matching types to arguments from left-to-right, rather than right-to-left, is to construct a list of types that matches the number of provided arguments; if too many arguments are provided we add `errorType()` to the front (which because the argument types are reversed places the `errorType()` on the arguments at the end which makes sense since we really can't type-check them) and if there are too few arguments we drop elements from the front of the list of types (which correspond to arguments that should be given at the end). This also requires that we handle the checking for whether the number of expected arguments matches the number of provided arguments in `snocAppExprs` since we will provide the `AppExprs` child a list of types that is equal to the number of arguments it has listed, so the checking in `oneAppExprs` will no longer detect that there are too few or too many children.

### DIFF
--- a/grammars/silver/compiler/definition/core/Expr.sv
+++ b/grammars/silver/compiler/definition/core/Expr.sv
@@ -897,12 +897,23 @@ top::AppExprs ::= es::AppExprs ',' e::AppExpr
 
   top.appExprSize = es.appExprSize + 1;
 
-  e.appExprIndex = es.appExprSize;
-  e.appExprTyperep = if null(top.appExprTypereps)
-                     then errorType()
-                     else head(top.appExprTypereps);
+  top.errors <- if length(top.appExprTypereps) > top.appExprSize
+                then [err(top.location, "Too few arguments provided to function '" ++ top.appExprApplied ++ "'")]
+                else if length(top.appExprTypereps) < top.appExprSize
+                then [err(top.location, "Too many arguments provided to function '" ++ top.appExprApplied ++ "'")]
+                else [];
 
-  es.appExprTypereps = if null(top.appExprTypereps) then [] else tail(top.appExprTypereps);
+  local correctNumTypes :: [Type] =
+    if length(top.appExprTypereps) > top.appExprSize
+    then drop(length(top.appExprTypereps) - top.appExprSize, top.appExprTypereps)
+    else if length(top.appExprTypereps) < top.appExprSize
+    then repeat(errorType(), top.appExprSize - length(top.appExprTypereps)) ++ top.appExprTypereps
+    else top.appExprTypereps;
+
+  e.appExprIndex = es.appExprSize;
+  e.appExprTyperep = head(correctNumTypes);
+  
+  es.appExprTypereps = tail(correctNumTypes);
 }
 concrete production oneAppExprs
 top::AppExprs ::= e::AppExpr
--- a/grammars/silver/compiler/definition/core/Expr.sv
+++ b/grammars/silver/compiler/definition/core/Expr.sv
@@ -183,10 +183,17 @@ top::Expr ::= e::Expr '(' es::AppExprs ',' anns::AnnoAppExprs ')'
   -- TODO: fix comma when one or the other is empty
   top.unparse = e.unparse ++ "(" ++ es.unparse ++ "," ++ anns.unparse ++ ")";
   
+  local correctNumTypes :: [Type] =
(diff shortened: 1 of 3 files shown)


grammars/silver/compiler/definition/core/Expr.sv
```diff
@@ -897,12 +897,23 @@ top::AppExprs ::= es::AppExprs ',' e::AppExpr
 
   top.appExprSize = es.appExprSize + 1;
 
-  e.appExprIndex = es.appExprSize;
```
krame505, line 900: It feels a little weird to be doing this check (and the padding) on `snocAppExpr`.  Could this be moved (along with the check on `oneAppExprs` to the `application` production?

krame505 (changes requested):

grammars/silver/compiler/definition/core/Expr.sv
```diff
@@ -183,18 +183,38 @@ top::Expr ::= e::Expr '(' es::AppExprs ',' anns::AnnoAppExprs ')'
...
   anns.funcAnnotations = anns.remainingFuncAnnotations;
   
   -- TODO: You know, since the rule is we can't access .typerep without "first" supplying
   -- .downSubst, perhaps we should just... report .typerep after substitution in the first place!
-  forwards to t.applicationDispatcher(e, es, anns, top.location);
+  local applied :: Expr = t.applicationDispatcher(e, es, anns, top.location);
+  
+  top.errors <- 
+    case applied of
+    | errorApplication(_, _, _) -> []
```
counc009, line 208: This feels a little hacky to me, but avoids errors about "too many arguments" when what we are trying to apply can't actually be applied.

grammars/silver/compiler/definition/core/Expr.sv
counc009: Yeah, my new commit rearranges this so that the check and padding are in `application` instead.

grammars/silver/compiler/definition/core/Expr.sv
krame505: Can't you just check `if t.isError then [] else ...`?  

Matching on an `Expr` like this is actually a flow error since you aren't supplying the forward flow dependencies.

krame505 (changes requested):

grammars/silver/compiler/definition/core/Expr.sv
counc009: That won't work because `t.isError` just checks whether the type is an `errorType`, but we also need to handle the type being a valid type but just not being a type we can apply:
```
local x :: Integer = 3;
local t :: String = x(4);
```
produces a "Too many arguments provided to function 'x'" error if we just do `if t.isError ...`

We could potentially add a new attribute `isAppliable` or some such.

krame505: Yeah `isAppliable` or even just `isFunction` would work.

grammars/silver/compiler/definition/core/Type.sv
```diff
@@ -119,5 +124,6 @@ top::Type ::= out::Type params::[Type] namedParams::[NamedArgType]
 {
   -- TODO: We don't seem to use this. Perhaps we should remove it?
   top.applicationDispatcher = functionApplication(_, _, _, location=_);
```
krame505, line 126: TODO can be removed :-)

grammars/silver/compiler/definition/core/Type.sv
```diff
@@ -15,13 +15,17 @@ synthesized attribute instanceOrd :: Boolean;
 synthesized attribute instanceNum :: Boolean;
 synthesized attribute instanceConvertible :: Boolean;
 
+synthesized attribute isApplicable :: Boolean;
```
krame505, line 18: I would maybe put this in `silver:definition:type:Util.sv` instead for consistency (`instanceEq`, `instanceOrd`, etc. will be going away soon), but otherwise looks good.

krame505 (approved):

PR #419, opened 2021-01-07
Add prefix to names of globals in Java translation to avoid keywords

### LINKED ISSUE #316: Naming a global 'try' causes invalid code to be generated
'try' is a keyword in Java, and naming a global this means we attempt to generate a variable with this name. There may be other such instances lurking for other Java keywords.  
We could fix this by just adding some sort of prefix to the names of generated variables for globals.  @tedinski do you know of any unforeseen difficulties that would be encountered in making such a change?  Alternatively we could just make sure all java keywords are reserved in Silver.

counc009:
Resolves #316 
As suggested in the issue, I fixed this just by prepending "global_" to the name of any global when we emit it or use it, which should avoid the names conflicting with Java keywords.


krame505 (approved):

krame505:
Can we hold off on merging this until #420 is merged?  My work on type classes is sort of gated on having updated jars from that PR be published, and Jenkins is pretty backlogged right now.

counc009:
Yeah, fine with me

PR #420, opened 2021-01-07
Bug fixes and added tests for type classes

krame505:


grammars/silver/compiler/definition/core/ClassDcl.sv
```diff
@@ -115,6 +117,18 @@ top::ClassBodyItem ::= id::Name '::' ty::TypeExpr '=' e::Expr ';'
     setUnionTyVars(top.classHead.freeVariables, ty.typerep.freeVariables);
   top.classMembers = [pair(fName, pair(ty.typerep, true))];
   
+  e.isRoot = true;
+  e.originRules = [];
+
+  -- oh no again!
```
ericvanwyk, line 123: Will this comment be useful to anyone other than you?

ericvanwyk (approved):

grammars/silver/compiler/definition/core/ClassDcl.sv
krame505: This comment and chunk of code is something that Ted added that shows up in a whole bunch of places.  I copied the comment so these are easy to find in case we want to clean this up eventually, somehow.

ericvanwyk: Fair enough - just felt I had to comment on something...

PR #421, opened 2021-01-07
Add runtime check for base definition of production collection attribute

### LINKED ISSUE #322: Failing to initialize a collection production attribute results in a NullPointerException
See title.  I ran into this using a collection production attribute of type `Maybe`, using `orElse` as the operator.  I don't have time to investigate this in more detail right now, but wanted to make a note of this for the future.  
Not a big deal since this is caught by the flow analysis, just kind of annoying.

counc009:
Resolves #322 
My solution is to check the result from `getBase()` in the Java translation and if the result is null report out own exception rather than continuing which produces the NPE the issue notes. Open to comments on what exactly the exception and message should be.


krame505 (approved):
Looks good, although this is apparently still using the old pre-string-template approach to code generation?  Generally we are trying to avoid introducing new code like this since it is hard to read and maintain - so you could change this to use the `s""" """` syntax, if you want.

counc009:
Yeah, I'll do that. I think I'll replace all the old style code generation in that file with `s""" """`.

Issue #399, opened 2020-12-22
Tuple syntax


krame505:
Eric pointed out that writing `Pair<String Integer>` in Silver is significantly more annoying than `(String, Integer)` in Haskell.  And more than 2 elements are even worse: `Pair<String Pair<Boolean Integer>>` vs. `(String, Boolean, Integer)`.  

We should add some nicer tuple syntax in Silver.  The easiest way to do this is probably inductively, such that `(a, b, c, d)` just forwards to `Pair<a Pair<b Pair<c d>>>`.  We would also want to add similar forwarding pattern and expression syntax: `(w, x, y, z)` forwards to `pair(w, pair(x, pair(y, z)))`.  

One could extract the members of a tuple via pattern matching or alternatively through the attributes on `Pair` (e,g, `.snd.snd.fst`).  Do we want to add more convenient syntax to access the nth member of a tuple?  Haskell doesn't have such syntax, but it is also possible there to immediately pattern match on a function argument or `let`-binding.  

Note that tuples in Haskell are *not* inductive, however the Bluespec language that I have been working on this year does do inductive tuples in this manner.  There is another side-benefit to this approach: one can write a type class instance for `Pair` and then automatically have instances for all tuple types.

RandomActsOfGrammar:
I'm in favor of doing better syntax for pairs at least.  I think tuple syntax would also be nice to have.

I think if we're going to add tuple syntax, it would be nice to have a way to access elements without needing to write the pattern matching or a chain of `fst` and `snd`.  Off the top of my head, I have a couple of suggestions.

We could have a `get` attribute which takes an integer for the index to get.  The attribute access `p.fst` would be equivalent to `p.get(1)`.

We could have syntax which lets us write the index as if it were an attribute access.  For example, `p.fst` would be equivalent to `p.1`.  The index in general could be translated to `index - 1` `.snd`s followed by a `.fst`.

ericvanwyk:
I'm not sure about doing this 'inductively'.  The types `(int * int) * int`, `int * (int * int)` and `int * int * int` are all different types.  I agree doing it this way is the easy way to have no limit on the size of tuples, but it does not seem correct to me.  I would suggest following Haskell here.

RandomActsOfGrammar:
I believe Idris treats tuples inductively, so `(int, int, int)` is equivalent to `(int, (int, int))` and vice versa.  Therefore this behavior isn't *completely* unprecedented, but it could lead to some unintuitive behavior.

[...]

PR #423, opened 2021-01-08
Feature/tuple syntax

alexandra-hanson:
I have added syntax for tuples as an extension, and done so inductively. We forward a tuple `(x, y, z)` to `pair(x, pair(y, z))`, as described in #399.

I would like feedback on if I have provided sufficient test cases, and if tuple behavior is as desired.


grammars/silver/compiler/extension/tuple/Tuple.sv
```diff
@@ -0,0 +1,73 @@
...
+
+nonterminal TupleList with location, unparse, translation;
+synthesized attribute translation :: Expr;
+
+-- Note: We consider only tuples containing two or more elements
+
+concrete production tuple_c
+top::Expr ::= '(' tl::TupleList ')'
+{
+    top.unparse = "(" ++ tl.unparse ++ ")";
```
krame505, line 19: Minor style note: we typically indent by 2 spaces throughout our Silver code base.

krame505 (commented):
What you have so far looks great - I will approve this once you have the type expression syntax finished.  

Also for completeness you could/probably should add `()` as a type that forwards to `Unit` and as a pattern/expression that forwards to `unit()`.

grammars/silver/compiler/extension/tuple/Tuple.sv
```diff
@@ -0,0 +1,90 @@
...
+
+concrete production tupleTypeExprn
+top::ListOfTypeExprs ::= te::TypeExpr ',' tes::ListOfTypeExprs
+{
+  top.unparse = te.unparse ++ "," ++ tes.unparse;
+  top.te_translation = Silver_TypeExpr {silver:core:Pair<$TypeExpr{te} $TypeExpr{tes.te_translation}>};
+}
+
+
+concrete production tuple_c
```
krame505, line 40: I would maybe just call this `tupleExpr`.

krame505:
This is pretty good as-is, however currently tuple types would still show up in error messages as the underlying chain of `Pair` types.  This could be fixed by adding a `tupleType :: (Type ::= [Type])` production that forwards to the expected chain of `Pair` types.  `tupleType` would override its `unparse` to use the parenthesized representation, while `tupleTypeExpr`  and `tupleExpr` would override their types to be `tupleType` applied to the appropriate arguments.

ericvanwyk:
Yes, nicer type error messages would be nice to have.

krame505 (approved):

krame505:
You might also consider trying to use this in the Silver compiler, e.g. https://github.com/melt-umn/silver/blob/develop/grammars/silver/compiler/metatranslation/Translation.sv#L83 and a few other places in the that file involve pathological chains of `Pair` types.

grammars/silver/compiler/metatranslation/Translation.sv
```diff
@@ -74,21 +75,21 @@ top::AST ::= prodName::String children::ASTs annotations::NamedASTs
...
-            consAST(rest, nilAST())) -> just(pair(n, pair(a, rest)))
+            consAST(rest, nilAST())) -> just((n, a, rest))
         | _ -> nothing()
         end;
-      -- pair(nonterminal short name, pair(cons production name, append production name))
-      trans::Pair<String Pair<String String>> <-
+      -- (nonterminal short name, cons production name, append production name)
+      trans::(String, String, String) <-
         lookup(antiquote.fst, collectionAntiquoteProductions);
       if prodName == trans.snd.fst then just(unit()) else nothing(); -- require prodName == trans.snd.fst
```
alexandra-hanson, line 94: The second element of the tuple is still accessed by trans.snd.fst. Is that confusing/is there a better way for this access to occur?

RandomActsOfGrammar: As I recall (and it looks like it says this at the bottom of the tuple issue as well), we agreed it would be a good idea to use `.<int>` to access elements of a tuple.  For your example here, that would be `trans.2`.

alexandra-hanson: My bad! I will work on that now.

krame505:
@alexandra-hanson is this ready to be merged?  Or did you want to finish up the new selector syntax in this branch first, before merging this?

ericvanwyk:
I think it might be good to wait.  @alexandra-hanson and I chatted today and I had some questions about how things were done - mostly down to my ignorance on things.  We may corner you, @krame505, for a conversation next week sometime.

alexandra-hanson:
I have added a `/tuple/README.md` to attempt to explain how tuples work. Probably the documentation is a bit excessive, but I needed to write it down to explain how it worked to myself. There also could be things I am missing. I have added notes that are questions and thoughts.

Probably the most pertinent section is about types. This contains the most thoughts/questions also.

grammars/silver/compiler/extension/tuple/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+--
+
+Allie has a few thoughts:
+1. If we use `tupleElems` the way that it is currently implemented, won't there be a `tupleElems` assigned for every single application of a constructor type, even if it is not a pair? This does not seem ideal...
+2. I am not sure if I entirely understand the default production. For tuple purposes, I think it is used to assign a value to `tupleElems` in the case when we have the last element type we need to append?
```
RandomActsOfGrammar, line 33: The default production is shorthand for writing the same equation on every production other than the ones we really care about.  If we have `top.attr = 0;` in the default production, this is the equation in all the productions in the grammar where we *don't* write an equation.

By putting `top.tupleElems = [top];` in the default production, you're saying that the list of types in the "tuple" represented by that type is simply the type itself.

grammars/silver/compiler/extension/tuple/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+Since `tupleElems` occurs on `Type`, we can use the `appType` production to append the individual types (like `Integer`) that may be in the the nested pairs into the `tupleElems` list. It is notable that the `appType` production represents the application of a constructor type (like `Pair`), with constructor `c` and argument `a`. In our new production, we pattern match on the `baseType`of `c` and assign the appropriate value to the `tupleElems` of the top-level `Type`. If the `baseType` of the constructor `c` is a `Pair`, then we know its `argTypes` should have only a single element, the `fst` element in the pair. We then append the `tupleElems` of the argument `a` to the list `argTypes` of `c`, which will either recurse on `a` if it is also the application of a constructor type. If it is not, its `tupleElems` will simply be assigned to the singleton list of the top-level as in the `default production`. 
+
+--
+
+Allie has a few thoughts:
+1. If we use `tupleElems` the way that it is currently implemented, won't there be a `tupleElems` assigned for every single application of a constructor type, even if it is not a pair? This does not seem ideal...
```
RandomActsOfGrammar, line 32: There will be a value of `tupleElems` for each tree of type `Type`.  If you look at the equation you have on `appType` for `tupleElems`, you will see that you only do something interesting in the case that it is a pair; otherwise, you do the same as every other type, where you have `top.tupleElems = [top];`.

krame505: Not sure if this is what you are concerned about, but remember that attributes are computed lazily - so this attribute on `Type` won't add performance overhead or anything where it isn't explicitly demanded by the tuple extension.

grammars/silver/compiler/extension/tuple/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+--
+
+In order to appropriately handle the `typerep` of a tuple, as in productions `tupleTypeExpr` and `tupleExpr` (*Should this also be in `tuplePattern`?*), we assign its value to the direct call of production `tupleType` on the `tupleElems` of the type representation of the `forward`. Please note that we must use `forward` rather than `translation` or `te_translation` in this call because we want to send in a decorated type (*Is this correct?*). Within the `tupleType` production, we properly assign the `typepp` or "pretty print" of the type to be the implosion of the list of types `ts` (that we previously obtained from `tupleElems`). We then forward to the pattern match on `ts`: (1) if we have an empty list of types, we forward to `Unit`; (2) if we have a singleton list, we forward simply to that single type; (3) finally, if we have a proper list of types, we then use the `appType` production to construct the `Pair` type with the head of the list and the recursive call to `tupleType` on the tail of the list.
+
+(*Allie Question: Why do we use/assign attributes `substituted` and `flatrenamed` in `tupleType`?*)
```
RandomActsOfGrammar, line 39: This lets substitutions be done without destroying the original structure.  If `substituted` were done by forwarding, the value of `substituted` would have the structure of the pair types to which this forwards rather than the tuple type it originally was.

RandomActsOfGrammar:
I like the idea of having a `README.md` for each extension to provide a guide for future development on it.

Your current version is good while you're still figuring things out, but a lot of it is just translating the code into English.  The final version can be shorter and include less detail.  For example, once this is merged, people can see that `Silver_Expr` is used to create the forwarded-to pairs; it isn't an unusual enough feature to be pointed out in documentation.

grammars/silver/compiler/extension/tuple/README.md
alexandra-hanson: Ah okay!! Yes, that makes a lot of sense. Thank you both.

grammars/silver/compiler/metatranslation/Translation.sv
```diff
@@ -75,8 +74,8 @@ top::AST ::= prodName::String children::ASTs annotations::NamedASTs
   
   -- "Collection" antiquote productions
   -- Key: antiquote production name
-  -- Value: (nonterminal short name, cons production name, append production name)
-  production attribute collectionAntiquoteProductions::[(String, String, String, String)] with ++;
+  -- Value: pair(nonterminal short name, pair(cons production name, append production name))
```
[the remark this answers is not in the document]
alexandra-hanson: Yeah, with what I haven't pushed yet -- trying to add the selector syntax -- it would show up with errors on translation during compilation and not notify me anything about what was wrong with my tuple syntax. I also had to merge in the changes from develop and there were conflicts so I thought it would just be easier to revert it. I will change it back when it is working. Sorry if this is not ideal.

ericvanwyk:
I agree - these README.md files are helpful.  Especially in an initial form that tracks the thoughts and questions that arise.  When Allie and I last spoke we thought something like this would be good to sort things out and get some feedback.  Thanks all.

alexandra-hanson:
@krame505 and @RandomActsOfGrammar Thank you both! Your comments were really helpful. I think that has cleared up where I was confused with the way types are implemented. It was helpful for me to write up the code into English to get my thoughts in order and figure out where my questions were but I will certainly shorten the README.md once I'm done.

grammars/silver/compiler/extension/implicit_monads/Case.sv
```diff
@@ -158,20 +158,20 @@ elst::[Expr] tylst::[Type] env::Decorated Env sub::Substitution f::BlockContext
                                             expectedMonad=em;}.mtyperep
            in
              if fst(monadsMatch(ety, em, sub))
-             then pair(pair(ety, pair(e, newName)) :: subcall.fst,
-                       baseExpr(qName(loc, newName), location=loc) :: subcall.snd)
-             else pair(subcall.fst, e::subcall.snd)
+             then ((ety, (e, newName)) :: subcall.fst, 
+                   baseExpr(qName(loc, newName), location=loc) :: subcall.snd)
+             else (subcall.fst, subcall.snd)
```
alexandra-hanson, line 163: I wanted to change these `.fst` and `.snd` accesses to tuple accesses, such that we say `subcall.1` and `subcall.2`; however, when I attempted to do so, I received the error:

`Case.sv:161:19: error: Argument 2 of function 'silver:core:cons' expected [(silver:compiler:definition:type:Type, silver:compiler:definition:core:Expr, String)] but argument is of type Decorated ([(silver:compiler:definition:type:Type, silver:compiler:definition:core:Expr, String)], silver:core:List<silver:compiler:definition:core:Expr>)
Case.sv:162:66: error: Invalid tuple selector index.`

I added a test just to make sure that the tuple selector syntax worked on 2 element tuples. I don't understand exactly what this error means/how I should approach fixing this.

grammars/silver/compiler/extension/implicit_monads/Case.sv
```diff
@@ -135,12 +135,12 @@ Boolean ::= elst::[Expr] env::Decorated Env sub::Substitution f::BlockContext gn
 --   well as a new list of expressions for the forward to use
 --use a name from names when that is not empty; when empty, use a new name
 function monadicMatchTypesNames
-Pair<[Pair<Type Pair<Expr String>>] [Expr]> ::=
+([(Type, (Expr, String))], [Expr]) ::=
```
krame505, line 138: This could just be `(Type, Expr, String)`

grammars/silver/compiler/extension/implicit_monads/Case.sv
krame505: Looks like on line 163 you have `subcall.snd` instead of `e :: subcall.snd`?

Also you probably need to add an equation for `tupleElems` on `listType`/listCtrType`, since the error message is showing `silver:core:List<...>` instead of `[...]`

grammars/silver/compiler/extension/tuple/README.md
```diff
@@ -0,0 +1,54 @@
+# Tuple Syntax
+
+Tuples use a comma "`,`" as an infix operator to seperate individual tuple elements. They are implemented inductively using the construction of ordered pairs, e.g. we forward a tuple `(a, b, c, d)` to `Pair<a Pair<b Pair<c d>>>`. This means the following examples are equivalent:
+
+    ("I", ("am", "a", (5, "tuple"))) = ("I", "am", "a", 5, "tuple")
```
alexandra-hanson, line 5: Not sure if this information should go in the README or on the website. I want to give a brief overview of how tuples work but I'm not sure if this is brief enough here. In this section I could just point to the website to explain usage?

grammars/silver/compiler/extension/tuple/Tuple.sv
```diff
@@ -0,0 +1,96 @@
...
+}
+
+concrete production tupleExpr
+top::Expr ::= '(' tl::TupleList ')'
+{
+  top.unparse = "(" ++ tl.unparse ++ ")";
+
+  -- substitution is cleaner
+  -- instantiates types on pairs & gets tuples off pairs
+  -- it avoids threading the env?
```
alexandra-hanson, line 33: Is this correct & do I need more information to explain this than this comment?

grammars/silver/compiler/extension/tuple/Tuple.sv
```diff
@@ -0,0 +1,96 @@
...
+  -- Forward gets the substitution context of the tuple
+  propagate downSubst, upSubst, freeVars;
+
+  local accessIndex::Integer = toInteger(a.lexeme);
+
+  top.unparse = tuple.unparse ++ "." ++ a.lexeme;
+
+  -- If tuple is decorated, the length of its tupleElems 
+  -- will be 1, so it must be undecorated? before we can take
+  -- the length of its tupleElems
```
alexandra-hanson, line 53: Is it correct to say that we are "undecorating" t. I'm not sure if this is the correct terminology.

grammars/silver/compiler/extension/tuple/Type.sv
```diff
@@ -0,0 +1,101 @@
...
+
+-- accepts a [Type] (will be tupleElems here)
+-- prints tuples w/ correct type syntax
+-- forwards to Pair type
+abstract production tupleType
+top::Type ::= ts::[Type]
+{
+
+  -- why?
+  -- we decorate the types in the tuple?
```
alexandra-hanson, line 56: Why do we do this with `top.substituted` and `top.flatRenamed` again?

grammars/silver/compiler/extension/tuple/README.md
krame505: I'm not sure how I feel about having README.md files scattered in the code base - this would be the only non-`.sv` file under `grammars/`, as far as I know.  The info before the "Implementation" section here (don't to mention patterns too!) could fit on the website under http://melt.cs.umn.edu/silver/ref/ext/.

The rest of the implementation description can probably move into the source code as doc comments in appropriate places - so this would (eventually) end up on the website as well if we decide to put up generated docs for the compiler internals.  

The way that this is implemented doesn't really have `,` as an infix operator, by the way - if it were so you could write just `[String, Integer]` without the parentheses.  But it can be helpful to think about that way, I guess.

grammars/silver/compiler/extension/tuple/Tuple.sv
krame505: Umm, that's not really super clear.  The key thing is that computing the specialized `tupleType` from `forward.typerep` is cleaner than performing type checking on the `TupleList`.  `performSubstitution` is needed because `forward.typerep` isn't instantiated into a chain of pairs until `upSubst` is applied.

grammars/silver/compiler/extension/tuple/Tuple.sv
krame505: Not exactly, typically by that we mean turning a decorated tree into an undecorated term.  Here we just need to get at the underlying `tupleType` from a type that might be inferred to be decorated by default.

grammars/silver/compiler/extension/tuple/Type.sv
krame505: We want to avoid transforming away the `tupleType` and turning it back into a chain of `Pair`s when performing substitutions, like we discussed today with lists.

grammars/silver/compiler/extension/tuple/README.md
RandomActsOfGrammar: I think the benefit of having `README.md` files in extension directories is to provide documentation on that extension for Silver developers, so if anyone needs to make changes to the extension they have a quick introduction to it.

With that interpretation, the part before `Implementation` doesn't really belong here, since that is more for users of the extension than for developers of it.

krame505: > I think the benefit of having README.md files in extension directories is to provide documentation on that extension for Silver developers, so if anyone needs to make changes to the extension they have a quick introduction to it.

Yes, but that is what doc comments are for.  There are a number of comments in the Silver compiler source that should become generated doc comments once that is working.  My main issue with using `README.md` files is that these won't be included in the generated documentation.

alexandra-hanson: I can translate the `README.md` into doc comments but I'm not entirely sure if there is a consensus on this so I'm not sure how I should proceed.

krame505: If @ericvanwyk thinks that we should have README files in extension grammars then I guess this is fine.  If these were doc comments instead they would still be in the source tree and also get added to the website.  But that can always be fixed later if/when we start putting generated docs from the `silver:compiler` grammars on our website.

ericvanwyk: Every repo should have a README since this might be some reader's first entrance into that material.  It might be that the best thing to have in the README are not much more than pointers to the Silver web page for how to use, in this case, the extension, and to the generated doc-comments that describe the implementation of the specifications in the repo.  In the case of this one I think it makes sense to say what the purpose of the extension is, a little bit about what it provides and then pointers to the relevant web pages that have the documentation. We do want to avoid multiple copies of large bits of documentation - but a paragraph or two that lives here in the README and also somewhere doesn't seem unreasonable. At least until we have the Silver documentation sorted out. What is here now might be more than we need - but again - we need to have the other documentation sorted out before we judge what is here as being too much or too little.

PR #424, opened 2021-01-08
Build all the tests on Jenkins, fix origins tests broken by #406

krame505:
This uses the silver test driver to build all of the silver tests, not just the ones under the top-level `test/` directory.  Also fixes failing origins test cases broken by #406 that were missed due to not being built on Jenkins.


remexre (approved):
I don't actually grok these changes, but the commit log is kinda funny in a tragic way.

PR #426, opened 2021-01-10
Adds AspectRHS antiquote production.

hypnoticOxbow:


krame505 (approved):

grammars/silver/compiler/extension/silverconstruction/Syntax.sv
```diff
@@ -81,6 +81,14 @@ top::Pattern ::= '$Pattern' '{' e::Expr '}'
       location=top.location);
 }
 
+
+concrete production antiquoteAspectRHS
+top::AspectRHS ::= '$AspectRHS' '{' e::Expr '}'
+{
+  top.unparse = s"$$AspectRHS{${e.unparse}}";
+  forwards to aspectRHSElemNil(location=top.location);
```
remexre, line 89: Uh, wait shouldn't this involve `e` in some way?

hypnoticOxbow: This is for when the production is found outside of the $Silver_* bits afaict. Look at the other productions like Expr, where it just forwards to error

remexre: oh, hm, I guess there's more magic in `quote*` than I realized...

ericvanwyk:
Who is going to add documentation for this to the website? We need to at least not add anymore Silver features that are not documented.

hypnoticOxbow:
I could add documentation for this antiquote specifically, but I don't really understand all the antiquotes available in silverconstruction. If there's a place I could add it on the website I can do that.

ericvanwyk:
Oops, I'm on the wrong issue.  It is less important to document this. I meant the aspect production feature itself.

krame505:
These features are really only of interest to Silver compiler developers.  Does that belong on the website or should it go somewhere else?

krame505:
Ah yes, agreed.

Issue #401, opened 2020-12-29
Using type classes


krame505:
Meta-bug to keep track of all the things that we want to change in the standard library once type classes are finished.  Some of these are partially doable now and some depend on attribute occurrence constraints.
- [ ] `Eq` typeclass
  - [ ] Non-`by` functions (`contains`, `lookup`, `union`, etc.)
  - [ ] Make `==` forward to calling `eq`
  - [ ] Default instance with `isEqual/isEqualTo` equality attributes
- [ ] `Ord` typeclass
  - [ ] Non-`by` functions (`sort`, etc.)
  - [ ] Make relational operators forward to member functions
  - [ ] Default instance based on attributes - introduce "comparison attributes" like equality attrs, but using `<` instead?
- [ ] `Convertible` typeclass with `toInteger`, `toFloat`, `toString`, and `toBoolean`, with instances implemented foreign functions
- [ ] `Monoid` type class
  - [ ] Allow default of monoid attributes to be inferred
- [ ] `Functor`/`Applicative`/`Monad`/`MonadFail` hierarchy
  - [ ] Fix `do`-notation
  - [ ] Refactor implicit monad extension (Dawn?)
- [ ] `Show` type class with `pp :: (Document ::= a)`
  * Have `show :: Show a => (String ::= Integer a)`
  * Default instance for `pp` attribute
  * Instance for `Show Document` where `pp = id`
- [ ] Make `deserialize` and `rewriteWith` be library functions with `reifyable` (`typeable`?) constraint
- [ ] Refactor origins using a `tracked` constraint

krame505:
Since unlike Haskell we don't have infix operator functions in Silver, what should the equivalents of `<*`, `*>` and `<*>` for the `Applicative` type class be named?

@remexre any suggestions?

remexre:
not 2 be wrong but we could just add `<*`, `*>`, `<*>`, `<$>`, `>>=`, `=<<`, `>>>`, `<<<`, `<=<`, `>=>`, `<$`, `$> :eyes: (I think these are all the ones I've ever actually used)

for names for the ones you listed, purescript names them `ap`, `applyFirst`, and `applySecond`; I've already got them written up into silver [here](https://git.sr.ht/~remexre/exps/tree/main/item/sylvan/bootstrap/xyz/remexre/category_theory/classes/Apply.sv)

krame505:
I'm vetoing non-alphanumeric function names (and I'm sure Eric would if I didn't :-P)

[...]

PR #428, opened 2021-01-11
Refactor standard lib with Eq and Ord type classes, other misc fixes

krame505:
I still need to update downstream stuff but the changes in Silver should be finished - opening this now so people can take a look if they want.  

I added `Eq` and `Ord` typeclasses mostly identical to the same in Haskell, however the symbols have been replaced with names as Silver does not have the ability to define new infix operators as functions.  Instead these are given names `eq`, `neq`, `lt`, `gte`, etc. to which the infix operators productions now forward.  Thus all type checking on them happens on the ordinary production for function application and a number of special case productions can be removed.   

Using the `Eq` and `Ord` classes, non-"By" versions of all the standard library functions are introduced (e.g. `contains`, `lookup`, `sort`, etc.)  The `empty` functions of the `silver:util:treeset`, `:treemap` and `:graph` have all been renamed to `emptyWith`, and new `empty` functions are introduced that use `Ord` for comparison.  

The `equalityTest` declaration of the Silver equality testing extension has also been updated to use `Eq` for equality, and tests have been updated appropriately.  

I added some special handling in the type checking of function application to deal with the `ntOrDecType` issue mentioned in #364.  

I also fixed a few misc. bugs with type classes and some unrelated bugs with the java translation.


grammars/silver/core/Eq.sv
```diff
@@ -0,0 +1,150 @@
+grammar silver:core;
+
+class Eq a {
+  eq :: (Boolean ::= a a) = \ x::a y::a -> !(x != y);
+  neq :: (Boolean ::= a a) = \ x::a y::a -> !(x == y);
+}
+
+equality attribute isEqualTo, isEqual;
+{- TODO: once we have occurence constraints...
+instance isEqual {isEqual} occurs on a => Eq a {
```
remexre, line 10: is the bracketed bit here a flowtype or something?

grammars/silver/core/Function.sv
```diff
@@ -0,0 +1,11 @@
+grammar silver:core;
```
remexre, line 1: hm, we could add a [Category typeclass](https://hackage.haskell.org/package/base-4.14.0.0/docs/Control-Category.html#t:Category) for these :)

the main instance other than functions worth thinking about being `Monad m => Category (Kleisli m)`

grammars/silver/core/Ord.sv
```diff
@@ -0,0 +1,234 @@
...
+}
+function gteInteger
+Boolean ::= x::Integer y::Integer
+{
+  return error("Foreign function");
+} foreign {
+  "java" : return "(%x% >= (int)%y%)";
+}
+
+instance Ord Float {
```
remexre, line 51: something something NaNs; do we have typeclass laws that permit this, or are we pulling the "okay but pretend that floats don't suck" card, or both?

grammars/silver/core/Ord.sv
```diff
@@ -0,0 +1,234 @@
+grammar silver:core;
+class Eq a => Ord a {
```
remexre, line 2: documenting laws would be good; e.g. I assumed in the (very recent) past that `compare` would only return `-1`, `0`, or `1`...

grammars/silver/core/TerminalId.sv
```diff
@@ -1,28 +1,15 @@
...
-  return t1 == t2;
-}
-
-function terminalIdLte
-Boolean ::= t1::TerminalId  t2::TerminalId
-{
-  return t1 <= t2;
-}
-
+-- TODO: Should these be generic list functions?
```
remexre, line 3: or, should terminal sets be sets? I think possibly yes for generic list functions if there's a compelling reason to keep them as lists.

grammars/silver/testing/TestSuite.sv
```diff
@@ -108,4 +112,16 @@ Boolean ::= f::(Boolean ::=)  times::Integer
...
+-- TODO: Use show typeclass
+class ShowTestValue a {
+  showTestValue :: (String ::= a);
+}
+
+instance ShowTestValue a {
+  showTestValue = \ x::a -> show(80, reflect(x).pp);
+}
+
+instance ShowTestValue a => ShowTestValue Decorated a {
```
remexre, line 124: why isn't this an overlapping instance? are type variables required to be non-decorated?

test/patt/Existential.sv
```diff
@@ -15,6 +15,14 @@ String ::= v::Existential
          end;
 }
 
+function toStringFromInteger
```
remexre, line 18: are these to be replaced by Show too? though, it'd be almost better imo if there were a separate ToString class from the Show one, and the Show one produced `Document`s, not `String`s?

test/silver_features/EqualityAttr.sv
```diff
@@ -1,37 +1,37 @@
 grammar silver_features;
 
-equality attribute isEqualTo, isEqual;
```
remexre, line 3: why not use the same one from the stdlib?

test/silver_features/Templating.sv
```diff
@@ -1,7 +1,7 @@
 
 import silver:testing;
 import silver:langutil:pp;
-import silver:util:deque;
+import silver:util:deque as dq;
```
remexre, line 4: er, does testing ambiently depend on this? otherwise, if nothing else in the file had to be changed, the import is useless, right?

remexre (commented):
only half done looking thru (by # of files), will be back after dinner

grammars/silver/core/Eq.sv
krame505: Whoops, that's a typo - should be `instance isEqual {isEqualTo} occurs on a => Eq a { ...`

Yeah, see #364 - this would be a flowtype and also a constraint that `isEqualTo occurs on a`.

grammars/silver/core/Function.sv
krame505: Yeah, will add it to future work.  

Ironically I added these but ended up going a different route that didn't need them (at one point I had a `monoid attribute ... :: (Substitution ::= Substitution) with id, compose;` which uncovered a bunch of unrelated bugs in our java translation.)

grammars/silver/core/Ord.sv
krame505: Yeah... I'm just copying Haskell here: https://hackage.haskell.org/package/ghc-prim-0.6.1/docs/GHC-Classes.html#Float

In practice I don't think we ever use these instances, currently.

test/patt/Existential.sv
krame505: Yeah - see #401 for the overall plan here.

krame505: Also this function was originally part of the standard lib, but I moved it here to not break this test that is testing something unrelated.

grammars/silver/core/TerminalId.sv
krame505: They get built in the initialization code we generate for Copper.  Not sure if that is a compelling reason or not.  

Actually it might make sense long-term to expose the underlying bit vector type that Copper uses as a new `TerminalSet` primitive (or foreign type) in Silver.  But I'm leaving this for now.

test/silver_features/EqualityAttr.sv
krame505: Hmm, I suppose it could.  This conflicted with the stdlib so I just renamed it without thinking.

test/silver_features/Templating.sv
krame505: Maybe it did at one point, but it seems like not any more.  Will remove.

grammars/silver/testing/TestSuite.sv
krame505: I am choosing to support basically Haskell's `-XFlexibleInstances` - when multiple classes match, the ambiguity is resolved in favor of the most specific match.  This lets one write a default fallback instance for a type class.  This is also the way it works in Bluespec, and I have found this ability rather useful at times.

remexre: Huh, I didn't actually know that `FlexibleInstances` allows overlaps... SGTM though, as long as this is documented.

remexre (approved):
I don't grok the type system changes (haven't read thru the typechecker, disinclined to do so now unless specifically directed); rest of it seems all good.

grammars/silver/core/Ord.sv
```diff
@@ -0,0 +1,245 @@
...
+}
+function gteString
+Boolean ::= x::String y::String
+{
+  return error("Foreign function");
+} foreign {
+  "java" : return "(%x%.toString().compareTo(%y%.toString()) >= 0)";
+}
+
+instance Ord TerminalId {
```
ericvanwyk, line 158: Why are `TerminalID`s ordered?  Is there a use case for this?

ericvanwyk (approved):

grammars/silver/core/Eq.sv
```diff
@@ -0,0 +1,160 @@
+grammar silver:core;
+
+class Eq a {
+  eq :: (Boolean ::= a a) = \ x::a y::a -> !(x != y);
```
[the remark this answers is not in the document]
remexre: `==` is now defined in terms of `eq`

grammars/silver/core/Ord.sv
remexre: Sets of `TerminalID`s, iirc

krame505: We would like to be able to sort lists of  `TerminalId`, or build a `silver:util:treeset:Set<TerminalId>`.

PR #429, opened 2021-01-12
Feature/silver typeexpr

alexandra-hanson:
I've added a production in silver construction for `Silver_TypeExpr`. This should be useful for finishing up the tuple syntax, and in general for constructing type expressions. I've also added a test file called SilverConstruction, and have included tests for `Silver_TypeExpr`, along with `Silver_Expr` and `Silver_Pattern`. Please let me know if this testing seems sufficient (I am unsure), and also if there is any other feedback you have for me with regards to this code and/or my pull request.


krame505 (approved):
Great, thanks!  

This doesn't test all of the antiquote productions (`$Expr{...}`, etc.), but you don't need to add those now unless you want to write them - previously this feature didn't have any direct tests.

PR #430, opened 2021-01-12
Replaces CopperRuntime.jar with CopperCompiler.jar.

remexre:
Currently I can't build the IDE tooling (even on `develop`) to check if this broke it, so that's maybe an area of concern.

Otherwise, this'll make it less annoying to finish `feature/copper-api`, and would be part of it anyway.

### DIFF
--- a/fetch-jars
+++ b/fetch-jars
@@ -7,9 +7,9 @@ REMOTE_STORE="http://melt.cs.umn.edu/downloads/silver-dev/jars"
 
 if [[ $* == *--copper* ]]; then
     # Only fetch the Copper jars, if requested
-    FILES="CopperCompiler.jar CopperRuntime.jar"
+    FILES="CopperCompiler.jar"
 else
-    FILES="CopperCompiler.jar CopperRuntime.jar silver.compiler.composed.Default.jar SilverRuntime.jar IDEPluginRuntime.jar"
+    FILES="CopperCompiler.jar silver.compiler.composed.Default.jar SilverRuntime.jar IDEPluginRuntime.jar"
 fi
 
 mkdir -p jars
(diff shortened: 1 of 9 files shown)


krame505 (approved):
IDE builds and works fine for me, by the way.

krame505:
Wait, actually I do have a question about this.  Is this including the entire CopperCompiler.jar in every Silver jar we build, not just silver.compiler.composed.Default.jar?  If so how much does that increase the total jar size?

remexre:
> Wait, actually I do have a question about this. Is this including the entire CopperCompiler.jar in every Silver jar we build, not just silver.compiler.composed.Default.jar? If so how much does that increase the total jar size?

Oh, hm, I hadn't thought of that... 44MB -> 46MB for ableC.jar, so not a huge increase.

krame505:
Alright, seems fair then.  

We have discussed an ableC-yacc extension where calling into Copper to build a parse table would be somewhat useful - so this would help enable that sort of extension anyway.  In the future if we do go that way, we might consider pulling the Silver FFI bindings for the Copper API into a seperate library under `silver:copper`, perhaps.

ericvanwyk:
So did we just merge the `Compiler` and `Runtime` into one?  Which seems fine to me...

remexre:
> So did we just merge the `Compiler` and `Runtime` into one? Which seems fine to me...

`Compiler` already included `Runtime`, this changes which one we depend on (and stops fetching `Runtime`).

---

Also, about to rebase this onto #428, since it's easier to re-bootstrap this than that. So don't merge this before #428 is ready!

remexre:
rebuilding in light of `feature/typeclass` merge...

PR #435, opened 2021-01-13
Feature/better aspect syntax

hypnoticOxbow:
Improves Aspect syntax such that it supports the following convenience syntax.
```
aspect isLValue on Expr of
| declRefExpr(_) -> true
| arraySubscriptExpr(lhs, _) -> lhs.isLValue
| memberExpr(lhs, _, _) -> lhs.isLValue
| _ -> false
end;
```
which would forward to (not technically what occurs, but the semantics are identical to this)
```
aspect production declRefExpr
top::Expr ::= _
{ top.isLValue = true; }

aspect production arraySubscriptExpr
top::Expr ::= lhs::Expr _
{ top.isLValue = lhs.isLValue; }

aspect production memberExpr
top::Expr ::= lhs::Expr _ _
{ top.isLValue = lhs.isLValue; }

aspect default production
top::Expr ::=
{ top.isLValue = false; }
```
When Clauses and multiple patterns with the same production are not supported. When clauses are an odd fit here, and the structure of the example seems to suggest that each production pattern produces one aspect production, and multiple patterns would cause issues with multiple aspect productions that change the same attribute (This behavior produces a warning in silver). When clauses require multiple patterns with the same production afaict, and I'm not certain how the semantics would work in the generated aspect production (what happens if the when clause is false?). If people have different ideas on those aspects please let me know.


krame505:
If there are multiple clauses with the same production, then this should generate a single aspect production containing a pattern match:
```
synthesized attribute simplify::Expr;
aspect value on Expr of
| addExpr(a, 0) -> a.simplify
| addExpr(a, b) -> addExpr(a.simplify, b.simplify)
| e -> e
```
should become
```
aspect production addExpr
top::Expr ::= a::Expr _e2::Expr
{
  top.simplify =
    case _e2 of
    | 0 -> a.simplify
    | b -> addExpr(a.simplify, b.simplify)
   end;
}
aspect default production
top::Expr ::= 
{ top.simplify = top; }
```
You might find it easiest when there are multiple patterns to just use generated names for all of the children in the production signature and matching on all of them in the production's case expression.  When-clauses can be handled similarly by including a case expression in the translation.  

If the patterns turn out to be overlapping or incomplete, then the appropriate warnings/errors should show up in the generated translation.

grammars/silver/compiler/extension/convenienceaspects/ConcreteSyntax.sv
```diff
@@ -0,0 +1,108 @@
...
+Integer ::= a
+{ return genInt(); }
+
+aspect production prodAppPattern_named
+top::Pattern ::= prod::QName '(' ps::PatternList ',' nps::NamedPatternList ')'
+{
+  local prodParamTypes::[Type] =
+    case prod.lookupValue.typeScheme.typerep of
+    | functionType(_, paramTypes, _) -> paramTypes
+    | _ -> error("invalid production type")
```
remexre, line 55: this also fails on when the user uses a not-imported type, so definitely needs some "proper" error reporting

grammars/silver/compiler/extension/convenienceaspects/ConcreteSyntax.sv
```diff
@@ -0,0 +1,108 @@
...
+    end;
+  local prodParamsList::[AspectRHSElem] =
+    map(\ty::Type -> aspectRHSElemFull(name("__generated_" ++ toString(genIntReally(ty)), top.location), ty, location=top.location),
+    prodParamTypes);
+  local prodParamNames::[QName] =
+    map((\asp::AspectRHSElem -> case asp of
```
remexre, line 61: generate the names as a list as a separate attr, and have `prodParamsList` and `prodParamNames` as `zip()`s of it? might be more readable /shrug

grammars/silver/compiler/extension/convenienceaspects/ConcreteSyntax.sv
```diff
@@ -0,0 +1,108 @@
...
+  local prodParams::AspectRHS = foldr(
+    aspectRHSElemCons(_, _, location=top.location),
+    aspectRHSElemNil(location=top.location),
+    prodParamsList);
+
+  top.makeAspect = \rhsExpr::Expr ->
+      Silver_AGDcl {
+        aspect production $QName{prod}
+        top::$TypeExpr{top.aspectTy} ::= $AspectRHS{prodParams}
+        { top.$QName{top.aspectAttr} = $Expr{caseExpr_c('case',
```
remexre, line 84: consider factoring the `caseExpr_c` out to a local, again for readability

krame505: And consider just using the abstract `caseExpr` production here instead of `caseExpr_c`.

remexre:
"stretch goal," but it'd be nice if this worked on collection attributes too. I suppose it'd have to specify `:=` vs `<-` if so

krame505:
After thinking this over a bit more, I guess I agree that it makes sense to always specify `:=` vs `<-` - this is at least consistent as both collection operators are distinct from the `=` used for regular attributes, so it makes sense that the aspect equation syntax should be different for all three as well.  

I'm not super happy about the `with :=`/`with <-` syntax though - this feels to me too much like the `with <op>` used in declaring collections (although I can live with this syntax if that's what we pick.)  What do people think about
```
aspect base errors on Expr of ... end
```
and
```
aspect contrib errors on Expr of ... end
```
?

ericvanwyk:
I would prefer `using :=` and `using <-` or something besides `with` since it is overloaded.  `aspect contrib` doesn't reall speak to me however.

It does seem odd that we write `aspect production` but here only `aspect` - instead of `aspect attribute` or something.  But it might be that `aspect production` is the thing that is a bit verbose...

krame505:
I agree `using` is probably better than `with`.  

I had the same thought about `aspect attribute`, but I do think that it is clear enough from the `<name> on <NT>` bit that `<name>` is an attribute.  Our syntax for flow specifications is `flowtype errors {env} on Expr;` which is quite similar.

ericvanwyk:
Good point on the `flowtype` stuff - `aspect <attr> on <NT> using := ...` sounds good to me.

hypnoticOxbow:
Am I still throwing a warning on '<-' or is there sufficient consensus here that this syntax won't be changed?

krame505:
I think the syntax we've settled on is good, it's not ambiguous if we always require `:=` or `<-` to specified with collections.

grammars/silver/compiler/extension/convenienceaspects/ConcreteSyntax.sv
hypnoticOxbow: Done.

grammars/silver/compiler/extension/convenienceaspects/ConcreteSyntax.sv
hypnoticOxbow: This has been done.

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
```diff
@@ -0,0 +1,150 @@
...
+        mkStrFunctionInvocation(location, "silver:core:error",
+            [stringConst(terminal(String_t,
+            "\"Error: pattern match failed at " ++ head(mRules).headPattern.location.unparse ++ "\\n\""), location=head(mRules).headPattern.location)]),
+            freshType(), location=head(mRules).headPattern.location);
+
+  local makeAspect::(AGDcl ::= Expr QName AspectRHS) =
+    \paramsCaseExpr::Expr prod::QName prodParams::AspectRHS ->
+      Silver_AGDcl {
+        aspect production $QName{prod}
+        top::$TypeExpr{aspectTy} ::= $AspectRHS{prodParams}
```
remexre, line 56: were you going to gensym top? actually, would like others' opinions on this: should `top` be in scope for body exprs? imo `top` should be a keyword at this point... (will argue this whenever we redo syntax...)

for consistency could be like

```
aspect foo on top::Bar of
...
end
```

but that's another syntax change :P

(though, that might just be changing the type to an AspectProductionLHS or w/e?)

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
```diff
@@ -0,0 +1,150 @@
...
+        { $ProductionStmt{genDef(defTop,aspectAttr,paramsCaseExpr, paramsCaseExpr.location)}}
+      };
+
+  local defTop::DefLHS = concreteDefLHS(qNameId(name("top",location), location=location), location=location);
```
remexre, line 60: see prev comment

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
```diff
@@ -0,0 +1,150 @@
...
+      },
+      [])
+    | matchRule(wildcPattern(_)::_,_,e) :: _ ->
+      pair(
+        Silver_AGDcl {
+          aspect default production
+          top::$TypeExpr{aspectTy} ::=
+          { $ProductionStmt{genDef(defTop,aspectAttr,e,head(rules).location)}}
+        },
+        [wrn(location, "wildcard patterns after this one are dead code.")])
```
remexre, line 91: uh wait what are the match semantics?

```
aspect unparse on stlc:Type of
| _ -> "unknown"
| arr(t1, t2) -> t1.unparse ++ " -> " ++ t2.unparse
end
```

what's `arr(bool(), bool()).unparse()`? `unknown` or `unknown -> unknown`

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
```diff
@@ -0,0 +1,150 @@
...
+
+}
+
+aspect production matchRule
+top::AbstractMatchRule ::= pl::[Decorated Pattern] cond::Maybe<Pair<Expr Maybe<Pattern>>> e::Expr
+{
+  top.aspectProdParamsList = case pl of
+    | prodAppPattern_named(prod, _, ps,_,_,_) :: _ ->
+      map(\pat::Decorated Pattern ->
+        name("__generated_" ++ toString(genIntReally(pat)), top.location),
```
remexre, line 135: I think `genIntReally` is unnecessary when partial function application isn't being used

self-compile
```diff
@@ -4,7 +4,7 @@
 set -eu
 
 # Defaults overridable by setting an environment variable:
-SVJVM_FLAGS=${SVJVM_FLAGS:-"-Xmx3G -Xss8M"}
+SVJVM_FLAGS=${SVJVM_FLAGS:-"-Xmx3G -Xss64M"}
```
remexre, line 7: note for other reviewers: this was needed for mwda to work on silver...

grammars/silver/compiler/modification/copper/ParserDcl.sv
```diff
@@ -2,7 +2,7 @@ grammar silver:compiler:modification:copper;
 
 import silver:compiler:driver:util only computeDependencies;
 
-terminal Parser_kwd 'parser' lexer classes {KEYWORD}; -- not RESERVED?
+terminal Parser_kwd 'parser' lexer classes {RESERVED};
```
remexre, line 5: should this be `KEYWORD,RESERVED`?

grammars/silver/compiler/extension/convenienceaspects/Project.sv
```diff
@@ -0,0 +1,15 @@
+grammar silver:compiler:extension:convenienceaspects;
+
+imports silver:compiler:analysis:typechecking:core;
```
remexre, line 3: are all of these necessary?

remexre (changes requested):

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
```diff
@@ -0,0 +1,150 @@
...
+
+  local transformPatternMatchRule::([AbstractMatchRule]::=[Decorated AbstractMatchRule]) =
+    \mRuleList::[Decorated AbstractMatchRule] ->
+      map((\mRule::Decorated AbstractMatchRule -> case mRule of
+         | matchRule(pl,cond,e) -> matchRule(
+           (foldr(append, [], (map(\pat::Decorated Pattern -> pat.patternSubPatternList, pl)) ) ),
+           cond,
+           e,
+           location=location)
+         | _ -> error("This error indicates possible productions for AbstractMatchRule have expanded.")
```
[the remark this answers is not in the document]
hypnoticOxbow: hmm, maybe I'll remove it actually, as I haven't found any other productions for matchRule, so this match is actually complete.

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
hypnoticOxbow: I wasn't going to, but I could also do it too. I don't have strong opinions on usage of top in your aspect production.

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
hypnoticOxbow: you get one default aspect production for unknown, one for the arr production
Note that these productions are grouped, so if you have a wildcard production match like this, that means there are more than one match with just wildcard

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
hypnoticOxbow: his change has been made.

grammars/silver/compiler/modification/copper/ParserDcl.sv
hypnoticOxbow: maybe. All I know is I asked "should I change this to reserved" in the meeting and got an answer of yes.

grammars/silver/compiler/extension/convenienceaspects/Project.sv
hypnoticOxbow: no, but I haven't yet gotten around to checking every one cause its like tedious. I'll fix this in a bit

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
hypnoticOxbow: ok I'm more confused by your example too. Its `unknown -> unknown` because thats what you've provided here.

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
krame505: You should definitely have some way reference `top` in easy-aspect productions.  To be consistent with strategy attributes/rewriting, maybe consider syntax such as
```
aspect foo on top::Expr of
| prod(x) -> top.bar
end;
```
and use the specified name as the LHS, if provided, or otherwise gensym it?

grammars/silver/compiler/extension/patternmatching/Case.sv
```diff
@@ -26,7 +26,7 @@ autocopy attribute matchRulePatternSize :: Integer;
 
 -- P -> E
 nonterminal MatchRule with location, config, unparse, env, frame, errors, matchRuleList, matchRulePatternSize;
-nonterminal AbstractMatchRule with location, unparse, headPattern, isVarMatchRule, expandHeadPattern;
+nonterminal AbstractMatchRule with location, unparse, env, headPattern, isVarMatchRule, expandHeadPattern;
```
krame505, line 29: Wait, why is this required?  We don't do type checking on `AbstractMatchRule`, and the env provided here would probably be incomplete/wrong.

grammars/silver/compiler/modification/copper/ParserDcl.sv
krame505: Yeah, I think we concluded that this needs to be reserved to avoid an ambiguity with `aspect parser attribute ...`

grammars/silver/compiler/extension/patternmatching/Case.sv
remexre: I did that, we need to look up production types from their names, and they're children of the AMR; it was significantly cleaner to dec the AMR than to rewrite groupMRules to decorate the matches

krame505: Hmm, I guess that's maybe OK.  But I would make sure to be clear that this is the *global* env.

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
hypnoticOxbow: this has been done now.

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
hypnoticOxbow: same deal, has been done.

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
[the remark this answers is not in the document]
hypnoticOxbow: I have removed it.

grammars/silver/compiler/extension/convenienceaspects/Project.sv
hypnoticOxbow: Got rid of the unneeded ones.

grammars/silver/compiler/extension/convenienceaspects/ConcreteSyntax.sv
hypnoticOxbow: Errors have been checked, they're properly reported later so I ignored checking this.

hypnoticOxbow:
So, I think I've mostly resolved all the outstanding issues, but the mwda analysis is returning errors
that I don't know how to resolve.

Does anybody have any ideas on how to best resolve these issues? afaict both `grammarName` and `frame` attributes are autocopy, so I'm not certain
why they aren't propagating down in these cases.

``` 
Errors for silver:compiler:extension:convenienceaspects
 [/home/ian/Projects/PL-Research/silver/grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv]
AbstractSyntax.sv:63:31: error: Pattern variable '__sv_pv_10057_name' has transitive dependencies with missing remote equations.
	Remote production: silver:compiler:extension:patternmatching:prodAppPattern
	Child: prod
	Missing inherited equations for: silver:compiler:definition:core:grammarName
AbstractSyntax.sv:121:4: error: Equation has transitive dependency on child ml's inherited attribute for silver:compiler:definition:core:frame but this equation appears to be missing.
AbstractSyntax.sv:123:20: error: Access of syn attribute matchRuleList on ml requires missing inherited attributes silver:compiler:definition:core:frame to be supplied
AbstractSyntax.sv:125:55: error: Equation has transitive dependency on child ml's inherited attribute for silver:compiler:definition:core:frame but this equation appears to be missing.
AbstractSyntax.sv:131:39: error: Equation has transitive dependency on child ml's inherited attribute for silver:compiler:definition:core:frame but this equation appears to be missing.
AbstractSyntax.sv:133:37: error: Equation has transitive dependency on child ml's inherited attribute for silver:compiler:definition:core:frame but this equation appears to be missing.
AbstractSyntax.sv:138:2: error: Forward equation exceeds flow type with dependencies on silver:compiler:definition:env:config
AbstractSyntax.sv:138:2: error: Equation has transitive dependency on child ml's inherited attribute for silver:compiler:definition:core:frame but this equation appears to be missing.
```

hypnoticOxbow:
Ok. MWDA issues have been resolved and I've added checks for wildcard patterns and behavior for `varPatterns` that creates a `let` binding with the appropriate name.

As far as documentation goes, I've added doc comments to the functions I've made, and made a PR for documentation on the website [here](https://github.com/melt-umn/melt-website/pull/20).

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
hypnoticOxbow: This behavior is resolved now. Patterns after wildcards are now ignored when aspect productions are made, and it outputs a warning if you have patterns after a wildcard.

grammars/silver/compiler/extension/patternmatching/Case.sv
hypnoticOxbow: I switched to using the concrete MatchRules and doing lookups with a function, so this change is no longer needed and was reversed.

grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
```diff
@@ -0,0 +1,518 @@
...
+
+{-
+Compares patterns, if they're
+both production patterns, compares production name
+otherwise compares the kind of pattern, (varname, wildcard, etc)
+-}
+function eqKindPattern
+Boolean ::= l::Pattern r::Pattern
+{
+  return case l of
```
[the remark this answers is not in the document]
hypnoticOxbow: oh yea tuple patterns exist now. I'll add these in a sec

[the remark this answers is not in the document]
krame505: That's not even a tuple pattern, just matching on multiple things at once.  This has been a feature for quite a while.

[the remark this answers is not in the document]
hypnoticOxbow: Ah, I misinterpreted what was added with tuple patterns. The fixes for the comments should be in now.

grammars/silver/compiler/modification/copper/ParserDcl.sv
hypnoticOxbow: Added KEYWORD also.

krame505 (approved):

krame505:
@remexre I think this PR needs you to approve it as well before it can be merged since you previously requested changes.

remexre (approved):

PR #441, opened 2021-01-17
Refactor functionType to allow partially-applied function types

krame505:
This refactors how function types are represented to bring them in line with the changes to `nonterminalType` made in #402.  The type `(String ::= Integer Float)` that was formerly represented as
```
functionType(stringType(), [integerType(), floatType()], [])
```
is now represented as
```
appType(appType(appType(functionType(2, []), integerType()), floatType()), stringType())
```
This enables us to write partially-applied function types (e.g. `(_ ::= _)`, of kind arity 2 - this is equivalent to the type `(->)` in Haskell.) This is useful as function types are instances of certain type classes, e.g. `instance Semigroupoid (_ ::= _) { ... }` - proposed by @remexre.

This change also dramatically simplifies the unification/substitution code, since we are no longer operating on lists of types and named types as the children of `functionType`.  This makes it more practical in the future to re-implement unification as some sort of automatically-propagated "unification attribute" or something in the future.

### DIFF
--- a/grammars/silver/compiler/definition/type/Unification.sv
+++ b/grammars/silver/compiler/definition/type/Unification.sv
@@ -170,12 +170,12 @@ top::Type ::= nt::Type  hidden::Type
 }
 
 aspect production functionType
-top::Type ::= out::Type params::[Type] namedParams::[NamedArgType]
+top::Type ::= params::Integer namedParams::[String]
 {
   top.unify = 
     case top.unifyWith of
-    | functionType(oo, op, onp) -> unifyFunctions(out :: params, oo :: op, namedParams, onp)
-    | _ -> errorSubst("Tried to unify function type with " ++ prettyType(top.unifyWith))
+    | functionType(op, onp) when params == op && namedParams == onp -> emptySubst()
+    | _ -> errorSubst("Tried to unify conflicting function types " ++ prettyType(top) ++ " and " ++ prettyType(top.unifyWith))
     end;
 }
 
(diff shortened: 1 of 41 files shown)


grammars/silver/compiler/definition/type/Unification.sv
```diff
@@ -170,12 +170,12 @@ top::Type ::= nt::Type  hidden::Type
...
 
 aspect production functionType
-top::Type ::= out::Type params::[Type] namedParams::[NamedArgType]
+top::Type ::= params::Integer namedParams::[String]
 {
   top.unify = 
     case top.unifyWith of
-    | functionType(oo, op, onp) -> unifyFunctions(out :: params, oo :: op, namedParams, onp)
-    | _ -> errorSubst("Tried to unify function type with " ++ prettyType(top.unifyWith))
+    | functionType(op, onp) when params == op && namedParams == onp -> emptySubst()
```
remexre, line 177: should `namedParams` and `onp` be compared to be set-equal?

krame505: No, these should always be stored in sorted order.

remexre: ah, no, because `(_ ::= foo=_ bar=_) Int String Boolean` isn't the same type as `(_ ::= bar=_ foo=_) Int String Boolean`.

though, should `(_ ::= foo=_ bar=_) Int String Boolean` be the same type as `(_ ::= bar=_ foo=_) String Int Boolean`? probably fine to leave as-is until we actually have named params

krame505: `(_ ::= foo=_ bar=_)` is parsed as `(_ ::= bar=_ foo=_)` since the named params are sorted - I think that there's a comment about this somewhere.

krame505: Well, actually there isn't concrete syntax for `(_ ::= foo=_ bar=_)` yet - the only way to get a type with named params is the type of a prod with annotations.  But if there was syntax it would work that way.

remexre (approved):

PR #445, opened 2021-01-19
Introduce and use Monoid type class

krame505:
This introduces `Monoid` and `Semigroup` type classes - `Semigroup` with `append :: (a ::= a a)` and `Monoid` as a subclass with `empty :: a`.  This separation is present in Haskell and allows for operations on types where there is an append operation but no empty value.  

Monoid attributes can now use the overloaded `append` and `empty` if no operators are specified.  The `++` operator is also changed to forward to application of `append`, eliminating the special handling previously needed for the overloaded operator.


ericvanwyk (approved):
Looks good.

PR #446, opened 2021-01-20
Subprocesses to interact with Silver

RandomActsOfGrammar:
This introduces a few functions for working with subprocesses which persist, rather than which run, producing output, and exit.  A Silver grammar may spawn such a process, send it input and read its output (both `stdout` and `stderr`), and wait for it to exit.

Let me know if there are other subprocess interactions that should be added with this now rather than in the future as they are needed.


runtime/java/src/common/ConsCell.java
```diff
@@ -48,6 +49,22 @@ public ConsCell(final Object h, final Object t) {
...
 		return cons;
 	}
+
+	/**
+	 * Creates a Java List from a cons list.
+	 *
+	 * @param l The cons list.
+	 * @return An equivalent Java List.
+	 */
+	public static <T> List<T> toList(ConsCell cons) {
```
krame505, line 59: Indentation is kinda weird here.

runtime/java/src/common/IOToken.java
```diff
@@ -359,5 +361,44 @@ private static int currentTime() {
 	public TypeRep getType() {
 		return new BaseTypeRep("silver:core:IO");
 	}
-	
+
```
krame505, line 364: As mentioned, these should probably move to `common.rawlib`

runtime/java/src/common/ProcessHandle.java
```diff
@@ -0,0 +1,99 @@
+package common;
```
krame505, line 1: Ditto.

krame505 (changes requested):
Looks good, generally.  But can you please also add some tests?

RandomActsOfGrammar:
If I move this to a library, I need to do one of the following:
1.
2.
3.

krame505:
@RandomActsOfGrammar I think part of your comment didn't save.

RandomActsOfGrammar:
My previous comment was sent before I was ready when I was trying to adjust the size of the box for writing it.  It should make more sense now.

krame505:
Hmm.  These functions should definitely take an IO token, and this really doesn't seem like it should be part of `silver:core`.  So I would maybe lean towards making `wrap` public?  Anyone else have thoughts on this?

krame505 (approved):

RandomActsOfGrammar:
It turns out---shockingly---that if the output you're reading from a subprocess doesn't end with a newline, reading a line of output  doesn't work.  Therefore I added a couple of functions that will read all available output (one for each of `stdout` and `stderr`) without looking for a particular ending.

PR #447, opened 2021-01-20
Move silver construction tests out of silver_features testsuite to speed up clean builds of silver_features

krame505:
@alexandra-hanson's change in #429 adding test cases for the silver construction syntax made the `silver_features` testsuite depend on the internals of the compiler - which makes clean builds of these tests for debugging purposes much slower.  This moves those tests into a seperate testsuite to remove that dependency.


alexandra-hanson (approved):
Looks good!

PR #451, opened 2021-01-21
Algebra typeclasses

remexre:
This won't build until docgen lands, and being able to document typeclass members (which appears NYI).


grammars/silver/core/Ring.sv
```diff
@@ -0,0 +1,29 @@
...
+  - * Left Inverse: `sub(x, x) = zero`
+  - * Right Inverse: `add(sub(zero, x), x) = zero`
+  -}
+class Semiring a => Ring a {
+  @{- The function corresponding to the binary `-` operator. -}
+  sub :: (a ::= a a);
+}
+
+@{- The function corresponding to the unary `-` operator. -}
+function negate
```
krame505, line 16: Couldn't this be a member of the Ring class with a default implementation?  Seems like this is something that could often be useful to specialize.

krame505 (commented):
Looks like a good start.  Other comments:
* Instances for Integer/Float/Boolean?
* We probably want to change the implementations of the built-in binary operators to use these class methods, right?  
* If we do change the built-in operators, then the instances for primitive types would need to rely on foreign functions (which would be great, since we can get rid of a big chunk of the translation code.)  And we might want to just change all operators to be implemented as forwarding to functions, for consistency (I think `%` is the only one not covered by the algebra stuff?)  
* If we don't change the built-in operators, then these shouldn't live in `silver:core`.

krame505:
Can we resurrect this PR?  Also this could let us make all the remaining operators to Silver forward to function calls, which would clean up translation/type checking a fair amount.

krame505:
What is the status of this pull request?  I seem to recall that we were doing this as multiple PRs, and I seem to remember something related to this being merged.  Is this branch still active?

krame505:
@remexre I see the build is passing.  is this ready to merge now, or what is left?

remexre:
I need to make negate a typeclass method and that's it... been too lazy to hunt down all the downstream branches and merge them :P I'll do it today.

remexre:
Okay, so I lied about the today... but also it looks like a jenkins-lib bug is preventing this from getting merged? Looks like if the branch ever used its own branches on child projects, using them again is "sticky" even if they got merged+deleted...

b/c of that (and since this _is_ like 8mo old, a re-review would not be out of place) gonna close this and open a new PR under a diff branch name?

krame505:
Yeah, known bug, there isn't really a sane way of querying jenkins for whether a job with a certain branch name can be built, so the logic that tries to build a branch of the same name and falls back to develop has to resort to some [evil hacks](https://github.com/melt-umn/jenkins-lib/blob/master/vars/melt.groovy#L168).

Issue #22, opened 2015-08-20
Incorrect location in pattern matching error


GoogleCodeExporter:
```
When running the following flawed code:

e.unbuild = case e1 of
              | lambda(_, lambda(_, def_foldr( apply( apply(_, apply(f,_), _) ), _, xs))) -> def_map( f.unbuild, xs.unbuild )
              | _ -> def_build( e1.unbuild )
              end;

The following errors were reported:

BuildFoldr.sv:527:93: error: Fewer patterns than expected in pattern list
BuildFoldr.sv:527:93: error: More patterns than expected in pattern list

The columns are listed as identical. This is incorrect.
```

Original issue reported on code.google.com by `kevin.ar...@gmail.com` on 20 Jun 2012 at 11:19

PR #452, opened 2021-01-22
Build for JVM 8.

remexre:


krame505:
What exactly is the meaning of this change?  Just setting the minimum version of Java that we require in the ant build files?

remexre:
Yes; with this patch, the build produces class files compatible with JVM 8 (is that the correct way to refer to it?), even on a system with a newer javac. Tested as [`feature/tuple-syntax` build `#22`](https://foundry.remexre.xyz/jenkins/job/melt-umn/job/silver/job/feature%252Ftuple-syntax/22/) (where [`#19`](https://foundry.remexre.xyz/jenkins/job/melt-umn/job/silver/job/feature%252Ftuple-syntax/19/) failed before applying it).

krame505 (approved):
Nice, thanks.  

Should help with my interm dependence on eclipse-luna which is only compatible with up to java 8...

PR #455, opened 2021-01-26
Introduce Functor/Applicative/Monad type classes

krame505:
This adds the `Functor`, `Applicative` and `Monad` type classes, and associated utility functions.  Naming conventions are followed from PureScript, and additionally the `bind` and `ap` functions are factored out into separate `Bind` and `Apply` type classes.  

Do-notation is also redesigned and no longer requires the bind and return functions to be specified explicitly.  [Applicative do](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/applicative_do.html) desugaring is used rather than always calling to `bind` and `pure`, which in some cases allows for more efficient translation, as well as the limited use of do-notation with types that are not `Monad`s but are `Functor`s and `Applicative`s.  

In order to clean up the hard-coded monad functions used in @RandomActsOfGrammar's implicit monads extension, the typeclasses `MonadPlus`, `MonadZero`, `MonadFail` and friends are also required - I will add these next in a separate pull request.


grammars/silver/core/List.sv
```diff
@@ -1,23 +1,25 @@
...
 
-{@comment
-  Applies a function to each element of the list.
+instance Functor [] {
+  map = \ f::(b ::= a) l::[a] ->
+    if null(l) then []
+    else f(head(l)) :: map(f, tail(l));
+}
 
-  This is a list of links.
```
remexre, line 6: :eyes:

remexre (approved):

PR #457, opened 2021-01-27
Fix Jenkins config which wasn't actually running tests, fix failing stdlib tests

krame505:


Jenkinsfile
```diff
@@ -155,13 +155,14 @@ def task_test(String testname, String WS) {
...
+      // Go back to our "parent" workspace, into the tests directory
+      dir(WS + '/test/') {
         // HACK: edit the test specs to specify the generated directory
-        sh "../set-generated-dir ${GEN}"
+        sh "./set-generated-dir ${GEN} ${testname}"
         // Run the tests
         withEnv (newenv) {
-          sh "java -jar ../silver.testing.bin.jar"
+	  echo "Running test ${testname}"
+          sh "java -jar silver.testing.bin.jar ${testname}"
```
RandomActsOfGrammar, line 165: Why are you running the tests for each directory under `test` separately?  Why not just run all the tests at once by using `test` as the grammar to test?

krame505: This is done to run the tests in parallel - some of the tests take a few minutes to finish while not consuming much CPU time, and this shaves several minutes off of Jenkins build times.

RandomActsOfGrammar (approved):

Jenkinsfile
RandomActsOfGrammar: It might be a good idea to add a comment explaining this.

PR #458, opened 2021-01-27
A basic file describing how to get started working on Silver

RandomActsOfGrammar:
We need a way to introduce new students to the logistics of developing the Silver system.  My hope is that this document will be something we can tell new students to read before they start working on the Silver system, and which can be referenced for some questions that might come up as they work.

I don't know if this document belongs in the Silver repository or if it should live somewhere else.

I'm hoping everyone will read this and either add their own text or suggest concepts that need to be added.  This includes people who have been working on Silver for a long time, and people who are just starting out and have their questions fresh in their minds.

A couple of notes:
* This is for people who can already write Silver code, but haven't worked on the Silver system before.  Therefore it doesn't need to explain what a production is or other similarly simple concepts.
* This isn't intended to answer every question possible.  We're going to get better results from asking people to read something short rather than a book, since they will remember more of a short text.  Everything added should be (1) rather short and (2) generally applicable.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -35,7 +35,7 @@ carefully to ensure you are thinking at the meta level as you work.
 
 
 
-## Building and Updating Scripts
+## Scripts for Logistics
 
 There are several scripts for building Silver and other logistical
 operations in the `silver` repository.
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -1,0 +1,192 @@
+
+# Getting Started
+
+This document is intended as an introduction to developing the Silver
+system.  It is *not* intended to introduce you to the concepts of
+extensible languages or using Silver, only the information you need
+for making changes to Silver itself.
+
+
+For anyone adding to this document:  
+Please keep this short.  It is just an introduction for anyone
+starting out, and we want them to actually read it.  Any questions
+which are too specific to be of general interest can be addressed to
+senior students.
+
+
+
+
+## Meta Level and Object Level
+
+One of the difficulties of developing the Silver system is keeping
+straight the meta level, the Silver compiler you are working on, and
+the object level, the Silver code you are compiling.  For example, if
+you are modifying the production for addition, you are writing code
+that works on `Expr`s with an attribute called `typerep` with a value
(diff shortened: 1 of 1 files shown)


krame505:
My vote is that this sort of documentation belongs on the website - there is already some similar (and largely duplicated) info at http://melt.cs.umn.edu/silver/dev/.  But we should definitely add a link to this from the Silver README.md file.

GETTING_STARTED.md
```diff
@@ -0,0 +1,192 @@
...
+
+The `deep-rebuild` script does several rebuilds of Silver to ensure
+that any changes which have been made still permit Silver to build
+itself.
+
+
+
+### `self-compile`
+
+The `self-compile` script does a simple build of Silver to create a
```
remexre, line 54: note that this doesn't write to jars, but sticks in in build; `cp build/silver.compiler.composed.Default.jar jars/silver.compiler.composed.Default.jar` may be necessary

GETTING_STARTED.md
```diff
@@ -0,0 +1,192 @@
...
+
+
+
+
+## Helpful Hints
+
+A good way to see where something (e.g. a function) is used in the
+Silver code base is to use the following command:
+```
+grep -r --include="*.sv" "function name"
```
krame505, line 181: I typically just use `git grep functionName` as that is quicker to type (and also shows up references in the java runtime.)  

Also there are some similar tools under `support/bin/scripts`.

GETTING_STARTED.md
```diff
@@ -0,0 +1,192 @@
...
+
+
+### `fetch-jars`
+
+This brings in new jar files which have been built by the
+continuous-integration server.  Some changes to Silver will require
+you to get new jars using this script.  If you have pulled the
+`develop` branch and it won't build, run this.
+
+
```
remexre, line 67: note that if a `new-jars` PR was just merged and jenkins hasn't finished rebuilding `develop` after the merge, this is likely to not work as expected

alexandra-hanson:
It is also good to periodically run `./update`, right?

ericvanwyk:
If it does live on the website then there should be something in Git repo that points to this information on the webpage.

PR #459, opened 2021-01-28
Checking case completeness

### LINKED ISSUE #317: Case Expressions Have No Completeness Analysis
The following compiles without raising any errors or warnings:
```
case x of
| just(a) -> a
end
```
There should probably be some form of completeness analysis to ensure that all possible constructors are matched. This would simply require checking that all constructors from the language in which the nonterminal type is introduced are matched.

RandomActsOfGrammar:
This closes #317 by adding completeness analysis for case expressions.  If it finds a case expression's patterns are not complete, it generates an example pattern which is not covered by any of the current patterns.  For a set of patterns to be complete, it either needs to have a variable or wildcard case, or have all non-forwarding productions included in the patterns.

To support testing the completeness analysis, I added a testing production to check that code does not produce a warning containing certain text.

There are several incomplete matches still present in Silver.  I didn't do anything with them because I don't know why they are acceptable, so someone who does can fill them in with default cases of the form
```
| _ -> error("<why this case isn't needed>")
```
The same thing needs to be done for downstream projects, such as AbleC.  I have no experience with these projects, and therefore I can't give good reasons for why the incompleteness is acceptable.

# Documentation
Since we've been discussing adding a required documentation section for pull requests, I will do so here.

I have a number of comments explaining the functions I wrote for checking completeness; I believe these comments are sufficient for someone else to modify the code if necessary.  These do not require doc comments because they are internal to case analysis.

I have a doc comment for the testing production I added, since this is something others might wish to use.

I am working on adding a description of completeness analysis to the website.  This should be coming shortly.


grammars/silver/compiler/extension/implicit_monads/ProductionBody.sv
```diff
@@ -38,9 +38,15 @@ top::ProductionStmt ::= 'implicit' dl::DefLHS '.' attr::QNameAttrOccur '=' ';'
   dl.defLHSattr = attr;
   attr.attrFor = dl.typerep;
 
-  forwards to if null(merrors)
-              then attr.attrDcl.attrDefDispatcher(dl, attr, case fail of | right(e) -> e end, top.location)
```
krame505, line 42: Could use `fail.fromRight` here, instead.

grammars/silver/compiler/extension/patternmatching/Case.sv
```diff
@@ -186,22 +254,366 @@ Pair<[AbstractMatchRule] [AbstractMatchRule]> ::= lst::[AbstractMatchRule]
   patterns in order, then putting each segment into its own match.
 -}
 function buildMixedCaseMatches
-Expr ::= es::[Expr] ml::[AbstractMatchRule] failExpr::Expr retType::Type loc::Location
+Pair<Expr [Message]> ::= es::[Expr] ml::[AbstractMatchRule] failExpr::Expr retType::Type loc::Location
```
krame505, line 257: Returning `Pair<Expr [Message]>` makes me suspicious that this would be better off as a pair of attributes on something (`AbstractMatchRules` if this was an actual nonterminal rather than a list?  IDK, but generally lists as abstract syntax trees are frowned upon.)  

But probably okay for now.

grammars/silver/compiler/extension/patternmatching/Case.sv
```diff
@@ -186,22 +254,366 @@ Pair<[AbstractMatchRule] [AbstractMatchRule]> ::= lst::[AbstractMatchRule]
...
+  return
+     foldr(\ grp::[[Decorated Pattern]] priorResult::Maybe<[Pattern]> ->
+             case priorResult of
+             | nothing() -> checkCompleteness(grp, env, flowEnv)
+             | just(x) -> just(x)
+             end,
+           nothing(), combinedRemoveFirst);
+}
+
+function generatePrimitiveMissingPattern
```
krame505, line 426: Wait, why do we want to be generating missing patterns?

grammars/silver/compiler/extension/patternmatching/Case.sv
```diff
@@ -298,6 +710,11 @@ top::AbstractMatchRule ::= pl::[Decorated Pattern] cond::Maybe<Pair<Expr Maybe<P
           named) ++
         tail(pl),
         cond, e, location=top.location);
+
+  top.hasCondition = case cond of
```
krame505, line 714: Could be `cond.isJust`.

grammars/silver/compiler/extension/patternmatching/PatternTypes.sv
```diff
@@ -67,6 +86,20 @@ top::Pattern ::= prod::QName '(' ps::PatternList ',' nps::NamedPatternList ')'
   top.patternSubPatternList = ps.patternList;
   top.patternNamedSubPatternList = nps.namedPatternList;
   top.patternSortKey = prod.lookupValue.fullName;
+
+  top.isPrimitivePattern = false;
+  top.isBoolPattern = false;
+  top.isListPattern = false;
+  top.patternTypeName =
+      getHeadTypeName(prod.lookupValue.typeScheme.typerep.outputType);
+  local getHeadTypeName::(String ::= Type) =
```
krame505, line 95: Why is this a "local function", since this is only used in one place?  Took me a minute to understand what was going on here.

grammars/silver/compiler/extension/patternmatching/Case.sv
RandomActsOfGrammar: It looked much nicer when compilation was done by iterative forwarding in the production, because the `[Message]` part was added into the `errors` attribute directly.  However, since we can only find the error for overlapping patterns by compiling down to no patterns, and compilation by iterative forwarding breaks completeness analysis, I think this is what we are going to have to use.

The other option, which we decided against when I moved compilation, is to have a compilation production.  This message would then just be part of the `errors` on that.

grammars/silver/compiler/extension/patternmatching/Case.sv
RandomActsOfGrammar: It generates an example of a value which is not covered by the given patterns, as well as checking that all the patterns are the same kind of primitive before generating the missing pattern.

If we have patterns `0`, `1`, and `5`, it will generate an example missing pattern of `2`, which will be shown in the warning message.  This is the same behavior as OCaml has.

grammars/silver/compiler/extension/patternmatching/PatternTypes.sv
RandomActsOfGrammar: It is a recursive function to get the absolute head name, so it needs to be a function rather than just matching.

krame505: Couldn't you just do `case prod.lookupValue.typeScheme.typerep.outputType.baseType of ...`, though? (`.baseType` extracts the applied constructor type at the base of the `appType` chain.)

RandomActsOfGrammar: I wasn't aware of the `baseType` attribute; I've changed this to use that now.

grammars/silver/compiler/extension/patternmatching/Case.sv
krame505: Ah cool, makes sense.

grammars/silver/compiler/extension/patternmatching/PatternTypes.sv
```diff
@@ -67,6 +86,16 @@ top::Pattern ::= prod::QName '(' ps::PatternList ',' nps::NamedPatternList ')'
   top.patternSubPatternList = ps.patternList;
   top.patternNamedSubPatternList = nps.namedPatternList;
   top.patternSortKey = prod.lookupValue.fullName;
+
+  top.isPrimitivePattern = false;
+  top.isBoolPattern = false;
+  top.isListPattern = false;
+  top.patternTypeName =
+      case prod.lookupValue.typeScheme.typerep.outputType.baseType of
```
krame505, line 94: Oh, this could probably also just be `.typeName` instead of pattern matching here.

krame505 (approved):
Approved - pending the downstream builds passing, of course.

krame505:
https://foundry.remexre.xyz/jenkins/job/melt-umn/job/ableC/job/develop/288/console

Is this a bug?
```
 [grammars/edu.umn.cs.melt.ableC/abstractsyntax/host/Stmt.sv]
Stmt.sv:284:16: warning: This pattern-matching is not exhaustive.  Here is an example of a case that is not matched:  _, silver:core:just(_)
```

krame505:
@RandomActsOfGrammar what is the status of this?  I took a look at fixing the case completeness errors in ableC, but the bug I mentioned above hasn't been fixed yet?

krame505:
No particular need to rush, but I do want to get this merged sooner rather than later as there are currently 8 open Silver pull requests.

RandomActsOfGrammar:
I've been working on other things, including changing implicit monads to use type classes.

I have an idea of what the bug is, but I haven't looked into fixing it yet.

RandomActsOfGrammar:
The bug in checking was due to a problem with grouping the patterns.  It turns out the `groupBy` function from `core` doesn't group everything in the list which is related together, only the first elements, so I wrote my own grouping function.

I'll take a look at the incomplete matches in Silver tomorrow, and see if there are any I can fix.

krame505:
In fixing some of the completeness errors I've noticed that the counterexamples could sometimes be more specific.  For example the following case:
```
        case decE1.isSimple, decE2.isSimple, dcls of
        | true, true, _ -> ...
        | true, false, seqStmt(_, d) -> ...
        | false, true, seqStmt(d, _) -> ...
        | false, false, _ -> ...
        end
```
gives the error `This pattern-matching is not exhaustive.  Here is an example of a case that is not matched:  _, _, edu:umn:cs:melt:ableC:abstractsyntax:host:injectGlobalDeclsStmt(_, _)`
Note that there are some ways that the `_`s could be filled in that would match successfully.  I would expect the counterexample to be something like ` true, true, edu:umn:cs:melt:ableC:abstractsyntax:host:injectGlobalDeclsStmt(_, _)`

krame505 (changes requested):
I fixed all the downstream incomplete matches, but I also noticed that convenience aspects now give bogus incomplete match warnings - see the results of building the `silver_features` testsuite.  Note that the top-level of these patterns matches should be handled by the MWDA, but we do want to warn about incomplete matches involving nested patterns since these do become cases in the translation.  

Should be fixed before this is merged, I think.

hypnoticOxbow:
Ah, so I looked at the case completeness for convenience aspects and it looks like those were legitimate issues with the case completeness for the generated aspects, caused by not providing a pattern for a case that is possible with the attribute. I have a fix for the test code available, and can push that here if you'd like.

krame505:
Yeah, go ahead and commit that - thanks.

krame505:
Huh, so
```
aspect bagList2 on top::BazExpr using <- of
| bazInit2(h::t,value) -> [h, toString(value)]
| coolName -> [coolName.hiddenAttr]
end;
```
is an incomplete match despite the default case, because the first thing has a nested pattern, and once it matches `bazInit2` we are in the equation for a particular production, and don't have an option to fall back to a different one.  That's a little weird, but I guess makes sense for what we are trying to express here.  

Do you have a test for a convenience aspect that doesn't cover all productions and doesn't have a default case gives a MWDA error rather than a match completeness error?  Or maybe rather that you can split up the equations for an attribute on a nonterminal across multiple `aspect` declarations without incomplete match warnings?

hypnoticOxbow:
No, I don't have a test for that, but could add one. The only concern is that I'm not certain that the mwda flag is passed for the silver_features tests, so the warning might not appear.

krame505:
Yeah, I meant mainly just testing that an incomplete case at the top-level doesn't produce an incomplete match error.

hypnoticOxbow:
To clarify, you're referring to cases like this right, where if we say there are 4 productions BazInit{1,2,3,4}, and we provide a new attribute and a convenience aspect like the following.
```
aspect bagList on top::BazExpr using <- of
| bazInit2(h::t,value) -> [h, toString(value)]
| bazInit3(_,val) -> [top.hiddenAttr]
| bazInit4() -> explode(top.hiddenAttr,"\t")
end;
```
where in this case we're missing one equation (bazInit1), and your question is whether this will produce an incomplete cases error? Atm the answer appears to be "no", and I can add this as a test case if you'd like.

krame505:
Yes, that is what I was wondering - e.g. if one decided to supply an equation for `bazInit1` with a regular aspect production and the rest in a convenience aspect.  If you could add a test for that it would be great.

hypnoticOxbow:
Sure, I can do that. Should I commit in this branch or make another PR?

krame505:
This branch is fine.

ericvanwyk:
@RandomActsOfGrammar - don't we handle this kind of situation in regular patterns by duplicating, in this case, the default case?

> Huh, so
> 
> ```
> aspect bagList2 on top::BazExpr using <- of
> | bazInit2(h::t,value) -> [h, toString(value)]
> | coolName -> [coolName.hiddenAttr]
> end;
> ```
> 
> is an incomplete match despite the default case, because the first thing has a nested pattern, and once it matches `bazInit2` we are in the equation for a particular production, and don't have an option to fall back to a different one. That's a little weird, but I guess makes sense for what we are trying to express here.
> 
> Do you have a test for a convenience aspect that doesn't cover all productions and doesn't have a default case gives a MWDA error rather than a match completeness error? Or maybe rather that you can split up the equations for an attribute on a nonterminal across multiple `aspect` declarations without incomplete match warnings?

RandomActsOfGrammar:
In a normal `case` expression, when the first pattern didn't match, the default case would be taken.

The issue here is whether that behavior is the right behavior for the aspect productions.  Do we expect default cases only to be used for equations in productions which weren't otherwise specified, or to be used as the default if the patterns for a production weren't complete?

krame505:
I suppose it would make sense to plug in the top-level default case as the default in each aspect production case if none is provided.  

But that is a separate issue from this pull request; I think this is ready to merge now, pending if @RandomActsOfGrammar was going to try fixing the reported counterexamples to be more specific.

krame505:
I'd like to get this merged quickly as there are several other branches waiting for this one to be merged - so I think I'm going to open a new issue for the counterexample issues and merge this now, unless @RandomActsOfGrammar has any objections.

krame505 (approved):

RandomActsOfGrammar:
I do object.  I found a bug while looking at the counterexample issue.  I'm currently working on it, and hope to have it done tonight.

krame505:
Ok, that's fine with me.

RandomActsOfGrammar:
I ended up having to rearrange the checking quite a bit to fix it.  Previously, it wasn't handling completeness checking with subpatterns correctly and missed some incomplete matches.  There are some incomplete matches in `grammars/silver/util/deque/` to be fixed which were previously missed because of this bug.  There could end up being some in AbleC as well.

This should improve the error messages for missing patterns some as well.  However, if a set of patterns is incomplete overall but an early pattern is completed by a variable, it can still show a variable instead of a missing value.  For example, the patterns
```
| 1, 2, _ -> "first" 
| _, _, 8 -> "second"
```
give a missing pattern `1, _, 0`.  If we fill in `2`, it is covered, but it isn't for any other value.  In the interest of getting this merged, I'll leave looking at that separate and handle it as #479.

grammars/silver/compiler/extension/patternmatching/Case.sv
```diff
@@ -333,93 +339,45 @@ Maybe<[Pattern]> ::= lst::[[Decorated Pattern]] env::Decorated Env
...
-                | nothing() -> nothing()
-                | just(lps) -> just(wildcPattern('_', location=bogusLoc())::lps)
-                end
-              | just(p) ->
-                just(p::repeat(wildcPattern('_', location=bogusLoc()), numPatts - 1))
-              end;
+  return
+     if numPatts == 0 || !allSameLen
+     then nothing()
+     else if isBoolPatts
```
krame505, line 335: For future reference, this sort of nested indentation with `if`s is something to avoid as it makes refactoring much more annoying.  We would typically write this as 
```
if numPatts == 0 || !allSameLen
then nothing()
else if isBoolPatts
then boolComp
else if isListPatts
then listComp
else ...
```
indenting the `then` clauses to be the same as the above `else` clauses.

krame505:
Looks like there was just one incomplete match that showed up after you fixed the bug, in the ableC-prolog extension.  I fixed it and will merge this as soon as the build passes on Jenkins.

PR #462, opened 2021-01-29
Implement Alternative, MonadFail, MonadPlus & friends

krame505:
This adds the type classes `Alt`, `Plus`, `Alternative`, `MonadZero`, `MonadPlus`, and `MonadFail`.  I believe these should be sufficient to avoid the hard-coding in the implicit monads extension, although I left out the more controversial `Either` instances for `MonadFail`/`MonadZero` until we have a chance to better discuss this (I'm still not convinced that these are safe/useful.)  

I also left out instances for `IOMonad`, since we don't do exception handling.

### DIFF
--- a/grammars/silver/core/Monad.sv
+++ b/grammars/silver/core/Monad.sv
@@ -62,7 +62,7 @@ The MonadZero type class has no members of its own; it just specifies that the t
 Instances should satisfy the following:
 
 Annihilation
-  bind(empty, f) = empty
+  bind(fail(s), f) = fail(s)
 -}
 class Monad m, Alternative m => MonadZero m {}
 
--- a/grammars/silver/core/Monad.sv
+++ b/grammars/silver/core/Monad.sv
@@ -70,14 +70,14 @@ instance MonadZero [] {}
 instance MonadZero Maybe {}
 
 {-
-The MonadZero type class has no members of its own; it just extends MonadZero with an additional law.
+The MonadPlus type class has no members of its own; it just extends MonadZero with an additional law.
 
 Instances should satisfy the following:
 
 Distributivity
   bind(alt(x, y), f) = alt(bind(x, f), bind(y, f))
 -}
-class Monad m, Alternative m => MonadPlus m {}
+class MonadZero m, Alternative m => MonadPlus m {}
 
 instance MonadPlus [] {}
 instance MonadPlus Maybe {}
--- a/grammars/silver/core/Monad.sv
+++ b/grammars/silver/core/Monad.sv
@@ -1,7 +1,5 @@
 grammar silver:core;
 
---TODO: {Alt, Plus, Alternative, MonadZero, MonadPlus, MonadFail}
-
 {-
 Monads support both lifting functions/values of arbitrary arity and sequential compostion.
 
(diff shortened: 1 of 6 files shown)


grammars/silver/core/Monad.sv
```diff
@@ -33,3 +31,53 @@ Monad m => m<Unit> ::= cond::m<Boolean>  body::m<Unit>
...
 Monad m => m<Unit> ::= cond::m<Boolean>  body::m<Unit>
 { return bind(cond, unless(_, body)); }
+
+{-
+Monads that support failure with an error message.
+
+Instances should satisfy the following:
+
+Annihilation
+  bind(empty, f) = empty
```
remexre, line 41: `fail`?

krame505: Whoops :-P

remexre (approved):
If we're requiring docs, you should add those to avoid shaming at the next meeting. If we're not doing those b/c the PR template isn't merged yet... fire away!

remexre:
Oh, though I guess it would be kinda nice to have a law on Plus and MonadFail saying something like,

"If a type is both a `Plus` and a `MonadFail`, it must additionally satisfy the law that `mzero == fail("mzero")`"

krame505:
There are docs in the comments on the type classes, which will become doc comments once @602p has those merged.

krame505:
> Oh, though I guess it would be kinda nice to have a law on Plus and MonadFail saying something like,
> "If a type is both a Plus and a MonadFail, it must additionally satisfy the law that mzero == fail("mzero")"

I thought about that, but isn't it kind of strange for a type class instance to have a law referencing another type class that isn't a direct superclass?  I think the PureScript approach here is to add an extra common subclass with the relevant laws attached, but adding a class for "things that are MonadZero and MonadPlus" felt weird, IDK.

grammars/silver/core/Monad.sv
```diff
@@ -33,3 +31,53 @@ Monad m => m<Unit> ::= cond::m<Boolean>  body::m<Unit>
...
+Annihilation
+  bind(fail(s), f) = fail(s)
+-}
+class Monad m, Alternative m => MonadZero m {}
+
+instance MonadZero [] {}
+instance MonadZero Maybe {}
+
+{-
+The MonadZero type class has no members of its own; it just extends MonadZero with an additional law.
```
RandomActsOfGrammar, line 73: This should start with `The MonadPlus type class`, not `MonadZero`.

grammars/silver/core/Monad.sv
```diff
@@ -33,3 +31,53 @@ Monad m => m<Unit> ::= cond::m<Boolean>  body::m<Unit>
...
+
+Instances should satisfy the following:
+
+Distributivity
+  bind(alt(x, y), f) = alt(bind(x, f), bind(y, f))
+-}
+class Monad m, Alternative m => MonadPlus m {}
```
RandomActsOfGrammar, line 80: Should this be `class MonadZero m, Alternative m => MonadPlus m {}` instead of `Monad m`?

RandomActsOfGrammar (approved):
It looks good, other than the two places I commented.

grammars/silver/core/Monad.sv
krame505: Oops, fixed.

grammars/silver/core/Monad.sv
krame505: Oops.

PR #464, opened 2021-01-29
Template for pull requests requiring documentation

RandomActsOfGrammar:
This provides a template for pull requests, with headings for describing the code changes made and the documentation for these changes.

This just shows up as prefilled text, so the person opening the pull request can remove it instead of following the structure; however, as Louis noted, this should be outside our threat model.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,10 @@
+# Changes
+
+# Documentation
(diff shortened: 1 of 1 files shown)


ericvanwyk (approved):
Love it!

krame505 (approved):

PR #465, opened 2021-02-01
Add a `typeable` context to fix skolem reification problem

### LINKED ISSUE #368: Reify can't handle Skolem types

krame505:
# Changes
Fixes #368.

This adds a primitive type constraint `typeable a` that requires `a` to have a determinable run-time representation.  This fixes the issue with using `reify` in polymorphic library functions; e.g. `deserialize` can now implemented as just
```
function deserialize
typeable a => Either<String a> ::= fileName::String text::String
{
  return case deserializeAST(fileName, text) of
  | left(err) -> left(err)
  | right(a) -> reify(a)
  end;
}
```
where previously this had to be a Silver extension.  

The `typeable` constraint is satisfiable for any non-skolem, non-foreign type, or a skolem type for which a `typeable` context exists.  This also means that attempting to reify a foreign type now is a compile-time rather than run-time error.   

I went ahead and did this to test out adding new kinds of specialized type constraints, as will be needed for attribute/decoration type constraints.  

## TODO
- [x] Add some more tests for using the `typeable` context in other places, e.g. as an instance constraint or superclass.  
- [ ] Write docs.

# Documentation
I will update the page on reflection to reflect these changes, before this gets merged.


krame505:
Forgot to mention this in the above writeup, but there's also a bit of a wrinkle with lambdas.  Consider the following function:
```
function foo
typeable a => ([a] ::= a) ::= x::a
{
  return \ y::a -> [x, y];
}
```
Note that the type of the returned lambda depends on the type of the function's arguments!  For example `(reify(reflect(foo(42)).fromRight)(37)` is fine, but `(reify(reflect(foo(42)).fromRight)(3.14)` is an reification error since we are attempting to reify a `([Integer] ::= Integer)` as a `([Float] ::= Float)`.  Translating the lambda requires solving a constraint `typeable ([a] ::= a)` in order to compute lambda's runtime `TypeRep` translation, which will depend on the function's `typeable a` context to utilize the `TypeRep` passed in to the function at runtime.  

However, now consider if the `typeable a` context on `foo` was omitted:
```
function bar
([a] ::= a) ::= x::a
{
  return \ y::a -> [x, y];
}
```
We still need to solve `typeable ([a] ::= a)` in order to translate the lambda, but now `typeable a` can't be resolved.  In the case of `reify` this would be a compile-time error, but we obviously don't want lambdas to always require `typeable` contexts for any mentioned skolem type vars.  The solution I took is to fall back to a rigid skolem runtime `TypeRep` (e.g. `new common.BaseTypeRep("a123")`) in these instances; thus both `(reify(reflect(bar(42)).fromRight)(37)` is fine, but `(reify(reflect(bar(42)).fromRight)(3.14)` would fail.  

I think this is probably a reasonable solution?  I don't really see a better option, here.

remexre (approved):
- imo `RuntimeTypeable` or `Reifiable` would be a better name than `Typeable`, though if documented, it's probably fine, at least by analogy to Haskell.
- if `reify` could be a foreign function, that'd plausibly be nice

ericvanwyk:
I agree with @remexre - `RuntimeTypeable` or `Reifiable` are nicer.  `Typable` is confusing since what does it mean for a type to be typable?  Does `Monomorphic` make sense here otherwise?  Or is that too restrictive?

krame505:
I went with `typeable` over `reifiable` since we could conceivably have other uses for runtime type info.  `runtimeTypeable` is a bit of a mouthful but could be okay.  Lowercase since this is a keyword, for syntactic consistency (I don't think we have any other uppercase keywords?) and since this is really a different kind of type constraint, not a type class (e.g. you can't write an instance for `typeable`.)  

Not sure how monomorphism is at all relevant, here.  There is no reason why we even need to enforce the monomorphism restriction for this constraint.

krame505:
Re: @remexre's suggestion of making `reify` a foreign function rather than a primitive, that *almost* works except that there is no way to consistently refer to a context's translation in the translation string - which should be straightforward to add.  We have `%name%` and `%?name?%` to refer to a particular argument's eager and lazy translation, but contexts don't have names so we would need to refer to them by index.  Not sure what the best syntax for this would be, how about `%@k@%` to refer to the kth context in the signature, e.g.
```
function reify
typeable a => Either<String a> ::= x::AST
{
  return error("Foreign function");
} foreign {
  "java" : return "common.Reflection.reifyChecked((originCtx!=null)?originCtx.rulesAsSilverList():null, %@0@%, (silver.core.NAST)%x%)";
}
```

PR #472, opened 2021-02-15
Make `new` a foreign function

krame505:
# Changes
This changes `new` from a built-in to a foreign function, as described in #115. 

# Documentation
None yet, this shouldn't have any significant user-facing change to the semantics of `new`, except that it can now be used without being immediately applied (e.g. `map(new, decList)`)


krame505:
Well, scratch this for the moment.  I had forgotten that `new` is treated specially by the flow analysis: it doesn't actually enforce any flow dependencies on its argument, while if `new` is a function then we are essentially taking a reference and immediately passing it into the function.  For example
```
synthesized attribute res::Expr occurs on Expr;
production foo
top::Expr ::= e::Expr
{
  top.res = new(e);
}
```
should not be a flow error, while if `new` is a function, then we will demand that the reference set is immediately supplied to `e`, before it is immediately undecorated.

PR #477, opened 2021-02-19
Implicit monad type classes and test cases

### LINKED ISSUE #463: Using type classes for implicit monads
Now that we have `Monad`/`MonadFail`/`MonadPlus` type classes, we should use them in the implicit monads extension to replace the hard-coded bind/return/mplus/mfail that work for only specific data types.  

@RandomActsOfGrammar can you take a look at this?

RandomActsOfGrammar:
# Changes
This changes implicit monads to use monad type classes instead of a predefined set of monads.  It also adds test cases, rather than just running it on some grammars.

Closes #463.

# Documentation
I added general comments for where the rewriting is non-obvious.

I didn't add any doc comments because I didn't add any new constructs.

I didn't add any documentation for users because this doesn't change anything for Silver users.

I didn't add to the website for developers.  However, I did add a README to explain general non-obvious points in the rewriting which occur  in multiple places throughout the extension.


krame505 (approved):
Looks good!  This is much nicer now than having these hard-coded.  

Also I believe the files `ListMonad.sv`, `MaybeMonad.sv` and `EitherMonad.sv` in `silver:core` can now all be deleted?

PR #482, opened 2021-03-01
Feature/polymorphism for globals

### LINKED ISSUE #453: Globals should allow polymorphism
Currently this is illegal:
```
global myMap::([b] ::= (b ::= a) [a]) = map;
```
and gives errors about types `a` and `b` not being defined.  This should be pretty easy to fix, by making `globalValueDclConcrete` put its `TypeExpr`'s lexical type variables in the env given to its children, as is done by productions and functions.  `globalValueDcl` would also need to report a `polyType` as its typerep.  

Globals should probably also allow type class constraints, which should work similarly to functions and productions.  This would require some small changes to the translation, such that globals with constraints would be translated as methods with parameters, accepting the appropriate instance dicts.  

This would be a good first issue for someone looking to get started with Silver's type system internals a bit more.

alexandra-hanson:
# Changes

This change enables globals to allow polymorphism. Type class constraints are allowed now as well. This will resolve #453.

Additionally, some refactoring was done to move the contextual information about where a constraint list occurs into a single constraintPos attribute.

Most file changes were small and are similar or identical to other declarations that allow polymorphism. The aspects on globalValueDclConcrete were changed to match the new signature that accommodates constraints. 

# Documentation

Comments were added in some of the changed files, particularly with the larger changes to globals in grammars/silver/compiler/definition/core/GlobalDcl.sv. Since most of the file changes were pretty small and/or identical to other declarations there were not many comments added there.

I am wondering if I should add a short note about polymorphism for globals to the "Global constant declarations" page on the website.


grammars/silver/compiler/definition/core/Expr.sv
```diff
@@ -187,7 +187,16 @@ top::Expr ::= q::Decorated QName
   top.unparse = q.unparse;
   top.freeVars <- ts:fromList([q.name]);
 
-  top.typerep = q.lookupValue.typeScheme.monoType; -- These aren't generalized, for now.
+  -- Type inference
+  production typeScheme::PolyType = q.lookupValue.typeScheme;
+  top.typerep = typeScheme.typerep;
+
+  -- Context resolution 
+  -- Performs final substition on all the contexts
```
krame505, line 195: substition -> substitution

krame505 (approved):
Looks good!

PR #490, opened 2021-03-18
Attempt to use the type variable names given in the code in pretty-printing types

krame505:
# Changes
Finally got fed up with confusing errors and decided to fix this -
```
global foo::(b, a) = 42;
```
would previously give an error mentioning the type `(a, b)`.  This changes things so that the pretty-printed type matches what is written.  That this is especially helpful in reporting error messages involving non trivial type class instances using type variables `a`, `b`, `c`, ..., where often the same var names are used in error messages, but in a different order.

# Documentation
None, this is mostly fixing a usability "bug" with error messages.


ericvanwyk (approved):
Looks good, thanks.

PR #493, opened 2021-03-22
Fix to MWDA handling of taking references

krame505:
# Changes
Turns out that there was a bug in #489 - we should be checking that the *maximum* possible set of inherited attributes that could be demanded from a reference are in fact available, rather than the minimum set.  In fact there are some cases where we don't know what attributes are demanded from a reference (and thus we say the reference is "unbounded"), for example
```
inherited attribute env1::[String];
inherited attribute env2::[String];
nonterminal Expr with env1, env2;
...
function getRefUnbounded
Decorated Expr with i ::= x::Expr
{
  x.env1 = [];
  return x;
}
```
Here we are claiming to return a reference to `x` that can have any set of `i` inherited attributes.  This should be some sort of flow error, as accessing an attribute that needs anything other than `env` on the result of `getRefUnbounded(...)` would be missing an equation.  

Note that something like this is okay, since the function can only be used in a way where the result has at most `{env1}`:
```
function getRefBounded
i subset {env1} => Decorated Expr with i ::= x::Expr
{
  x.env1 = [];
  return x;
}
```

The fix for this is fairly straightforward, and can reuse the same logic added in #489 for intelligently solving `subset` constraints, to raise an error when we are unable to determine a maximum set of attributes corresponding to an `InhSet` type used in taking a reference.  

# Documentation
Will be written up in my WPE paper, will eventually make it to the website.

### DIFF
--- a/grammars/silver/compiler/definition/type/syntax/TypeExpr.sv
+++ b/grammars/silver/compiler/definition/type/syntax/TypeExpr.sv
@@ -126,12 +126,12 @@ top::TypeExpr ::= InhSetLCurly_t inhs::FlowSpecInhs '}'
 {
   top.unparse = s"{${inhs.unparse}}";
   
-  top.typerep = inhSetType(sort(inhs.inhList));
+  top.typerep = inhSetType(sort(nub(inhs.inhList)));
   inhs.onNt = errorType();
 
   -- When we are in a refTypeExpr where we know the nonterminal type,
   -- decorate the inhSetTypeExpr with onNt for better errors and lookup disambiguation.
-  -- TODO: Make this Decorated FlowSpecInhs and only redecorate with onNt
+  -- TODO: Make this ref(inhs)
   production ntInhs::FlowSpecInhs = inhs;
   ntInhs.config = top.config;
   ntInhs.grammarName = top.grammarName;
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -79,7 +79,7 @@ melt.trynode('silver') {
 
   stage("Test") {
     // These test cases and tutorials are run as seperate tasks to allow for parallelism
-    def tests = ["silver_features", "copper_features", "patt", "stdlib", "performance", "csterrors", "silver_construction", "origintracking", "implicit_monads"]
+    def tests = ["silver_features", "copper_features", "patt", "flow", "stdlib", "performance", "csterrors", "silver_construction", "origintracking", "implicit_monads"]
     def tuts = ["simple/with_all", "simple/with_do_while", "simple/with_repeat_until", "simple/with_implication", "simple/host", "dc", "lambda", "turing", "hello", "stlc"]
 
     def tasks = [:]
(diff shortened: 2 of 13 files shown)


RandomActsOfGrammar:
I would prefer to see some documentation written now, even if it is preliminary, rather than waiting until you finish your WPE.

Because these changes are being merged now, people might need to read something about it now rather than in a few weeks or months when you finish your WPE.  I would also expect your WPE, which is likely focused on the theory behind this, to be quite different from documentation for users, which should be focused on the implementation and its use.

Also, it is quite easy to plan on adding documentation once a paper is completed; following through when there is something new and exciting to work on is more difficult.  Leaving the documentation until later is why we instituted the documentation requirement for pull requests in the first place.

krame505:
Understood, but
* I have finite bandwidth right now, and I am still iterating on some of the details of these features and how to best explain them.  Subset constraints and polymorphism over reference sets aren't super useful without attribute occurs-on constraints, so I imagine that the documentation would need to be updated/rewritten with better examples soon anyway.  
* Unbounded references are something that can only happen using the new `Decorated ... with ...` syntax, so not something that users can unexpectedly stumble over for the moment.  
* There may be further unforeseen issues with these features, so I don't really want to "advertise" that people should go and use them.
* This is fixing something in a previous PR that was already merged.  

So I'm going to defer to @ericvanwyk on whether I should take time to go and update the website before merging this bug fix.

ericvanwyk:
I think you're both right here.  There should be documentation but the features are not yet stable, it seems.  @krame505  - how about some placeholder documentation that at least provides the syntactic forms that are now allowed with a brief description of what they mean.  This can be simple and will then need to be updated later.  I just hate to have nothing here, as a matter of principle as much as anything else. 

Thanks.

krame505:
Ok, docs are updated.  Could @ericvanwyk and/or @remexre please review soon when you get a chance?

grammars/silver/compiler/definition/type/syntax/TypeExpr.sv
```diff
@@ -126,12 +126,12 @@ top::TypeExpr ::= InhSetLCurly_t inhs::FlowSpecInhs '}'
 {
   top.unparse = s"{${inhs.unparse}}";
   
-  top.typerep = inhSetType(sort(inhs.inhList));
+  top.typerep = inhSetType(sort(nub(inhs.inhList)));
```
remexre, line 129: Should this be sort-then-nub, if the goal is to treat the list as a set?

remexre: (There's a bunch of these, if so.)

krame505: Why would the order matter?  Removing duplicates and then sorting should be the same as sorting and then removing duplicates, I think?  And these are small enough that I don't think performance will be a concern.

remexre: Oh, I thought `nub` was like the Unix `uniq`, wrt `f [1, 2, 1, 1] = [1, 2, 1]`; yep it's not, and yep these are short.

remexre (approved):

PR #497, opened 2021-03-30
Literate Silver

remexre:
# Changes

Silver now build `.sv.md` files found in a grammar.
Code blocks that have an info string of (exactly) `silver` have the code blocks extracted and concatenated as a preprocessing step.

# Documentation

- `CompileLiterateFile.sv.md` acts as an example, testcase, and documentation of use.
- I'll probably add some user documentation to the website once locations are fixed.
- I don't think developer documentation on the website is needed, as the implementation is fairly straightforward (excepting the literal lexical reading of `extractCodeBlocks`...)

# Questions

- Should this intersect with docgen in any way?
- Multi-line foreigns?

# TODOs

- [x] Preserve locations with a `#line` directive
- [x] Website documentation
- [x] Warn about fenced code blocks without an info string

### DIFF
--- a/grammars/silver/compiler/driver/CompileLiterateFile.sv.md
+++ b/grammars/silver/compiler/driver/CompileLiterateFile.sv.md
@@ -49,7 +49,7 @@ function extractSilverCodeBlocks
 ```
 
 We do the actual extraction with the [commonmark-java](https://github.com/commonmark/commonmark-java) library.
-Tragically, foreign functions seem not to support multiline strings, so the code it generates is fairly hard to read.
+The actual conversion is in `Markdown.java` in the runtime; see there for the fine details.
 
 ```silver
 function extractCodeBlocks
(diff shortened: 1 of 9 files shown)


ericvanwyk (approved):
Looks cool.

If one wraps a Silver code block with HMTL comments I'm guessing that it is still extracted but it is then not rendered in the Markdown - is that true?  This would let use write Silver documentation that we can then tests to make sure it has valid Silver code in it.

What do error messages look like if I have a type error in my Silver code?  I'm guessing that I don't get proper line/column info because the extraction can't (using the current approach) handle multiline code blocks without mashing them into one line.

If so, this could be noted as an issue so that it gets fixed at some point.

Great idea - looking forward to using it!

grammars/silver/compiler/driver/CompileLiterateFile.sv.md
```diff
@@ -0,0 +1,85 @@
...
+[String] ::= markdown::String
+{
+  return map(\block::(String, Integer, String) -> block.3,
+             filter(\block::(String, Integer, String) -> block.1 == "silver",
+                    extractCodeBlocks(markdown)));
+}
+```
+
+We do the actual extraction with the [commonmark-java](https://github.com/commonmark/commonmark-java) library.
+Tragically, foreign functions seem not to support multiline strings, so the code it generates is fairly hard to read.
```
krame505, line 52: Honestly, I'd probably just move this to a function in the runtime. Or is that problematic with making the runtime depend on something that is only included in the compiler but not the built jars? I don't really understand how class path stuff in Java works very well.

krame505 (commented):
I would like to see line number handling before merging, if you have time. But looks good otherwise!

PR #504, opened 2021-04-08
Fixes incorrect `unification attribute ... occurs on ...;' behavior.

remexre:
# Changes

Previously, code like:

    nonterminal Repr;
    unification attribute foo {}, bar, baz occurs on Repr;

would result in:

    definition.sv.md:16:22: error: Attribute 'foo' already occurs on 'Repr'.
    definition.sv.md:16:22: error: Attribute 'foo' already occurs on 'Repr'.

# Documentation

As far as I can tell (grepping Silver and ableC) nobody actually *uses* unification attributes, so... no user-visible change -> no doc update required 🙃


krame505 (approved):
Good catch.

Unification attributes are more of an experiment than some of the other types of automatic attributes.  They used in https://github.com/melt-umn/meta-ocaml-lite/tree/feature/autoattr, which apparently never got merged?  I should probably do that.

PR #505, opened 2021-04-08
Ensures tuple selector does not hide child errors

### LINKED ISSUE #503: Tuple selectors hide child errors
This code should give an error, since `x` is an unknown name:
```
function fun
String ::=
{
  return x.1;
}
```
However, the `.1` seems to hide this error, so we end up getting an internal compiler error instead:
```
Internal compiler error: translation not defined in the presence of errors
```
We do not have this problem if we change the function body to
```
return x.fst;
```

alexandra-hanson:
# Changes

Adds a line to collect tuple.errors in top.errors, which explicitly includes the tuple errors since exprRef on line 69 of tuple/Tuple.sv assumes tuple has already been checked for errors and suppresses them. Adds Dawn's example from the original issue to the test file to show that the bug is fixed.

There are some concerns here with adding this explicit synthesized equation on the forwarding production, but Lucas said there will be upcoming changes with exprRef and changes in behavior can be dealt with then.

Resolves #503 

# Documentation
I did not add documentation since I only added one line.  Should I add a note about the concerns regarding the error collection as a comment?

### DIFF
--- a/grammars/silver/compiler/extension/tuple/Tuple.sv
+++ b/grammars/silver/compiler/extension/tuple/Tuple.sv
@@ -51,6 +51,7 @@ top::Expr ::= tuple::Expr '.' a::IntConst
   local accessIndex::Integer = toInteger(a.lexeme);
 
   top.unparse = tuple.unparse ++ "." ++ a.lexeme;
+  top.errors <- tuple.errors;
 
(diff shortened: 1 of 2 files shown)


krame505 (approved):
Thanks!  No need to add documentation for internal bug fixes.

PR #509, opened 2021-04-20
Removes Windows-style newlines.

remexre:
# Changes

`find -name '*.sv' -or -name '*.java' -exec dos2unix {} \;`

# Documentation

None needed, unless we wanna put a note somewhere being like "use Unix newlines."


ericvanwyk (approved):
Reading the diffs isn't too helpful.... :)

Looks good.

PR #510, opened 2021-04-20
Fix for #290

### LINKED ISSUE #290: aspect default productions with collection attributes
It'd be useful to specify a base equation for a collection attribute in a aspect default production, for example:

```silver
aspect default production
top::Expr ::=
{
  top.constraints := case top.inhTy of
  | just(ty) -> [constraintEq(top.synTy, ty)]
  | nothing() -> []
  end;
}
```

to ensure that checked and inferred types are the same when bidirectional typing is used.

remexre:
# Changes

Fixes #290 by checking for the bad condition (described in thread, but basically an equation existing for a collection attribute in both an `aspect default production` and any other production) and patching it when the attribute is demanded for the first time (per production, not just per node).

# Documentation

Comment pointing to #290, which has a good description of the problem.

# TODOs

- [x] Doing this at runtime is kinda sucky; waiting for Jenks to come back to see if there's a visible perf impact.
- [ ] Was trying to do this in the `postInit()` of the grammar/package that declares the occurrence of the collection attribute on the NT, but ran into the bug in my fix (I think); can do that if it'd be preferred or if this turns out to slow things down.
- [x] Regression test

### DIFF
--- a/runtime/java/src/common/CollectionAttribute.java
+++ b/runtime/java/src/common/CollectionAttribute.java
@@ -19,6 +19,7 @@ public abstract class CollectionAttribute implements Lazy {
 	
 	private ArrayList<Lazy> pieces;
 	private Lazy base;
+	public boolean appliedNTFix;
 
 	public CollectionAttribute() {
 		this.pieces = new ArrayList<Lazy>();
--- a/runtime/java/src/common/DecoratedNode.java
+++ b/runtime/java/src/common/DecoratedNode.java
@@ -314,6 +314,19 @@ public class DecoratedNode implements Typed {
 		Lazy l = self.getSynthesized(attribute);
 		if(l != null) {
 			try {
+				if(l instanceof CollectionAttribute) {
+					// This is necessary to ensure collection
+					// attribute equations in aspect default
+					// productions work; see
+					// https://github.com/melt-umn/silver/issues/290
+					CollectionAttribute prodAttr = (CollectionAttribute)l;
+					if(!prodAttr.appliedNTFix) {
+						prodAttr.appliedNTFix = true;
+						CollectionAttribute ntAttr = (CollectionAttribute)self.getDefaultSynthesized(attribute);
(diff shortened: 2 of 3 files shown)


remexre:
Eyeball statistics says this isn't actually slower; @krame505 do you care whether this is runtime vs compile-time?

krame505 (approved):
Runtime seems fine, I expect the JVM should be able to handle this branch without too much overhead.

The fix does seem like a bit of a hack, but I don't see a better way, so it's probably fine.  

Using `<-` in a default production does seem a bit weird, though - I've never come across a place where a collection production attribute wasn't sufficient.  So I might suggest just banning this, unless you have found a reasonable use case.

remexre:
My reasonable use-case is bidirectional typechecking; I think the OP of #290 was an old attempt at making Stahl in Silver.

The default aspect productions in whgvah's [type checker](https://git.sr.ht/~remexre/whgvah/tree/84fc655df2aa8db213d4134829dcaf4042980b92/item/xyz.remexre.whgvah/metalang/TypeCheck.sv.md#expressions) and [kind checker](https://git.sr.ht/~remexre/whgvah/tree/84fc655df2aa8db213d4134829dcaf4042980b92/item/xyz.remexre.whgvah/metalang/KindCheck.sv.md#metalanguage-kind-checking) are my examples here; what would these look like with a collection production attribute?

krame505:
I guess that's a pattern that I haven't thought of before, seems reasonable enough. Collection production attributes aren't at all relevant here.  

I do wonder with those examples though, why you are passing in an inherited `expType` rather than just threading around a substitution and doing unification?

remexre:
Bidirectional typing lets me avoid doing unification at all, which arguably makes things simpler. (And I haven't done HM+RankNTypes before, and didn't want to screw it up... plus I'm familiar w/ bidirectional typing and on a deadline to get this working before I present class, so didn't want to try too many new things, lol.)

Issue #516, opened 2021-04-30
Convenience Aspect ICE on nonexistent prod


remexre:
https://git.sr.ht/~remexre/whgvah/commit/ce658088c940740be8154548aaa4d60273700ebd

```
(U.91): INTERNAL ERROR: Accessing dcl of value rootConsInclude at ConcreteSyntax.sv.md:851:2
```

https://git.sr.ht/~remexre/whgvah/tree/ce658088c940740be8154548aaa4d60273700ebd/item/xyz.remexre.whgvah/metalang/ConcreteSyntax.sv.md?view-source#L851

Assume this is because `rootConsInclude` is not defined; still, shouldn't be an ICE

@hypnoticOxbow

hypnoticOxbow:
Is there like instructions written down somewhere to reproduce this? Looking at the wgvah repo I'm struggling to find a INSTALL or build.md file, etc.

remexre:
`nix build` or `build.sh` if you have nix

else, `silver --mwda xyz:remexre:whgvah` ought to work?

hypnoticOxbow:
As I was unable to reproduce this issue, and nathan wasn't able to either, I'm closing this for now. If the problem does come up again we could reopen it later.

remexre:
this repros for me right now: https://git.sr.ht/~remexre/sv2json/tree/defd246cdd628b91507ff7dbac9b304b559555d2/item/sv2json/AGDcl.sv
![image](https://user-images.githubusercontent.com/4196789/123861027-4c5d5a80-d8ec-11eb-9cb0-13e9f1ab3ff0.png)

remexre:
log: https://silo.remexre.xyz/silver/silver-issue-516-sv2json-repro.txt

@krame505 can you take a look at this, actually? I think this might not actually be convenience attributes' fault (in a moral sense), but rather that _something_ later on (possibly a production that gets fwded to) isn't documenting "you must check that this name is in scope," so convenience attributes aren't checking

PR #521, opened 2021-05-12
Don't crash when a convinience aspect has an invalid pattern type

krame505:
# Changes
When a convenience aspect has an invalid type of pattern, forward to an empty `AGDcl` instead of to `error()` to avoid an unneeded fatal crash.

# Documentation
None, this is a bug fix.


hypnoticOxbow:
Makes sense. I'm assuming the error will still be reported regardless?

hypnoticOxbow (approved):

remexre:
Is this #516?

krame505:
No, that's a different bug.

PR #531, opened 2021-06-05
Fixes environment call in convenience aspects.

hypnoticOxbow:
When looking up productions, It was using getValueDclInScope, but should
have been using getValueDcl.

-------

# Changes
Changes call to getValueDclInScope to getValueDcl for looking up productions

# Documentation
As this is a bugfix, there is no documentation as of yet. Documentation tags for the
comments already written is coming in another separate pull request.


counc009 (approved):
This fixes the bug I was seeing

Issue #512, opened 2021-04-22
GraalVM Native Image support


remexre:
Looks like as of GraalVM 21.0.0, `native-image` can build Silver and Silver programs with JVM fallback; I don't think this is actually useful yet, but it's progress so I figure I'll open a tracking issue. (Note that this issue is wholly unrelated to SilvIR-on-Truffle or anything, and is just "can we do initialization and JIT warmup at build time.")

`native-image`'d Silver (with JVM fallback) currently isn't usable; it exits with 0 when it should call out to Ant.

On a small program that does not use autocopy attributes (after patching `silver:langutil:pp` to remove its use of them), `native-image` with fallback cuts about 7% of execution time off (NB: didn't do proper stats, just eyeballed 5 runs after a 5-run warmup of each).

On the same program, `--no-fallback` fails due to Copper's use of `java.io.ObjectInputStream`; it appears a config file is needed to statically list the classes this will be used with. Providing https://p.remexre.xyz/Xj6WvLaiMAA= (as generated by the GraalVM native-image agent) via a bit of zipfile surgery, I get an amazing speedup from ~1.56sec to ~0.02sec!

Unfortunately, said config files don't work for Silver, which still dies early on when doing stuff with autocopy:

<details><pre><code>Exception in thread "main" common.exceptions.SilverInternalError: Error while applying autocopy decorators.
        at common.Decorator.decorateAutoCopy(Decorator.java:41)
        at silver.rewrite.DgivenStrategy.decorate(DgivenStrategy.java:10)
        at common.Decorator.applyDecorators(Decorator.java:22)
        at silver.core.Init.postInit(Init.java:63)
        at silver.langutil.reflect.Init.postInit(Init.java:36)
        at silver.compiler.modification.let_fix.java.Init.postInit(Init.java:54)
        at silver.compiler.composed.Default.Init.postInit(Init.java:158)
        at silver.compiler.composed.Default.Main.main(Main.java:11)
Caused by: java.lang.NoSuchFieldException: occurs_inh
        at java.lang.Class.getField(DynamicHub.java:1078)
        at common.Decorator.decorateAutoCopy(Decorator.java:36)
        ... 7 more
</code></pre></details>

I'll try replacing every `autocopy` attribute in Silver with `inherited`+`propagate` after today's meeting, I guess?

krame505:
Cool!  

Would finishing the Copper API work help at all with the ant and copper issues?

[...]

PR #533, opened 2021-06-08
Copper API

remexre:
# Changes

Changes Copper integration to use the Copper API. Before merging, will also remove the Ant requirement (which will hopefully get SubstrateVM working, #512!).

# Documentation

- [ ] Remove Ant dependency from docs, once Ant dependency is actually gone.

# TODOs

- [ ] Remove Ant
- [ ] Stop XML escaping terminal pretty-prints


krame505 (commented):
What you have here so far looks good - let me know when the bugs are fixed and this is ready for final review.

remexre:
oh, didn't realize this wasn't `feature/copper-api`; closing, and I'll open a PR from that once it works

PR #534, opened 2021-06-10
Adds Documentation for Convenience Aspects and refactors slightly.

hypnoticOxbow:
# Changes

95% of the changes are to doc comments, with one small change to move a function foldrLastElem to Core/Lists.

# Documentation

The documentation added here consists only of conversion of previous comments to doc comments,
as well as adding the appropriate doc tags.
Documentation for convenience aspects on the website for both users and developers has been pushed already.


RandomActsOfGrammar (approved):
The changes look good to me.

I don't know what is causing Jenkins to fail, but it looks like it may be a Jenkins problem.  It should be investigated before merging just to be sure it isn't something you did.

PR #536, opened 2021-06-22
Fix/doc comment arity check

602p:
# Changes
Fixes issues where doc comments containing `@param` blocks would incorrectly report an arity mismatch if:
 - A parameter had been `_`-ed out in an aspect or
 - A parameter had been specified by type only or
 - A parameter had been shorthanded by a terminal literal


grammars/silver/compiler/extension/doc/core/ArgNames.sv
```diff
@@ -36,5 +37,45 @@ top::ProductionRHSElem ::= id::Name '::' t::TypeExpr
...
+top::AspectRHSElem ::= '_'
+{
+  top.argNames = []; -- Don't consider _ parameters against count
+}
+
+aspect production aspectRHSElemFull
+top::AspectRHSElem ::= id::Name t::Type
+{
+  top.argNames = [id.name];
+}
```
remexre, line 81: trailing newline

remexre (approved):

PR #539, opened 2021-06-30
Remove use of java reflection

602p:
,


grammars/silver/compiler/translation/java/core/ProductionDcl.sv
```diff
@@ -180,36 +187,48 @@ ${makeTyVarDecls(2, namedSig.typerep.freeVariables)}
...
+                throw new common.exceptions.SilverError("Production ${fName} expected " + common.Util.namesToString(expectedAnnotationNames, "no") + " annotation(s), but got " + common.Util.namesToString(annotationNames, "none") + ".");
+            }
+            
+            ${implode("\n\t\t", map(makeChildReify(fName, length(namedSig.inputElements), _), namedSig.inputElements))}
+            ${implode("\n\t\t", map(makeAnnoReify(fName, _), namedSig.namedInputElements))}
         
-        String[] expectedAnnotationNames = new String[] {${implode(", ", map((.annoNameElem), annotationsForNonterminal(namedSig.outputElement.typerep, top.env)))}};
-        if (!java.util.Arrays.equals(annotationNames, expectedAnnotationNames)) {
-            throw new common.exceptions.SilverError("Production ${fName} expected " + common.Util.namesToString(expectedAnnotationNames, "no") + " annotation(s), but got " + common.Util.namesToString(annotationNames, "none") + ".");
+            ${if !null(namedSig.contexts) then s"""throw new common.exceptions.SilverError("Production ${fName} containes type contexts, which are not supported by reify"); // TODO""" else
```
remexre, line 222: containes

runtime/java/src/common/Util.java
```diff
@@ -394,20 +393,23 @@ private static void hackyhackyUnparseList(ConsCell c, StringBuilder sb) {
...
+		// 			.map(Util::terminalToTerminalDescriptor)
+		// 			.collect(Collectors.toList());
+		// 		return ConsCellCollection.fromList(tds);
+		// 	});
+		// } catch(Throwable t) {
+		// 	throw new TraceException("Failed to reflect to getTokens()", t);
+		// }
+
+		return ConsCell.nil;
+		//TODO: FIXME
```
remexre, line 412: I _believe_ this is only used by the never-good-and-probably-bitrotted-by-now monto-silver stuff, which is mostly removed anyway

runtime/java/src/common/Prodleton.java
```diff
@@ -0,0 +1,19 @@
+package common;
+
+
+public abstract class Prodleton<T> {
```
krame505, line 4: Would be good to add a javadoc comment for this class explaining what a "Prodleton" is (and why the name...)  

Also, indentation in this file seems messed up.

runtime/java/src/common/Reflection.java
```diff
@@ -176,27 +176,34 @@ public int compareTo(AnnotationEntry other) {
...
-				if (e.getTargetException() instanceof SilverException) {
-					throw (SilverException)e.getTargetException();
-				} else {
-					throw new SilverInternalError("Error invoking reify for class " + className, e.getTargetException());
-				}
 			}
+
+			return pton.reify(ast, rules, resultType, childASTs, annotationNames, annotationASTs);
+			
+			// Invoke the reify function for the appropriate production class
```
krame505, line 187: Delete commented-out code

runtime/java/src/common/Reflection.java
```diff
@@ -207,27 +214,35 @@ public int compareTo(AnnotationEntry other) {
...
-					throw (SilverException)e.getTargetException();
-				} else {
-					throw new SilverInternalError("Error constructing class " + className, e.getTargetException());
-				}
 			}
+
+			return tton.construct(lexeme, location);
+			
+			
+			// Invoke the reify function for the appropriate terminal class
```
krame505, line 226: Delete commented-out code

runtime/java/src/common/Terminalton.java
```diff
@@ -0,0 +1,7 @@
+package common;
+
+
+public abstract class Terminalton<T> {
```
krame505, line 4: Now there's two of them!  This is getting out of hand :upside_down_face: 

Seems reasonable to handle terminals in this way as well, but definitely needs a doc comment explaining why (and the name...)

runtime/java/src/common/RTTI.java
```diff
@@ -0,0 +1,41 @@
+package common;
+
+import java.util.*;
+
+abstract public class RTTI {
```
krame505, line 5: Needs a doc comment.  

Also, it might not be immediately obvious that RTTI is short for "runtime type information".  Not sure that is even the best name since we already do have some runtime type info infrastructure, that is unrelated to this.  Maybe even just move this to be static members/methods of the `Prodleton`/`Terminalton` classes?

remexre, line 5: Why make this abstract instead of final-with-a-private-constructor?

runtime/java/src/common/Reflection.java
remexre, line 226: zap this too

runtime/java/src/common/RTTI.java
```diff
@@ -0,0 +1,41 @@
...
+	private static final Map<String, Prodleton> productionsByName = new HashMap<String, Prodleton>();
```
remexre, line 6: `Prodleton` vs `Prodleton<?>`? (I thought the former was a warning)

runtime/java/src/common/Util.java
krame505: Uhh isn't this part of how we call every Copper parser?

krame505 (commented):
Awesome!  This is somewhat less invasive than I expected.

runtime/java/src/common/Util.java
remexre: Yes, but nothing uses the `TerminalDescriptor`s; they'd need to be removed on the Silver side too, but I'm fairly sure they can be removed.

runtime/java/src/common/Decorator.java
```diff
@@ -15,9 +15,9 @@
  * @author tedinski
  */
 abstract public class Decorator {
-	abstract public void decorate(Class<?> production);
+	abstract public void decorate(Prodleton production);
```
remexre, line 18: More `<?>` here?

runtime/java/src/common/Util.java
krame505: Aren't they also used in the eclipse IDE?  I need to check though.  And wouldn't they probably also be useful in a LSP implementation?

remexre: Dunno about Eclipse; I thought we were gonna pair tree-sitter + LSP, so we wouldn't need this for tree-sitter; if we aren't, then yeah I suppose it'd make sense to keep them around.

602p: This is a TODO because we are bizarrely (`git blame`: nathan :cowboy_hat_face: ) using reflection to call a private method on the parser. We should just update Copper to add `getTokens()` to the interface instead of doing this nonsense.

602p: (fwiw, all the CI tests pass with this changed to return `[]` so if we are using it in the IDE stuff that's the only place...)

remexre: Oh, were you gonna kill this on the Silver side? otherwise, all lgtm

remexre: Oh, were you gonna kill this on the Silver side? otherwise, all lgtm

remexre (approved):

runtime/java/src/common/RTTIManager.java
```diff
@@ -0,0 +1,84 @@
...
+ * where .getNonterminalton() -> Nonterminalton<N super T>. These are registered into these lookup
+ * tables at silver init time.
+ * 
+ * This is used for the following:
+ *  - silver reflection
+ *  - decorators (read: implementing autocopy)
+ * 
+ * @author louisg
+ */
+public final class RTTIManager {
```
krame505, line 21: I'm still not thrilled about the name `RTTIManager` ...but I also don't have a better one.

krame505 (approved):
Build time on Jenkins is essentially unchanged - so this looks good to merge.

PR #545, opened 2021-07-19
Fix occurs kind crash

krame505:
Fixes a bug with type checking for attribute occurs-on that leads to an ill-kinded type being placed in the environment even when an error is raised.


RandomActsOfGrammar (approved):
This looks fine to me.

PR #548, opened 2021-08-12
Increase maximum size of a nativley serialized list

602p:
# Changes

Backwards-compatibly adds a 32-bit-length list entry to the native serialization system to remove the size limit on lists. Also adds a test, and reworks some list helpers to avoid stack exhaustion in very large (~2^16) lists.

# Documentation

Implementation change only. Some inline docs.

### DIFF
--- a/runtime/java/src/common/ConsCell.java
+++ b/runtime/java/src/common/ConsCell.java
@@ -108,7 +108,17 @@ public class ConsCell implements Typed {
 	 * @return The length of the list.
 	 */
 	public int length() {
-		return 1 + tail().length();
+		// return 1 + tail().length(); // Blows up the stack on big lists
+
+		ConsCell x = this;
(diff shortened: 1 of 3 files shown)


runtime/java/src/common/ConsCell.java
```diff
@@ -108,20 +108,44 @@ public boolean nil() {
 	 * @return The length of the list.
 	 */
 	public int length() {
-		return 1 + tail().length();
+		// return 1 + tail().length(); // Blows up the stack on big lists
+
+		ConsCell x = this;
+		int count = 0;
+
+		while (!(x instanceof NilConsCell)) {
```
krame505, line 116: Huh, I thought the JIT would be able to do this optimization.

krame505 (approved):

PR #549, opened 2021-09-03
Fixing nOptions and adding documentation

RandomActsOfGrammar:
# Changes
The `nOptions` production for `Flag` in `silver:util:cmdargs` was not parsing arguments in the expected way, and was apparently not working at all.  This fixes it so parsing arguments with `interpretCmdArgs([("-flag", nOptions(2, prod))], ["-flag", "a", "b", "c"])` will parse `"a"` and `"b"` as the arguments to `-flag`.  This behavior is in line with the `option` production's behavior.

While I was working on this grammar, I also added doc comments to the grammar in general.

# Documentation
I added doc comments for the declarations in this grammar, which will go onto the website for users of Silver to view.

### DIFF
--- a/grammars/silver/util/cmdargs/CmdArgs.sv
+++ b/grammars/silver/util/cmdargs/CmdArgs.sv
@@ -128,11 +128,11 @@ top::Flag ::= n::Integer  ast::(CmdArgs ::= [String] CmdArgs)
       then []
       else drop(n + 1, top.flagInput);
   top.flagModified =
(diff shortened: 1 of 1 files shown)


RandomActsOfGrammar:
I should add that it does not appear the `nOptions` production is used anywhere currently, based on a search on our GitHub organization, so no other repositories need to be updated for this change.

krame505 (approved):
Looks good, wait to merge until Jenkins is fixed though.

PR #557, opened 2021-09-28
Update fetch-jars script to allow fetching jars for a specific branch or commit

krame505:
# Changes
I modified the `fetch-jars` script to optionally allow jars for a specific revision to be pulled from foundry, rather than fetching the latest jars.

# Documentation
Added a usage comment to the script

### DIFF
--- a/fetch-jars
+++ b/fetch-jars
@@ -2,8 +2,21 @@
 
 set -eu
 
-LOCAL_STORE=/web/research/melt.cs.umn.edu/downloads/silver-dev/jars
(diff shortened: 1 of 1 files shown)


remexre (approved):

remexre:
arrrrrrrrguably we should parse args better (e.g. `./fetch-jars --copper 25d563ee082e6d6672f941af765126b4bec7f52f` won't work, right?) but this is prolly fine

krame505:
Agreed, but I looked at https://stackoverflow.com/questions/192249/how-do-i-parse-command-line-arguments-in-bash and immediately decided that figuring out the "right" way was overkill here.

PR #562, opened 2021-10-01
Refactor involving various automatic attributes

### LINKED ISSUE #431: Refactoring equality attributes
This came up when discussing equality attributes with @remexre. 

We probably want to add some sort of "ordering attributes" that are similar to equality attributes but using the `<=` operator, for eventual use in deriving the `Ord` type class.  The inherited attribute in these that deconstructs the corresponding tree (`isEqualTo` in `equality attribute isEqualTo, isEqual;`) would be defined the same between equality and ordering attributes; additionally this type of attribute may be useful on its own.  Thus we may want to instead write
```
deconstruction attribute compareTo;
equality attribute isEqualTo with compareTo;
ordering attribute isLessOrEqualTo with compareTo;
```
This would be a fairly straightforward change.  I don't really like the name `deconstruction` though, any better ideas?  

Note that these are sort of an inherited counterpart to functor attributes: for example propagating `functor attribute foo;` on `addExpr` would give
```
aspect production addExpr
{
  top.foo = addExpr(e1.foo, e2.foo);
}
```
while `deconstruction attribute bar;` would give
```
aspect production addExpr
{
  e1.bar = case top.bar of addExpr(a, _) -> a | _ -> error("demanded when no match") end;
  e2.bar = case top.bar of addExpr(_, a) -> a | _ -> error("demanded when no match")  end;
}
```
Note that for the same values of the attributes on the children, the top value would be the same, except that the information flows in the opposite direction.

krame505:
# Changes
Fixes #431 

* Separates "destruct" attributes that were a part of equality attributes into something that can be propagated separately
* Destruct attributes are now decorated
* Implemented ordering attributes
* Rename `isEqualTo` to `compareTo`, shared with `compare`/`compareKey` ordering attribute pair
* Added default instances for `Eq` and `Ord` using equality and ordering attributes
* Removed instances for the above that are now obsoleted by these default instances
* the unfinished prototype of "unification attributes" was simplified to also reuse destruct attributes

# Documentation
Currently being written for the website, will make a separate pull request for that when ready.

### DIFF
--- a/grammars/silver/compiler/definition/core/InstanceDcl.sv
+++ b/grammars/silver/compiler/definition/core/InstanceDcl.sv
@@ -12,7 +12,8 @@ top::AGDcl ::= 'instance' cl::ConstraintList '=>' id::QNameType ty::TypeExpr '{'
   production dcl::DclInfo = id.lookupType.dcl;
   dcl.givenInstanceType = ty.typerep;
   
-  production superContexts::Contexts = foldContexts(if id.lookupType.found then dcl.superContexts else []);
+  production superContexts::Contexts =
+    foldContexts(if id.lookupType.found && !foldContexts(cl.contexts).isTypeError then dcl.superContexts else []);
   superContexts.env = body.env;
   
   top.defs := [instDef(top.grammarName, id.location, fName, boundVars, cl.contexts, ty.typerep)];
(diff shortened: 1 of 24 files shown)


krame505:
Note that this includes the changes from #563 and #561 and should probably be reviewed after those.

hypnoticOxbow (approved):

PR #569, opened 2021-10-09
Fix to default production environment

krame505:
# Changes
The environment for the type in default productions wasn't being handed properly, thus
```
nonterminal Foo<a>;
aspect default production
Foo<a> ::=
{}
```
would raise an error on the `a` in the signature.  

# Documentation
None, this is a minor bug fix


remexre:
Ooh, I hadn't thought about that case before! Out right now so don't have time to review till tonight or tomorrow, but is it legal to write an aspect default for Foo<String>?

krame505:
A default for `Foo<a>` should be legal if that's what you meant.  Just `Foo` would have the wrong kind.

krame505:
Do you mean `Foo<String>`?  That definitely isn't legal (and isn't showing up on github either)

krame505:
Hmm I guess we never check that the LHS in a default prod is a nonterminal either:
```
aspect default production
top::Integer ::=
{}
```
I'll go fix that too.

PR #575, opened 2021-10-12
Nix and Vim stuff

remexre:
# Changes

A small tweak to the Vim syntax, and Nix stuff to automatically put Silver's dependencies into `$PATH` (if the user opts in via direnv).

I don't have `ant` or `wget` in my system-wide or user-wide `$PATH`, so this is somewhat nicer.

Also changes shebangs to use `/usr/bin/env` to find bash, since it isn't in `/bin` on NixOS (and it's not on the BSDs either).

# Documentation

Documentation not written, but once the [melt.nix](https://git.sr.ht/~remexre/melt.nix) thing is done (needs some way to hook into Jenkins, a fix for https://github.com/melt-umn/jenkins-lib/issues/19, and the rest of the ableC extensions getting packaged before I'm "announcing it") the Nix stuff here probably deserves a subsection.

### DIFF
--- a/support/vim/syntax/sv.vim
+++ b/support/vim/syntax/sv.vim
@@ -10,7 +10,7 @@ endif
 
 syn keyword svlangExternals grammar import imports exports build hiding as only option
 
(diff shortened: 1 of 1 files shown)


remexre:
fetch-jars uses wget, I use curl, but not out of any antipathy for wget, I just know the flags better.

Issue #10, opened 2015-08-20
Refactor EnvItem


GoogleCodeExporter:
```
Provide an explanation of what the project should entail:

EnvItem makes a few mistakes:
1. It's hard coded to 'Decorated DclInfo', when it should be parameterized (see 
issue 3)
2. It's passed around 'Decorated' and shouldn't be.
3. It's basically written like a dumb Pair type, when it should be smarter than 
that.

Provide a concrete example of a problem this project will fix:

abstract production foo
bar::Foo ::= baz::String foobar::Integer
{}
aspect production foo
baz::Foo ::= foobar::String bar::Integer
{}

Should be valid. It fails because the environment doesn't work properly because 
of point #3 above...

Provide a short guide on how to start this project:

The place to start is silver:definition:env EnvItem.sv

Provide a rough expectation of what the set of patches completing this
project might look like:

1. Make it not Decorated.
2. Parameterize it, if it hasn't been already.
3. Make it three productions, instead of three functions yielding one 
production.
3. Murder with prejudice the 'explodeEnvItems' function in EnvTree.sv SEE BELOW

Probable challenges in completing this project:

Figuring out the env item renaming helper functions in there. That's all 
related to 'import foo:bar only foo, bar with foo as foozilla, bar as 
barzilla;' etc etc.

Additional Information:

Here's the "SEE BELOW". Remember that bug with aspect I mentioned?

Right now there are three kinds of "EnvItems":
1. defaultEnvItem  --  "a:b:c:d" becomes "a:b:c:d" and "d"
2. renamedEnvItem  --  "a:b:c:d" becomes "a:b:c:d" and something else
3. fullNameEnvItem --  "a:b:c:d" becomes "a:b:c:d"

We're missing a fourth case:
4. onlyRenamedEnvItem --  "a:b:c:d" becomes something else entirely, no 
"a:b:c:d" at all

This matters for that aspect bug, because it means "bar" gets 'renamed' to 
"baz" in the aspect... but is still around as "bar" because, well, that missing 
case is missing.

So silver:definition:env Defs.sv in the add defs functions for aspect aliases 
needs case 4.

The end result is that EnvItem<a> should have four productions (not functions, 
productions:) those four just listed above, three of which are currently 
functions.

And there should be an attribute 'envContribs :: [Pair<String a>]' on 
EnvItem<a> that replaces the functionality of 'explodeEnvItems'.

I hope this makes sense! :D
```

[...]

PR #578, opened 2021-10-12
Major environment refactor, splitting DclInfo into multiple nonterminals

### LINKED ISSUE #3: Break DclInfo down into multiple types
```
Provide an explanation of what the project should entail:

The type "DclInfo" is used to represent some generic information about 
declarations in Silver.

Currently, this type is the same for all different kinds of declarations:
1. values
2. types
3. attributes
4. occurs
5. lexer classes
6. probably more!

The purpose of this project is to make each of these differents types of 
declarations actually
different TYPES.  i.e. ValueDclInfo, TypeDclInfo, AttributeDclInfo, 
OccursDclInfo, etc.

Provide a concrete example of a problem this project will fix:

The change will be somewhat superficial for the moment, because we don't have 
the tools (YET)
to take advantage of this change.

In the future, the advantage will be that the set of attributes that need to be 
defined for a
DclInfo will be machine-discernable and checkable.  (Currently, there are 
attributes that
occur on DclInfo that only apply to one/some of the above.  Lexer classes do 
not have types, etc.)

At the present, it will simply make the code much higher quality, and easier to 
understand.

Provide a short guide on how to start this project:

Begin by looking at EnvItem and EnvTree in silver:definition:env.
and at QNameLookup in silver:definition:core QName.sv.
Then take a look at Env and Defs in silveR:definition:env.

Provide a very rough and probably inaccurate expectation of what the set of 
patches completing this project would look like:

1. Parameterize EnvItem, EnvTree, and QNameLookup, but leave everything as is
   (i.e. the parameter is always 'Decorated DclInfo' in Env and Defs).
2. Start by trying to remove 'Decorated' from occurs declarations. This should 
shake-out any problems by starting to have
   the type be different, without requiring too many invasive changes.
3. Now introduce an OccursDclInfo nonterminal that has only occurs-declaration 
relevant info on it.
4. After each of these change is done, working, and committed, continue until 
they are all done.

Probable challenges in completing this project:

You find (and be able to fix) a couple of kludges during this.  For example 
there's some awful code
involved in getting the type of an attribute from the occurs declaration and a 
nonterminal type that
currently de-decorates and re-decorates the dclinfo, which doesn't need to 
happen once it's not
passed around decorated anymore.

Do TypeDclInfo and LexerClassDclInfo last, because once to try to do them, 
you're going to run into
some awful code in the copper modification that you'll probably need my help to 
fix.  Alternatively,
forgo introducing LexerClassDclInfo for the time being, and make it just be a 
TypeDclInfo.
(This is related to some places where "terminals or lexer classes" are 
expected, so there will
be some difficulty in having two different types.)

Additional Information:

Technically, removing the 'Decorated' from DclInfo could be a whole separate 
project to itself.


```

Original issue reported on code.google.com by `tux...@gmail.com` on 10 Jan 2012 at 12:44

krame505:
# Changes
Fixes #3 and #10. 

This splits the `DclInfo` nonterminal into separate `ValueDclInfo`, `TypeDclInfo`, `AttributeDclInfo`, etc. nonterminals for different sorts of declarations.  Much of the env plumbing is now parameterized by the type of declaration being considered.  I also fixed a number of flow errors.  

# Documentation
This is an internal refactoring with no observable behavioral changes.  Comments have been added to the source where appropriate.

### DIFF
--- a/grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
+++ b/grammars/silver/compiler/extension/convenienceaspects/AbstractSyntax.sv
@@ -8,7 +8,7 @@ import silver:compiler:modification:let_fix;
   - @param defaultLoc the location to provide for the patternListNil construct (which needs a location).
   - @return A patternList List-Like Nonterminal instance.
   -}
-function makePatternListfromListofPatterns
+function makePatternListFromListofPatterns
 PatternList ::= l::[Pattern] defaultLoc::Location
 {
   return
@@ -48,7 +48,7 @@ function collectPatternsFromPatternList
 function extractSubPatternListsFromProdPatterns
 PatternList ::= pl::PatternList
 {
-  return makePatternListfromListofPatterns(
+  return makePatternListFromListofPatterns(
     foldr(
       append,
       [],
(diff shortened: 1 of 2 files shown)


krame505:
I figured out a way to make occurs-on (and also `runtimeTypeable`) contexts work with `reify` and `nativeDeserialize`, by looking at the runtime type of a child that has the type mentioned in the context.  This fixes the issue with deserializing the environment and non-clean builds are working again now.

remexre:
Ooh, nice; I think I should be able to finish reading over it tonight.

PR #590, opened 2021-11-11
Adds upload-override-jars script.

remexre:
# Changes

Adds a script for uploading JARs to Jenkins for use with `OVERRIDE_JARS`.

# Documentation

Does this need docs?

### DIFF
--- a/upload-override-jars
+++ b/upload-override-jars
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+
+if [[ -n "$(git status --porcelain)" ]]; then
+	echo "You have uncommitted changes! Refusing to run."
+	exit 1
+fi
+
+branch_name="$(git rev-parse --abbrev-ref HEAD)"
+commit_hash="$(git rev-parse HEAD)"
+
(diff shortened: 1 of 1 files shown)


upload-override-jars
```diff
@@ -0,0 +1,18 @@
...
+
+if [[ -n "$(git status --porcelain)" ]]; then
+	echo "You have uncommitted changes! Refusing to run."
+	exit 1
+fi
+
+branch_name="$(git rev-parse --abbrev-ref HEAD)"
+commit_hash="$(git rev-parse HEAD)"
+
+rsync -Pa jars/ foundry.remexre.xyz:/melt/jenkins/export-scratch/melt-jenkins/${commit_hash}-jars/
```
krame505, line 15: Not super familiar with rsync, but will this work if the user from which the script is run doesn't happen to match the user on foundry?

remexre: One can provide a username to use via `~/.ssh/config` (rsync just uses SSH under the hood)

krame505: You could maybe add a comment to the script mentioning that.  I guess it's maybe not a huge issue anyway as only melt people should ever run this, and we can pick the account names on foundry.

remexre: Added.

krame505 (approved):

PR #593, opened 2021-11-19
Fix issue with type class constraints on functions called in rewrite rules

krame505:
# Changes
Previously functions in type constraints weren't permitted in rewrite rules since in the translation, the type of the referenced function would be unconstrained, and the constraints would not be resolved successfully.  By inserting let expressions in the transformed expression with the originally inferred types, this can be avoided.  Not sure why I didn't think of this sooner....

# Documentation
This fixes an undocumented bug/missing feature, so this is also a documentation win I guess?


krame505:
Ok, a couple of other refactorings and bug fixes related to run-time type checking made it in here,.  This should actually be ready for review now.

remexre (approved):

PR #597, opened 2021-12-07
Updates to silver emacs mode

krame505:
# Changes
Adds some missing keywords and removes some long-deprecated ones.

# Documentation
This does not affect any source code.


RandomActsOfGrammar (approved):
For once, I agree that you don't need any documentation.

PR #598, opened 2021-12-09
Some fixes for pp"""...""" tempates

krame505:
# Changes
The silver:langutil:pp library assumes that strings passed to the `text(...)` constructor do not contain newlines, as including them breaks indentation.  However the `pp"""..."""` multi-line string template quite happily creates `text` blobs containing them.  This fixes the translation of templates to use the `realLine()` constructor instead (which always introduces a newline, but respects indentation in rendering.)

A similar issue exists for single-line `pp"..."` templates with `\n` - this is also now translated as a `realLine()`.  

Also some cleanup to the implementation (now making use of the meta-translation library) and some unrelated bug fixes (The regex `/\\n/` was incorrectly being flagged as containing a literal newline.)

# Documentation
Added comments in the source code, where needed.


remexre (approved):
lgtm, tho metatranslation remains subtle

PR #599, opened 2021-12-10
Support .ag file extension as an alternative to .sv

krame505:
# Changes
This permits .ag as an alternative file extension to .sv - .sv conflicts with System Verilog, which poses some unavoidable issues for a certain use case of Silver.  Message me privately if you want more details.  

I updated the vim and emacs plugins to recognize both extensions, but the gedit plugin install script is broken for modern versions of that editor, so I'm leaving that one alone for now.  

# Documentation
None yet.  As best I can tell we never say that "sv is the canonical file extension for Silver" in any existing documentation, so nothing is out of date per-se, but we may want to add something?

### DIFF
--- a/grammars/silver/compiler/driver/CompileLiterateFile.sv.md
+++ b/grammars/silver/compiler/driver/CompileLiterateFile.sv.md
@@ -14,18 +14,18 @@ String ::= path::String  contents::String
 If the file is a normal Silver file, we return the contents unchanged.
 
 ```silver
-    if endsWith(".sv", path)
+    if endsWith(".sv", path) || endsWith(".ag", path)
     then contents
 ```
 
 If it's a Literate Silver file instead, we extract the relevant code blocks, then join them with newlines.
 
 ```silver
-    else if endsWith(".sv.md", path)
+    else if endsWith(".sv.md", path) || endsWith(".ag.md", path)
     then implode("\n", extractSilverCodeBlocks(path, contents))
 ```
 
-Since these are the only two extensions allowed by `isValidSilverFile` in `CompileGrammar.sv`, they're the only two we need to handle.
+Since these are the only extensions allowed by `isValidSilverFile` in `CompileGrammar.sv`, they're the only two we need to handle.
 
 ```silver
     else error("Unknown filetype for " ++ path);
--- a/grammars/silver/compiler/driver/CompileLiterateFile.sv.md
+++ b/grammars/silver/compiler/driver/CompileLiterateFile.sv.md
@@ -25,7 +25,7 @@ If it's a Literate Silver file instead, we extract the relevant code blocks, the
     then implode("\n", extractSilverCodeBlocks(path, contents))
 ```
 
-Since these are the only extensions allowed by `isValidSilverFile` in `CompileGrammar.sv`, they're the only two we need to handle.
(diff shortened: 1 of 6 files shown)


grammars/silver/compiler/driver/CompileLiterateFile.sv.md
```diff
@@ -14,18 +14,18 @@ String ::= path::String  contents::String
...
 
 If it's a Literate Silver file instead, we extract the relevant code blocks, then join them with newlines.
 
 ```silver
-    else if endsWith(".sv.md", path)
+    else if endsWith(".sv.md", path) || endsWith(".ag.md", path)
     then implode("\n", extractSilverCodeBlocks(path, contents))
 ```
 
-Since these are the only two extensions allowed by `isValidSilverFile` in `CompileGrammar.sv`, they're the only two we need to handle.
```
remexre, line 28: four now, right?

grammars/silver/compiler/driver/util/Util.sv
```diff
@@ -3,6 +3,8 @@ grammar silver:compiler:driver:util;
 imports silver:compiler:definition:type;
 imports silver:compiler:definition:env;
 
+global allowedSilverFileExtensions::[String] = [".sv", ".ag", ".sv.md", ".ag.md"];
```
remexre, line 6: Maybe leave a comment noting that `CompileLiterateFile.sv.md` needs updating if the list changes?

grammars/silver/compiler/extension/doc/core/RootSpec.sv
```diff
@@ -35,8 +44,8 @@ function toSplitFiles
   return case g of
        | consGrammar(this, rest) ->
            if getSplit(this.localDocConfig) then toSplitFiles(rest, grammarConf, forIndex, formatFile(
-             substitute(".sv", ".md", this.location.filename),
-             getFileTitle(this.localDocConfig, substitute(".sv", "", this.location.filename)),
+             silverToMdFilename(this.location.filename),
```
remexre, line 47: Is this broken for files with names like `foo.sv.sv`? If so I think it would be broken before too and nobody should do that, so I'm fine if this doesn't get fixed now.

remexre (approved):
Fine with me, though this means we need a different file extension for the hypothetical cleaner-alternate-Silver-syntax in the future. (I suppose `.sv2`/`.ag2` would work for that though lol.)

remexre:
I guess a topic of discussion might be, do we want to encourage people to write new code with `.sv` or `.ag` if this lands? (We might wanna discuss this on Monday unless the usecase that requires this is urgent enough that it's important that this land before then.)

krame505:
I can't make it to the Monday meeting unfortunately, I'm giving a talk about Silver at the same time.  

This can wait on merging until Monday, but I wouldn't like to wait any later than that.

remexre:
Hm, maybe we should move the meeting earlier in the day?

ericvanwyk:
Are we doing this only because of how GitHub matches files to languages?  I dont' see a real benefit of doing this, but it doesn't matter much either.  If we do ever have an alternative syntax we can reclaim .ag by just moving all these files back to being .sv files...

krame505:
See my message to you on Slack

PR #600, opened 2021-12-13
Feature/cleaner pattern matching

### LINKED ISSUE #387: Pattern Matching Clause Order and Forwarding
Consider the following Silver grammar:
```
nonterminal B;

abstract production b1
top::B ::=
{ }

abstract production b2
top::B ::=
{ forwards to b1(); }

abstract production b3
top::B ::= x::B y::B
{ }

function matchB
String ::= b::B
{
  return case b of
         | b3(b1(), b2()) -> "b3 of b1, b2\n"
         | b3(b2(), b1()) -> "b3 of b2, b1\n"
         | _ -> "other\n"
         end;
}
```
The result of `matchB(b3(b2(), b2()))` is `"b3 of b2, b1\n"`, which comes from matching the pattern in the second rule.  We can see why we get this result by looking at (an expression equivalent to) the compiled version of this match:
```
case b of
| b3(a, b) -> case a of
              | b1() -> case b of
                        | b2() -> "b3 of b1, b2\n"
                        | _ -> "other\n"
                        end
              | b2() -> case b of
                        | b1() -> "b3 of b2, b1\n"
                        | _ -> "other\n"
                        end
              | _ -> "other\n"
              end
| _ -> "other\n"
end
```
We first match `b3`, then we match `b2()` against `b1()` and `b2()`.  Because it matches `b2()` better, even though that is not the first pattern, we take that match.  We then try to match `b2()` to `b1()`, which succeeds by forwarding.

In a way, it seems that we should take the first match rule in the original match above.  While the first child of `b3` matches better in the second match rule, the second child doesn't match perfectly.  This is the opposite of the first match rule, where the first child doesn't match perfectly but the second child does.  In some sense, both are equally good matches, overlapping in matching behavior even though they don't overlap syntactically, and therefore the first one should be taken.  If we had a rule with the pattern `b3(b2(), b2())` later in the patterns, we would want to take that because it matches the structure better than either of the other two.

So it seems we want to determine when two patterns match "equally well" and take the first one.  However, this gets really complicated with more things being matched.  Suppose we are matching `a, b, c` against a couple of patterns `p1a, p1b, p1c` and `p2a, p2b, p2c`.  Suppose we forward zero times to match `p1a`, zero times to match `p1b`, and two times to match `p1c` (forwarding vector, the number of times it forwards to match each successive match, `(0, 0, 2)`) and we forward one time to match `p2a`, one time to match `p2b`, and zero times to match `p2c` (forwarding vector `(1, 1, 0)`.  Which pattern set matches the input better?
* `p1a, p1b, p1c` matches better, because only one thing forwards
* `p2a, p2b, p2c` matches better, because we only forward at most once in any single match

The current method of left precedence (whatever matches best furthest left) gives us a method of determining the best match without needing to consider the complexities of matching "equally well", but it can give us unintuitive results.  I'm not sure if this should be considered as a bug or just a strange bit of Silver matching behavior that should be documented.

-------------------------------------------------------------------------------------------

That's where I was going to end this.  However, I tried a related match and found something that is at least a bit related which I'm more certain should be considered a bug.  Perhaps this should go in its own issue, but I'll put it here to start.

Suppose we change our `matchB` function but leave the productions alone:
```
function matchB
String ::= b::B
{
  return case b of
         | b3(b1(), b2()) -> "b3 of b1, b2\n"
         | b3(b2(), b3(_, _)) -> "b3 of b2, b3\n"
         end;
}
```
What should be our result for `matchB(b3(b2(), b2()))` now?  We have two match rules.  The second one doesn't match, but the first one does match if we forward the first child, so our result should be `"b3 of b1, b2\n"`.  Silver begs to differ:
```
Error: pattern match failed at temp Temp.sv:118:9

An error occured.  Silver stack trace follows. (To see full traces including java elements, SILVERTRACE=1)

(DN.315): While evaling syn 'core:io' in 'core:ioval' (dd3b207)
(core.Pprint in Pprint.java:102): Error while evaluating function core:print
(temp.PmatchB in PmatchB.java:105): Error while evaluating function temp:matchB
(core.Perror in Perror.java:93): Error while evaluating function core:error
(U.70): Error: pattern match failed at temp Temp.sv:118:9

```
Why does this fail?  Let's look at how it compiles:
```
case b of
| b3(a, b) -> case a of
              | b1() -> case b of
                        | b2() -> "b3 of b1, b2\n"
                        end
              | b2() -> case b of
                        | b3(_, _) -> "b3 of b2, b3\n"
                        end
              end
end
```
Once we have matched the `b2()` for the first child, we are committed to the first child being a `b2()`, and we don't backtrack and consider that it might match another pattern as well.  Thus we get a `case` expression containing a matching pattern which gives a match failure error.

The issue which causes this unintuitive behavior is that we implemented pattern matching based on Phil Wadler's chapter in *The Implementation of Functional Programming Languages*, where if an input matches a certain constructor, it *cannot* match a different constructor.  Since we will match through forwarding, this assumption does not hold in our case.

-------------------------------------------------------------------------------------------

So @tedinski and @ericvanwyk, have I found at least one bug?  Is this expected behavior and I just need to read the functional-programming part of Ted's thesis to understand what is going on here?  Even if there are no bugs here, we need to work on the documentation so it describes these cases.

RandomActsOfGrammar:
# Changes
Closes #387.  This makes the semantics of pattern matching linear from the top down, rather than searching for a forwarding production, then backtracking.  The details of what the original problem was can be found in #387, as well as what this changes it to do.

# Documentation
I wrote numerous comments explaining the functions I wrote and what they are doing, as well as describing what is happening in the places they are called.

I did not add any doc comments because nothing I have changed is intended to be used outside the file changed other than by building a case expression.

I did not add any documentation to the website because this is consistent with what I see already on the website about pattern matching and changes the behavior to the obvious behavior.


grammars/silver/compiler/extension/patternmatching/Case.sv
```diff
@@ -159,16 +182,143 @@ Pair<[AbstractMatchRule] [AbstractMatchRule]> ::= lst::[AbstractMatchRule]
...
+                         decorate wildcPattern('_', location=loc) with {
+                            frame=hd.frame; config=hd.config; env=hd.env;
+                            patternVarEnv=hd.patternVarEnv;
+                         }
+                       | just(p) -> p
+                       end,
+                     annos) ++ tl,
+                 cond, body, failExpr, retType, loc)
+          in
+          let primPatt::PrimPattern =
```
krame505, line 260: This transformation on `Pattern`s might be more cleanly expressed as an attribute of type `(PrimPattern ::= Expr)` rather than a case expression bound in a let.

grammars/silver/compiler/extension/rewriting/Expr.sv
```diff
@@ -126,30 +126,34 @@ top::Expr ::= e::Decorated Expr es::Decorated AppExprs anns::Decorated AnnoAppEx
     | productionReference(q), _ -> prodCallASTExpr(q.lookupValue.fullName, es.transform, anns.transform)
 
     -- Special cases for efficiency
-    | classMemberReference(q), snocAppExprs(oneAppExprs(presentAppExpr(e1)), _, presentAppExpr(e2))
```
krame505, line 129: Was this change required by the changes to pattern matching semantics, or just a general cleanup?  According to my understanding of the new semantics, this code shouldn't be affected.  (I agree the new version is slightly neater anyway.)

I guess this could be even further improved:
```
| classMemberReference(q), snocAppExprs(oneAppExprs(presentAppExpr(e1)), _, presentAppExpr(e2)) ->
  case q.lookupValue.fullName of
  | "silver:core:eq" -> eqeqASTExpr(e1.transform, e2.transform)
  | "silver:core:neq" -> neqASTExpr(e1.transform, e2.transform)
  | ...
  | _ -> applyASTExpr(e.transform, es.transform, anns.transform)
  end
| classMemberReference(q), oneAppExprs(presentAppExpr(e)) ->
  case q.lookupValue.fullName of
  ...
  end
```

test/patt/Test.sv
```diff
@@ -60,7 +60,52 @@ equalityTest ( tNameAll(d()), "d", String, pat_tests ) ;
 
 
 
---- Part 2: GADTS
+--- Part 2: Match Order
```
krame505, line 63: Yay, more tests!

deep-rebuild
```diff
@@ -3,7 +3,7 @@
 set -eu
 
 export SILVER_HOME=$(pwd)
-JVM_ARGS="-Xss16M -Xmx4G -jar ../jars/silver.compiler.composed.Default.jar"
+JVM_ARGS="-Xss16M -Xmx6G -jar ../jars/silver.compiler.composed.Default.jar"
```
krame505, line 6: Did this change actually make Silver's memory usage that much worse?  

If so, is this an issue of the Silver compiler being less efficient when compiling pattern matching, or are downstream projects built from the new version of Silver (e.g. ableC) more memory hungry now too?

krame505 (commented):
Looks good overall, but a few considerations/questions before merging this.

grammars/silver/compiler/extension/patternmatching/Case.sv
RandomActsOfGrammar: I didn't write it; I just copied it from what was already there.  It could be turned into an attribute, but that might not belong in this pull request.

grammars/silver/compiler/extension/rewriting/Expr.sv
RandomActsOfGrammar: When I was trying to get it to compile, it was ending up hanging when trying to compile this grammar.  When I changed this, it was no longer hanging.

I just tried it again, and it still hangs with the original version.  I believe this is because the compilation of a `case` expression is now a lot larger, especially on this example.  Before it would combine all the patterns that were the same together into one match, with the `when`s as `if`s within there.  The new compilation doesn't do this, instead turning each pattern into a single match.

We could probably make it more efficient by combining the patterns if we have the same patterns in a row, like this example.  I would have to think about it more to make sure doing combinations like this wouldn't mess with the ordering.

deep-rebuild
RandomActsOfGrammar: This was left over from when I was trying to fix memory errors due to the bug you fixed with type unification for possibly-decorated nonterminals.  I am reverting this change.

krame505 (approved):
Looks good!  
... I'm going to trust that this works since it is passing all the tests, rather than trying to double-check all of the logic.

Issue #52, opened 2015-11-10
Locals with same name in aspect have duplicate equation errors


krame505:
If in grammar a you have a production with a local and inherited equation:

```
abstract production foo
e::Expr ::=
{
  local bar::Expr = ...;
  bar.env = emptyEnv();
}
```

and an aspect with the same local and equation in grammar b:

```
aspect production foo
e::Expr ::=
{
  local bar::Expr = ...;
  bar.env = emptyEnv();
}
```

you get a 'duplicate equation' warning on the line `bar.env = emptyEnv();` in both productions.  There should not be errors, as the locals are totally unrelated.

tedinski:
So... this is a general Silver environment bug. I'm not going to tackle it at the moment. Here's a quick diagnosis:

I wrote a stupid test case with production 'a' and two locals 'y' (in separate grammars.) This actually works, but look at the indexes generated for these:

    generated/src/bug/a/Init.java:public static final int y__ON__bug_a = bug.Init.count_local__ON__bug_a++;
    generated/src/bug/Init.java:public static final int y__ON__bug_a = bug.Init.count_local__ON__bug_a++;

Notice how they're exactly the same except that they're in the two different grammars (bug and bug:a). So, Silver is working essentially by accident here. They're correctly treated as two different locals in the runtime and everything because accesses for each go via the index from the correct grammar. But this immediately suggests a problem if we put the aspect in the same grammar:

    [javac] /home/tedinski/repos/silver/generated/src/bug/Init.java:63: error: variable y__ON__bug_a is already defined in class Init
    [javac] public static final int y__ON__bug_a = bug.Init.count_local__ON__bug_a++;

Silver is trying to use the same translated name for each of them and so boom, java compile error. No Silver error.

## Solution

Silver's environment uses `fName::String` routinely. These take 3 different forms currently:

 * full:grammar:path:shortname. This is standard
 * shortnameonly. This is for things like children, which are names never exported from a grammar and exist only in local environments. This is sort of okay. Probably.
 * full:grammar:path:productionshortname:local:shortname. Literally just the string :local: jammed in there between the name of the production and the name of the local. This is what locals are usually given as a name.

We probably ought to upgrade `fName`s from a `String` to a proper datatype in their own right. Then properly solve the problem of coming up with unique names for locals. Possibly by just genInt()ing an integer in there.

[...]

PR #602, opened 2021-12-16
More cleanup and misc. bug fixes

krame505:
# Changes
* Fix issues with name collisions between locals with the same name causing bogus flow errors and invalid translations.  Note that Ted suggested introducing a `FullName` type to use for full names instead of `String`s as a part of fixing this, in #52.  I'm not sure I understand how that would significantly improve the current state of affairs, honestly.  So I'm deferring that to a future refactor if needed.  
* Give production attributes their own host production instead of trying to make them forward into locals
* Cleanup to how we handle various error reporting flags in the MWDA
* Some minor cleanup in the implicit monads extension, use the existing `.isDecorated` attribute on types instead of declaring a new function that does the same thing.  

# Documentation
These are all internal code changes and cleanups that shouldn't change the compiler behaviour, aside from some bug fixes.  I added comments in a few places.

### DIFF
--- a/grammars/silver/compiler/analysis/typechecking/core/ProductionBody.sv
+++ b/grammars/silver/compiler/analysis/typechecking/core/ProductionBody.sv
@@ -156,6 +156,12 @@ top::ProductionStmt ::= 'local' 'attribute' a::Name '::' te::TypeExpr ';'
   top.errors <- te.errorsKindStar;
 }
 
+aspect production productionAttributeDcl
+top::ProductionStmt ::= 'production' 'attribute' a::Name '::' te::TypeExpr ';'
+{
+  top.errors <- te.errorsKindStar;
+}
+
 aspect production localValueDef
 top::ProductionStmt ::= val::Decorated QName  e::Expr
 {
--- a/grammars/silver/compiler/definition/core/ProductionBody.sv
+++ b/grammars/silver/compiler/definition/core/ProductionBody.sv
@@ -146,13 +146,13 @@ top::ProductionStmt ::= 'local' 'attribute' a::Name '::' te::TypeExpr ';'
   top.unparse = "\tlocal attribute " ++ a.unparse ++ "::" ++ te.unparse ++ ";";
 
   production attribute fName :: String;
-  fName = top.frame.fullName ++ ":local:" ++ a.name;
+  fName = s"${top.frame.fullName}:local:${top.grammarName}:${substitute(".", "_", top.location.filename)}:${toString(top.location.line)}:${toString(top.location.column)}:" ++ a.name;
 
   top.defs := [localDef(top.grammarName, a.location, fName, te.typerep)];
 
   top.errors <-
-        if length(getValueDclAll(fName, top.env)) > 1 
-        then [err(a.location, "Value '" ++ fName ++ "' is already bound.")]
+        if length(getValueDclInScope(a.name, top.env)) > 1 
+        then [err(a.location, "Value '" ++ a.name ++ "' is already bound.")]
         else [];
 
   top.errors <- if !top.frame.permitLocalAttributes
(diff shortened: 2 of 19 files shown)


grammars/silver/compiler/analysis/typechecking/core/ProductionBody.sv
```diff
@@ -156,6 +156,12 @@ top::ProductionStmt ::= 'local' 'attribute' a::Name '::' te::TypeExpr ';'
   top.errors <- te.errorsKindStar;
 }
 
+aspect production productionAttributeDcl
+top::ProductionStmt ::= 'production' 'attribute' a::Name '::' te::TypeExpr ';'
+{
+  top.errors <- te.errorsKindStar;
```
remexre, line 162: oof, was this just missing before?

grammars/silver/compiler/definition/core/ProductionBody.sv
```diff
@@ -146,13 +146,13 @@ top::ProductionStmt ::= 'local' 'attribute' a::Name '::' te::TypeExpr ';'
   top.unparse = "\tlocal attribute " ++ a.unparse ++ "::" ++ te.unparse ++ ";";
 
   production attribute fName :: String;
-  fName = top.frame.fullName ++ ":local:" ++ a.name;
+  fName = s"${top.frame.fullName}:local:${top.grammarName}:${implode("_", filter(isAlpha, explode(".", top.location.filename)))}:${toString(top.location.line)}:${toString(top.location.column)}:" ++ a.name;
```
remexre, line 149: feels like we should have some "real" mangling scheme instead of the `implode("_", filter(isAlpha, explode(".", ...)))` thing (or if this is a common pattern, it should be factored out to a function)

remexre: esp since the filter means there can still be collisions between files with different non-alphabetic chars, though that does seem like an really rare case since we mangle with the line+col too

grammars/silver/compiler/definition/core/ProductionBody.sv
```diff
@@ -163,17 +163,21 @@ top::ProductionStmt ::= 'local' 'attribute' a::Name '::' te::TypeExpr ';'
...
 top::ProductionStmt ::= 'production' 'attribute' a::Name '::' te::TypeExpr ';'
 {
-  -- TODO: we should unparse the production keyword, not local here!!
-  --top.unparse = "\tproduction attribute " ++ a.unparse ++ "::" ++ te.unparse ++ ";";
+  top.unparse = "\tproduction attribute " ++ a.unparse ++ "::" ++ te.unparse ++ ";";
 
-  top.productionAttributes := forward.defs;
-  top.defs := [];
+  production attribute fName :: String;
+  fName = top.frame.fullName ++ ":local:" ++ top.grammarName ++ ":" ++ a.name;
```
remexre, line 169: should this be `:production:` now?

remexre (approved):
hm, I kinda wish we could filter warnings at the top, but I can't think of a way to make that as efficient...

grammars/silver/compiler/analysis/typechecking/core/ProductionBody.sv
krame505: Yeah.

grammars/silver/compiler/definition/core/ProductionBody.sv
krame505: Yeah there could be collisions in some slightly contrived situations, but these full names also show up in error messages - thus they still need to be human readable and make logical sense.  So a better name mangling algorithm would require some thought.  Anyway this is much less broken in terms of collisions than how it previously worked!

grammars/silver/compiler/definition/core/ProductionBody.sv
krame505: Production attributes have always been `production:full:name:local:something`, since they are in the "local" scope of the production I guess?  IDK maybe this should change, but I'm leaving it for now.

krame505:
@remexre what is your comment about filtering warnings in reference to?

remexre:
Ah, just seeing all the branches around producing warnings; I feel like it'd be somewhat cleaner if we didn't need the checks when creating the errors, but could branch on them above. I can't see a reasonable way to do it though, at least not while having errors be collected in a normal cons list.

krame505:
Ideally we might have a seperate monoid attribute for every class of errors/warnings in which we are interested, and just grab the ones we want off the root of the tree.  But that probably doesn't save any real complexity in this case.

PR #604, opened 2021-12-23
Revamped random AST generating extension

krame505:
This branch is based on #602 and so that should be merged before reviewing this PR.  

# Changes
This is major cleanup (and mostly a rewrite) of Ted's random AST generation extension that was originally intended for use with property testing.  That had a number of shortcomings, including a reliance on specially-named functions and no easy ability to generate trees containing different sets of extension productions.  

This introduces new syntax for declaring an AST generator that is similar to parser specifications, e.g.
```
generator gen :: Root {
  my:host:lang;
  my:extensions:ext1;
  my:extensions:ext2;
}
```
to produce a function of type `(Root ::= Integer)`, parameterized by the desired depth of the AST to be generated.  The `Arbitrary` type class is used to specify how primitive types should be generated, or to customize the generation behavior of nonterminals such as names.  

# Documentation
This is currently a rather experimental feature, and may be subject to change.  Thus I don't think we should document this on the website, just yet.  

The source code contains comments as needed.


remexre:
> This is currently a rather experimental feature, and may be subject to change. Thus I don't think we should document this on the website, just yet.

/me coughs

grammars/silver/compiler/extension/treegen/Arbitrary.sv
```diff
@@ -1,29 +1,128 @@
...
+imports silver:compiler:metatranslation;
+
+terminal Generator_t 'generator' lexer classes {KEYWORD};
+
+concrete production generatorDcl
+top::AGDcl ::= 'generator' n::Name '::' t::TypeExpr '{' grammars::GeneratorComponents '}'
+{
+  top.unparse = s"generator ${n.unparse} :: ${t.unparse} { ${grammars.unparse} }";
+
+  -- Generator components must be imported for the translation here,
```
remexre, line 22: could the solution here be to have imports be an attr w/ a very small flowtype? well, I suppose forwarding might break that... would it make sense to have some notion of an attribute (or an occurrence?) that always needs an explicit equation, but then can have a flowtype smaller than the forward's?

grammars/silver/compiler/extension/treegen/Arbitrary.sv
```diff
@@ -1,29 +1,128 @@
...
+  -- but an AGDcl can't (currently) forward to an import ModuleStmt -
+  -- resorting to a slight interfering workaround for now.
+  propagate moduleNames;
+  forward.env = newScopeEnv(grammars.defs, top.env);
+
+  production specEnv::Decorated Env = newScopeEnv(grammars.defs, emptyEnv());
+  production specNTs::[TypeDclInfo] =
+    filter(\
```
remexre, line 30: not a huge fan of the backslash at the end of the line..... OTOH, not like we actually have a styleguide /shrug

grammars/silver/compiler/extension/treegen/Arbitrary.sv
```diff
@@ -1,29 +1,128 @@
...
+      map((.dcl), foldr(consDefs, nilDefs(), grammars.defs).typeList));
+
+  forwards to Silver_AGDcl {
+    function $Name{n}
+    $TypeExpr{t} ::= depth::Integer
+    {
+      $ProductionStmt{
+        foldr(
+          productionStmtAppend(_, _, location=top.location),
+          errorProductionStmt([], location=top.location), -- TODO: No nullProductionStmt?
```
remexre, line 41: any reason to map to a bunch of `productionStmtAppend`s instead of a `ProductionStmts`? (and I guess, what's the point of having both?)

grammars/silver/compiler/extension/treegen/Arbitrary.sv
```diff
@@ -1,29 +1,128 @@
...
+          map(genNtLocalDecl(top.location, top.env, specEnv, _), map((.fullName), specNTs)))}
+      return $Expr{genForType(top.location, top.env, specEnv, Silver_Expr { depth }, t.typerep)};
+    }
+  };
+
+  -- Uncomment for debugging
```
remexre, line 47: F

grammars/silver/compiler/extension/treegen/Arbitrary.sv
```diff
@@ -1,29 +1,128 @@
...
-        deriveArbitraryOn(_, top.env),
-        map((.qnwtQN), names.qnames)));
+  top.unparse = c1.unparse ++ c2.unparse;
 }
 
+concrete production generatorComponent
+top::GeneratorComponent ::= m::ModuleName ';'
+{
+  top.unparse = m.unparse ++ ";";
+  top.defs <- unsafeTracePrint([], hackUnparse(m) ++ ": " ++ hackUnparse(m.defs) ++ "\n\n\n");
```
remexre, line 72: is this supposed to be left in?

grammars/silver/compiler/extension/treegen/Arbitrary.sv
```diff
@@ -1,29 +1,128 @@
...
+}
 
+-- Generate the expression for constructing a type
+function genForType
+Expr ::= loc::Location  env::Decorated Env  specEnv::Decorated Env  depth::Expr  t::Type
+{
+  return
+    case t of
+    -- Monomorphic nonterminals that don't have an explicit Arbitrary instance,
```
remexre, line 81: Any advantage to this case over a `derive Arbitrary`?

grammars/silver/compiler/extension/treegen/Arbitrary.sv
```diff
@@ -45,110 +144,63 @@ function takeWhile2
...
-            mkStrFunctionInvocation(l, "genRand", []), '*',
-            floatConst(terminal(Float_t, toString(toFloat(num_lowest_arity)/toFloat(length(prods)))), location=l), location=l),
-          'else', mkStrFunctionInvocation(l, "genRand", []),
-          location=l),
-        ';', location=l),
-      returnDef('return', generateExprChain(0, prods, length(prods), l), ';', location=l)
-    ];
+  local result::Expr =
+    if null(prods)
+    then Silver_Expr { error($Expr{stringConst(terminal(String_t, "\"no generatable productions for nonterminal " ++ nt ++ "\""), location=loc)}) }
```
remexre, line 161: shouldn't this be a compile-time error?

the case where this would show up would be like

```
nt Foo;

prod foo
top::Foo ::= x::Foo
{}
```

right? where there's no finite generatable tree?

grammars/silver/compiler/extension/treegen/Arbitrary.sv
```diff
@@ -1,29 +1,128 @@
...
+    | appType(listCtrType(), elemT) ->
+      Silver_Expr {
+        map(
+          \ depth::Integer -> $Expr{genForType(loc, env, specEnv, Silver_Expr { depth - 1 }, elemT)},
+          take(toInteger(genRand() * toFloat($Expr{depth}))), reverse(range(0, $Expr{depth})))
+      }
+
+    -- Primitives and polymorphic nonterminals (e.g. Pair for tuples) are
+    -- handled by the Arbitrary type class.
+    | _ -> Silver_Expr { $Name{name("genArb", loc)}($Expr{depth}) }
```
remexre, line 98: is a way to override this on the todo list? for eg strings custom generators seem p important

also, terminals ought to be automatically generatable too, right?

grammars/silver/compiler/extension/treegen/TestFor.sv
```diff
@@ -1,6 +1,9 @@
 
 import silver:compiler:extension:testing;
 
+-- TODO: Resurrect this at some point
```
remexre, line 4: didn't read it, but is the existing code actually useful enough to be worth keeping?

grammars/silver/compiler/extension/treegen/Arbitrary.sv
```diff
@@ -45,110 +144,63 @@ function takeWhile2
...
 
 We should look up 'Expr' and get a list of productions, from that we get the probabilities.
 
-We then map deriveGenerateOn over the list of productions and generate a big if-then-else tree based on the genRand()
+We then map genForType over the list of productions and generate a big if-then-else tree based on the genRand()
 
 -}
-}
 
 function generateExprChain
```
remexre, line 188: add a comment documenting that this is non-total? IMO any non-total function **needs** a comment explaining when it crashes or loops.

grammars/silver/core/Arbitrary.sv
```diff
@@ -0,0 +1,39 @@
...
+  genArb :: (a ::= Integer);
+}
+
+instance Arbitrary Integer {
+  genArb = \ depth::Integer -> toInteger(genRand() * toFloat(depth));
+}
+
+instance Arbitrary Float {
+  genArb = \ depth::Integer -> genRand() * toFloat(depth);
+}
```
remexre, line 16: why do these use depth? imo, we should allow specifying non-arbitrary-instance generator functions, and the arbitrary instances should return random values amongst all values (incl the NaN space), possibly just generating 32 random bits and casting from them

grammars/silver/core/Arbitrary.sv
```diff
@@ -0,0 +1,39 @@
...
+
+instance Arbitrary Boolean {
+  genArb = \ depth::Integer -> genRand() > 0.5;
+}
+
+instance Arbitrary String {
+  genArb = \ depth::Integer ->
+    let chars::String = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789`~!@#$%^&*()_-+=\\|[]{}/?'\";:,.<>\n\n\n\t\r                                       "
```
remexre, line 24: uh

remexre: TBH, I'd almost be inclined to _not_ have an `Arbitrary String` instance, requiring a custom generator function be passed instead, but have regex literal syntax for specifying them, plus `map(.lexeme, arbitrary<SomeTerminal>())`

tutorials/simple/arb_with_all/out
```diff
@@ -0,0 +1,178 @@
+Compiling simple:composed:simple_arb_all
```
remexre: uhhhhhhhhhhhhhhhhhhhhh should this have been committed? notably contains `/home/lucas`...

tutorials/simple/src/simple/abstractsyntax/Expr.sv
```diff
@@ -19,35 +19,39 @@ n::Name ::= s::String
   n.lookup = lookupValue(s, n.env);
 }
 
+instance Arbitrary Name {
+  genArb = \ depth::Integer -> name(charsToString(randShuffle(stringToChars("abcd"))), location=genArb(depth));
```
remexre, line 23: hm, if this used origins, it wouldn't need the arbitrary location, right?

grammars/silver/compiler/extension/treegen/Arbitrary.sv
krame505: I don't quite understand what you are proposing here?  The non-interfering fix here would be to forward to "error production"-style productions to inject `moduleNames` and `defs`.  I think what you are suggesting would still be interfering, i.e. what if some transformation rewrote away this production?  But maybe I'm just not following.

grammars/silver/compiler/extension/treegen/Arbitrary.sv
krame505: I don't think there is a `$ProductionStmts` antiquote production, and I guess I didn't feel like adding one?  I think the point of `productionStmtAppend` is for `ProductionStmt` productions that want to forward to more than one `ProductionStmt`.  But that is an abstract prod that would be ambiguous for parsing.

grammars/silver/compiler/extension/treegen/Arbitrary.sv
krame505: The issue with doing this entirely via type classes is that the derived instance would always be orphaned, unless you derive it in the host lang, in which case extension productions would be missed.  I'm sure there is a way to handle polymorphism but I didn't have a need for it at the moment.

grammars/silver/compiler/extension/treegen/Arbitrary.sv
krame505: The idea is you can structure your AST e.g. with a `Name` nonterminal, and define an instance of `Arbitrary` for that.  

I guess with terminals the issue is the random lexeme must match the AST, but on further thought there is probably some nice way of doing that, maybe based on Brzozowski derivatives?  Would be a nice way of solving the generating-random-valid-names problem, you can just generate a random CST.

grammars/silver/compiler/extension/treegen/TestFor.sv
krame505: Possibly.  Just needs some tweaking to call a particular generator instead of the specially-named gen functions I think?

grammars/silver/core/Arbitrary.sv
krame505: I guess I was interpreting it to be some notion of "size", but that may make more sense?  I don't expect these instances to actually be used much if we take the "random CST to AST" approach, but probably still good to have for completeness.

grammars/silver/core/Arbitrary.sv
krame505: See above comment, probably want _some_ sort of instance for completeness here, but generally should be covered by generating random CSTs directly.

tutorials/simple/src/simple/abstractsyntax/Expr.sv
krame505: Yeah.  Arguably the origin info for a random AST is less useful than a location that includes the depth.

grammars/silver/compiler/extension/treegen/Arbitrary.sv
krame505: Oops, fixing.

grammars/silver/compiler/extension/treegen/Arbitrary.sv
krame505: In theory yes, but I don't want to implement a traversal through the grammar to determine what nts are reachable from the start nt right now.  I'll add a TODO comment.

grammars/silver/core/Arbitrary.sv
krame505: I think you are right that using depth here doesn't really make sense.  I'm not sure if covering the entire 32-bit range is always the most useful behaviour though?  I changed it to some smaller fixed constants for the moment and added a TODO comment.

tutorials/simple/arb_with_all/out
krame505: Oops.

remexre (approved):

grammars/silver/compiler/definition/concrete_syntax/TerminalDcl.sv
```diff
@@ -10,6 +10,7 @@ terminal Named_kwd       'named'       lexer classes {KEYWORD};
 terminal Left_kwd        'left'        lexer classes {KEYWORD};
 terminal Association_kwd 'association' lexer classes {KEYWORD};
 terminal Right_kwd       'right'       lexer classes {KEYWORD};
+terminal RepeatProb_kwd  'repeatProb'  lexer classes {KEYWORD};
```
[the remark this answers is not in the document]
krame505: `TerminalModifier` is a closed NT, so in principle extensions/modifications can add new ones (and we do in the copper "modification", with lexer classes and stuff that just gets added to the cstast.)  The trouble here is that this modifier introduces info that needs to go in the regular environment `TypeDclInfo`.  That could be made extensible with nonterminals for representing collections of "terminal env info items", but doing so probably adds more complexity than it would be worth.

grammars/silver/core/Arbitrary.sv
```diff
@@ -1,39 +1,50 @@
 grammar silver:core;
 
 @{--
-  - Represents primitive (terminal) types for which random values can be generated. 
+  - Represents primitive (terminal) types for which arbitrary random values can be generated.
   -}
 class Arbitrary a {
```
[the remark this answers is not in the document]
krame505: Interesting.  Not something on my critical path at the moment, but I will add a comment as something to look into later.

PR #605, opened 2021-12-28
Implement a RandomGen monad

krame505:
# Changes
I got sick of dealing with lots of impure generator functions for various types.  Also, implicit monads would be helpful in generating random regex match strings.  So, here's an actual monad for random generation.  Also allows for pure computations from a fixed seed, which is helpful for reproducible test cases.  

Also, fixed a bug in the `upload-override-jars` script, which wasn't properly encoding the printed Jenkins build URL.

# Documentation
I added doc comments for the new `Random` type class.

### DIFF
--- a/grammars/silver/core/Random.sv
+++ b/grammars/silver/core/Random.sv
@@ -20,7 +20,7 @@ instance Random Boolean {
 }
 
 @{-
-Types for which random values can be generated, uniformly distributed over some range.
+Types for which random values can be generated, uniformly distributed over some closed range [min, max].
 
 Note that this is not a subclass of Ord since we may have instances for partial orders.
 
@@ -33,11 +33,15 @@ class Random a => RandomRange a {
 }
 
 instance RandomRange Integer {
-  randomRange = \ min::Integer max::Integer -> map(\ i::Integer -> i % (max - min) + min, random);
+  randomRange = \ min::Integer max::Integer ->
+    if min > max then error(s"Empty Integer range [${toString(min)}, ${toString(max)}]")
+    else map(\ i::Integer -> i % (max - min + 1) + min, random);
 }
 
 instance RandomRange Float {
-  randomRange = \ min::Float max::Float -> map(\ f::Float -> f * (max - min) + min, random);
+  randomRange = \ min::Float max::Float ->
+    if min > max then error(s"Empty Float range [${toString(min)}, ${toString(max)}]")
+    else map(\ f::Float -> f * (max - min) + min, random);
 }
 
 instance RandomRange Boolean {
@@ -45,7 +49,7 @@ instance RandomRange Boolean {
     case min, max of
     | false, false -> pure(false)
     | false, true -> random
-    | true, false -> error("Invalid range")
+    | true, false -> error("Empty Boolean range")
     | true, true -> pure(true)
     end;
 }
--- a/grammars/silver/core/Random.sv
+++ b/grammars/silver/core/Random.sv
@@ -101,7 +101,7 @@ instance Monad RandomGen {}
 
 @{-
   Run a RandomGen computation, using an arbitrary seed.
-  Warning: this function is nondeterministic (may vary between runs) and thus impure -
+  Warning: this function is nondeterministic (may vary between runs) and thus impure;
   use at your own risk!
   
   @param r  The computation to run
@@ -117,7 +117,7 @@ a ::= r::RandomGen<a>
 
 @{-
   Run a RandomGen computation, using an arbitrary seed.
-  Warning: this function is nondeterministic (may vary between runs) and thus impure -
+  Warning: this function is nondeterministic (may vary between runs) and thus impure;
   use at your own risk!
 
   @param seed  The initial seed value for the random number generator
--- a/grammars/silver/core/Random.sv
+++ b/grammars/silver/core/Random.sv
@@ -2,19 +2,25 @@ grammar silver:core;
 
 @{-
 Types for which random values can be generated.
+
+This class imposes no restrictions on the range or distribution of random values,
+as there may be instances for types with no notion of ordering/equality.
 -}
 class Random a {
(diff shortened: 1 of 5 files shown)


grammars/silver/core/Random.sv
```diff
@@ -0,0 +1,138 @@
...
+  runRandomGen(randomRange(min, max)) < max
+-}
+class Random a => RandomRange a {
+  randomRange :: (RandomGen<a> ::= a a);
+}
+
+instance RandomRange Integer {
+  randomRange = \ min::Integer max::Integer ->
+    if min > max then error(s"Empty Integer range [${toString(min)}, ${toString(max)}]")
+    else map(\ i::Integer -> i % (max - min + 1) + min, random);
```
remexre, line 38: Is this still a uniform distribution?

I think I remembered the "uniform between 0 and n" algo being something like,
```
do {
  x = the low ceil(log2(n)) bits of uniformInt()
} while(x >= n)
return x
```

but that might have been for some special case, I don't remember.

runtime/java/src/common/RandomGen.java
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	public static Object runRandomGen(final OriginContext ctx, final NRandomGen r) {
+		return runRandomGen(ctx, new Random(), r);
+	}
+
+	// "Interpret" a RandomGen computation using the specified Random generator
+	@SuppressWarnings("unchecked")
+	public static Object runRandomGen(final OriginContext ctx, final Random rng, final NRandomGen r) {
+		if (r instanceof PmapRandomGen) {
```
remexre, line 20: should this check the forward?

grammars/silver/core/Random.sv
```diff
@@ -0,0 +1,138 @@
+grammar silver:core;
+
+@{-
+Types for which random values can be generated.
```
remexre, line 4: is it part of the typeclass that values are uniformly distributed?

grammars/silver/core/Random.sv
```diff
@@ -0,0 +1,138 @@
...
+-}
+class Random a {
+  random :: RandomGen<a>;
+}
+
+instance Random Integer {
+  random = randomInteger();
+}
+
+instance Random Float {
```
remexre, line 14: add a comment specifying whether NaNs and infinities can be generated?

remexre (approved):

grammars/silver/core/Random.sv
```diff
@@ -0,0 +1,138 @@
...
+}
+
+@{-
+Types for which random values can be generated, uniformly distributed over some closed range [min, max].
+
+Note that this is not a subclass of Ord since we may have instances for partial orders.
+
+If a has an instance for Ord, then instances should satisfy:
+  runRandomGen(randomRange(min, max)) >= min
+  runRandomGen(randomRange(min, max)) < max
```
remexre, line 29: isn't this open range?

grammars/silver/core/Random.sv
krame505: Well I'm starting with a uniform Float on [0.0, 1.0) and scaling it linearly, which should still be uniform I think?

remexre: You're doing it on an integer, not a float?

runtime/java/src/common/RandomGen.java
krame505: That would introduce a fair bit of overhead, since we would need to decorate the `RandomGen` into a tree... and I can't think of how that could possibly be useful, anyway.  So I'm just making it throw an error if we ever get a production that we didn't expect.  

... I guess this could be another argument in favor of having data types in Silver.  I'm still not convinced though :-P

grammars/silver/core/Random.sv
krame505: This class doesn't say anything about the range of the values, just that they are somehow random, so I guess I'm not sure how that could be a requirement for this type class.  I think I made it separate from `RandomRange` to cover types that have no concept of ordering.

grammars/silver/core/Random.sv
krame505: Oof, yeah I forgot to update that.

grammars/silver/core/Random.sv
remexre: hm, we should probably have doc comments on all the instances specifying the distribution and range then (unless the range is "all values" for all of them, in which case put that in the typeclass doc comment ig)

grammars/silver/core/Random.sv
krame505: Oh I wasn't paying attention and though you were commenting on the float instance, lol.  

Yeah I guess that makes sense, but Silver doesn't have shift/bitwise operators at the moment and I don't feel like adding them right now.  For small ranges (which is what I care about at the moment), modulo is probably faster and good enough.  I added a TODO that this might not actually be perfectly uniform, though.

remexre: Could you just make `randomRangeInteger` a primitive and define the `Random Integer` instance in terms of it?

grammars/silver/core/Random.sv
```diff
@@ -0,0 +1,149 @@
...
+Types for which random values can be generated.
+
+This class imposes no restrictions on the range or distribution of random values,
+as there may be instances for types with no notion of ordering/equality.
+-}
+class Random a {
+  random :: RandomGen<a>;
+}
+
+-- Uniform random integer on [0, INT_MAX]
```
remexre, line 13: Does this really not generate negative values?

krame505: Oh good catch, I guess I misread the Random.nextInt javadoc.

grammars/silver/core/Random.sv
krame505: Oh yeah, good idea.

Issue #526, opened 2021-05-20
Make Jenkins yell about whitespace issues


remexre:
e.g. using tabs in `.sv` (and `.sv.md`? just in silver blocks there?) files, not having trailing newlines in the same

ericvanwyk:
This seems reasonable but what is the motivation for doing this? Is it a problem now?

remexre:
Ended up finding a bunch of it in the initial docgen PR, seems like it'd save some time if something other than human review would find it, since it's trivially machine-checkable

ericvanwyk:
Yes but what is the problem? Do markdown not render well with these?

remexre:
well, my editor has tabs = 8 spaces, and louis' had tabs = 4... and the editor introduces things like `\t\t  foo.bar = baz;`, which is just ugly

trailing newline is just unix convention, which... sublime text, I think he uses? wasnt' respecting

ericvanwyk:
OK - that makes sense. Let's do this.

krame505:
Isn't there a nicer way now to set this up as a check on github?  I would prefer to not have Jenkins builds failing due to formatting issues, and it would be good to avoid giving Jenkins additional sorts of responsibilities if we think we may want to replace it at some point.

PR #606, opened 2022-01-04
Also parameterize random tree generation by a minimum depth

krame505:
# Changes
When generating random trees for a grammar such as
```
Root ::= Decls
Decls ::= Decl Decls
Decls ::=
Decl ::= 'function' '{' Stmts '}'
Stmts ::= Stmt Stmts
Stmts ::=
Stmt ::= ...
```
there is a 50/50 chance that we will immediately take the empty `Decls` production, and generate an empty tree.  The same is true for `Stmts` - thus effectively 75% of the generated trees contain no `Decl`s or only an empty function `Decl`.  However in most cases we probably don't care about these "boring" trees.  

This fixes the problem by parameterizing the generator functions with a minimum depth; until reaching this depth no arity-0 productions will be generated if there are arity > 0 productions available.  

# Documentation
This feature is not yet documented as it may still change significantly, however I need to merge these changes now for other reasons.

### DIFF
--- a/grammars/silver/compiler/extension/treegen/Arbitrary.sv
+++ b/grammars/silver/compiler/extension/treegen/Arbitrary.sv
@@ -49,7 +49,7 @@ top::AGDcl ::= 'generator' n::Name '::' t::TypeExpr '{' grammars::GeneratorCompo
 
   forwards to Silver_AGDcl {
     function $Name{n}
-    silver:core:RandomGen<$TypeExpr{t}> ::= depth::Integer
+    silver:core:RandomGen<$TypeExpr{t}> ::= minDepth::Integer maxDepth::Integer
     {
       $ProductionStmt{
         foldr(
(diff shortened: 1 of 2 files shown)


grammars/silver/compiler/extension/treegen/Arbitrary.sv
```diff
@@ -184,15 +189,23 @@ ProductionStmt ::= loc::Location  env::Decorated Env  specEnv::Decorated Env  nt
...
   
   local result::Expr =
     if null(prods)
     -- TODO: This could be a compile-time error, in theory
     then Silver_Expr { error($Expr{stringConst(terminal(String_t, "\"no generatable productions for nonterminal " ++ nt ++ "\""), location=loc)}) }
     else Silver_Expr {
       silver:core:bind(
-        if depth <= 0
+        if depth >= maxDepth
+	-- Exclude all but the lowest-arity productions
```
remexre, line 201: tabs vs spaces again... I should rly do https://github.com/melt-umn/silver/issues/526 one of these days lol... maybe as a "15-minute task" for the hackathon?

remexre:
would a different solution to this maybe be to generate depths from a normal distr instead of a uniform one?

remexre (approved):

krame505:
Do you mean like having the probability of generating a prod being a normal distribution w.r.t. the arity of the productions?  I don't think that would work, e.g. the `Decls` nonterminal has 1 arity-0 and 1 arity-2 production, so we would still just be choosing 50/50 between them.

PR #607, opened 2022-01-05
Fix bug with overly aggressive optimization in strategy attribute traversal combinators

krame505:
# Changes
Some strategy expression optimizations are only valid in contexts where we know on what nonterminal the strategy will be applied.  Apparently these were being applied on traversal operands, which are strategies applied on the children of productions, which may be a different nonterminal.

In the process I performed some additional cleanup of flowtypes in the strategy attribute extension (a bunch of things depend on frame that shouldn't...) and found (and fixed) another bug as a result - some nonterminal-dependent optimizations were still being performed inappropriately.  

# Documentation
This is an internal cleanup.  I added a few TODO comments where appropriate.

### DIFF
--- a/grammars/silver/compiler/extension/strategyattr/StrategyExpr.sv
+++ b/grammars/silver/compiler/extension/strategyattr/StrategyExpr.sv
@@ -80,14 +80,15 @@ partial strategy attribute prodStep =
       end));
 attribute prodStep occurs on MRuleList;
 
-strategy attribute simplify = innermost(genericStep <+ ntStep);
+strategy attribute genericSimplify = innermost(genericStep);
+strategy attribute ntSimplify = innermost(genericStep <+ ntStep);
 strategy attribute optimize =
-  (sequence(optimize, simplify) <+
+  (sequence(optimize, ntSimplify) <+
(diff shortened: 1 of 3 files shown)


remexre:
Is there a simple test case that fails because of the optimization being incorrectly applied but passes now?

krame505:
Yeah, I can add one.

krame505:
OK, test case added.  Should actually be finished now, assuming Jenkins passes.

grammars/silver/compiler/extension/strategyattr/StrategyExpr.sv
```diff
@@ -728,7 +747,14 @@ top::StrategyExpr ::= id::Name ty::TypeExpr ml::MRuleList
       assignExpr(id, '::', ty, '=', errorExpr([], location=top.location), location=top.location),
       caseExpr(
         [hackExprType(ty.typerep, location=top.location)],
-        ml.matchRuleList, false,
+	-- TODO: matchRuleList on MRuleList depends on frame for some reason.
```
remexre, line 750: tabs vs spaces

remexre (approved):

PR #608, opened 2022-01-11
Support for arbitrary nesting of patterns

### LINKED ISSUE #495: Pattern parsing for lists
The following should be a valid pattern
```
| (head::firsttail)::finaltail ->
```
The pattern has type `[[a]]` for some type `a`, where `head` is the head of the first list in the overall list, `firsttail` is the tail of the first list in the overall list, and `finaltail` is the tail of the overall list.  However, it gives the following syntax error (this is at the beginning of line 47):
```
ProofState.sv:47:18: error: Syntax error:
Error at line 47, column 18 in file ProofState.sv
         (parser state: 707; real character index: 1859):
  Expected a token of one of the following types:
   [silver:compiler:definition:core:BlockComments,
    '::' (silver:compiler:definition:core:ColonColon_t),
    ',' (silver:compiler:definition:core:Comma_t),
    silver:compiler:definition:core:Comments,
    silver:compiler:definition:core:WhiteSpace]
   Input currently matches:
   [silver:compiler:extension:templating:syntax:QuoteWater,
    silver:compiler:extension:templating:syntax:SingleLineQuoteWater]
```

This does not seem to be a problem for expressions, since the expression
```
(head::firsttail)::finaltail
```
parses fine as an expression instead of as a pattern.

ericvanwyk:
# Changes

This adds a single concrete production to support the arbitrary nesting of patterns and fixes #495.

# Documentation
No new documentation since this is a bug fix.

### DIFF
--- a/grammars/silver/compiler/extension/patternmatching/PatternTypes.sv
+++ b/grammars/silver/compiler/extension/patternmatching/PatternTypes.sv
@@ -200,6 +200,7 @@ top::Pattern ::=
 concrete production nestedPatterns
 top::Pattern ::= '(' p::Pattern ')'
 {
+  top.unparse = s"(${p.unparse})";
   forwards to p;
 }
 
--- a/grammars/silver/compiler/extension/patternmatching/PatternTypes.sv
+++ b/grammars/silver/compiler/extension/patternmatching/PatternTypes.sv
@@ -196,6 +196,13 @@ top::Pattern ::=
   top.patternTypeName = "";
 }
 
+-- arbitrary nesting of patterns
+concrete production nestedPatterns
+top::Pattern ::= '(' p::Pattern ')'
+{
+  forwards to p;
+}
+
 --------------------------------------------------------------------------------
 
 -- Below are the non-canonical patterns, i.e. those for other types
@@ -390,8 +397,6 @@ top::NamedPattern ::= qn::QName '=' p::Pattern
   top.patternVars = p.patternVars;
   top.namedPatternList = [pair(qn.lookupAttribute.fullName, p)];
 }
-
-
 
 --helper function for building patternLists from lists of patterns
 function buildPatternList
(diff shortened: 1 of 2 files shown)


grammars/silver/compiler/extension/patternmatching/PatternTypes.sv
```diff
@@ -196,6 +196,13 @@ top::Pattern ::=
   top.patternTypeName = "";
 }
 
+-- arbitrary nesting of patterns
+concrete production nestedPatterns
```
RandomActsOfGrammar, line 200: You should add an equation for `unparse` here so the pattern will appear correctly grouped if printed.

RandomActsOfGrammar (approved):
Once you add the equation for `unparse`, it will be fine.  I won't need to look at it again.

grammars/silver/compiler/extension/patternmatching/PatternTypes.sv
RandomActsOfGrammar: You might be able to do it by building the pattern by hand then taking its `unparse` attribute.  This would require importing the pattern matching grammar into the test suite to be able to use its productions; I don't think any other test cases do anything similar.  The testing framework and test suite isn't really set up to do things like this for Silver.

ericvanwyk: OK - I'll skip the test then. Thanks.

RandomActsOfGrammar: The errors for incomplete matches display patterns which aren't covered, not the patterns that are there.

krame505: Ah, in that case never mind.

ericvanwyk: I added a `wrongCode` test but it doesn't use `unparse` in a way that would show this.  I've pushed it, but it doesn't get to this issue.

krame505: This is a very minor thing to be trying to test.  I wouldn't worry about it.

PR #609, opened 2022-01-13
Implement partially-decorated references and decoration of a reference with additional inherited attributes

krame505:
# Changes
This adds support for decorating an already-decorated tree with additional inherited attributes.  Partially-decorated reference types have the same run-time representation as regular references, but with added checks in the flow analysis to ensure that they have _only_ the attributes specified in the reference type, and that at most one partial reference can be taken to any decoration site.  

Children and locals that are partially decorated references can be decoration sites (like nonterminal children/locals), to be supplied with additional inherited attributes.  This is implemented by mutating the provided tree to add the newly provided attribute equations.

Note that this does create one significant hole with respect to modular well-definedness.  It is an error to attempt to decorate a partially-decorated reference with the same attribute twice.  While I have added checks in the flow analysis ensuring we can only take a single partially decorated reference to any tree (and directly decorate it at most once), there is nothing to prevent e.g. wrapping it in a production that takes a partially decorated reference as a child, then decorating the resulting term twice.  Doing so would result in a runtime error that is not caught by the flow analysis.  

The solution to this is probably an additional analysis based on linear types or [uniqueness types](http://docs.idris-lang.org/en/latest/reference/uniqueness-types.html); this requires further thought, I will open a separate issue.  

# Documentation
Documentation of this experimental feature is in https://github.com/melt-umn/melt-website/pull/38, largely written by @remexre. 

Comments have been added in the source code as appropriate, particularly in the flow analysis.


remexre:
> This feature is not yet documented.

You know what sucks less than reading Copper's source?
Writing docs.

Can you check that this is correct?
A performance test that shows the performance difference might be worthwhile; maybe choosing to use `decExpr` vs `pdecExpr` based on an inh?

---

# Partial Decoration

**NOTE**: This feature is more experimental than average, and is subject to change!

## The Exponential Redecoration Problem and `decExpr`

Let's take as an example, compiling a simple language with integers, booleans, and overloaded operators.
This language has an overloaded `&`, so that:

```
> true & false // logical
false
> 12 & 5       // bitwise
4
```

One way to implement this operator might look something like:

```silver
production andExpr
Expr ::= lhs::Expr  rhs::Expr
{
  forwards to
    case lhs.type, rhs.type of
    | boolType(), boolType() -> call(global("boolAnd"), exprsCons(lhs, exprsCons(rhs, exprsNil())))
    |  intType(),  intType() -> call(global( "intAnd"), exprsCons(lhs, exprsCons(rhs, exprsNil())))
    | _, _ -> errorExpr(...)
    end;
}
```

However, this has a big performance problem hiding in it!
`lhs` and `rhs` must be undecorated terms in order to be passed to `exprsCons`, so after decoration inference, the code is equivalent to:

```silver
production andExpr
Expr ::= lhs::Expr  rhs::Expr
{
  forwards to
    case lhs.type, rhs.type of
    | boolType(), boolType() -> call(global("boolAnd"), exprsCons(new(lhs), exprsCons(new(rhs), exprsNil())))
    |  intType(),  intType() -> call(global( "intAnd"), exprsCons(new(lhs), exprsCons(new(rhs), exprsNil())))
    | _, _ -> errorExpr(...)
    end;
}
```

This code may re-type-check an expression a number of times exponential in the nestedness of the expression!
To see why, let's look at how the expression `1 & (2 & (3 & 4))` gets type-checked, and count how many times `4` gets type-checked.

- To type-check `1 & (2 & (3 & 4))`, we need to evaluate its forward and type-check that; to do that, we need to type-check `1` and `2 & (3 & 4)`.
  - Type-checking `1` is trivial.
  - To type-check `2 & (3 & 4)`, we need to evaluate its forward and type-check that; to do that, we need to type-check `2` and `3 & 4`.
    - Type-checking `2` is trivial.
    - To type-check `3 & 4`, we need to evaluate its forward and type-check that; to do that, we need to type-check `3` and `4`.
      - Type-checking `3` is trivial.
      - Type-checking `4` is trivial. **(Type-checked 1 time)**
      - `3 & 4` forwards to `intAnd(3, 4)`
      - To type-check `intAnd(3, 4)`, we need to type-check `3` and `4`.
        - Type-checking `3` is trivial.
        - Type-checking `4` is trivial. **(Type-checked 2 times)**
    - `2 & (3 & 4)` forwards to `intAnd(2, 3 & 4)`.
    - To type-check `intAnd(2, 3 & 4)`, we need to type-check `2` and `3 & 4`.
      - Type-checking `2` is trivial.
      - To type-check `3 & 4`, we need to evaluate its forward and type-check that; to do that, we need to type-check `3` and `4`.
        - Type-checking `3` is trivial.
        - Type-checking `4` is trivial. **(Type-checked 3 times)**
        - `3 & 4` forwards to `intAnd(3, 4)`
        - To type-check `intAnd(3, 4)`, we need to type-check `3` and `4`.
          - Type-checking `3` is trivial.
          - Type-checking `4` is trivial. **(Type-checked 4 times)**
  - `1 & (2 & (3 & 4))` forwards to `intAnd(1, (2 & (3 & 4)))`.
  - To type-check `intAnd(1, intAnd(2, intAnd(3, 4)))`, we need to typecheck `1` and `intAnd(2, intAnd(3, 4))`.
    - Type-checking `1` is trivial.
    - To type-check `2 & (3 & 4)`, we need to evaluate its forward and type-check that; to do that, we need to type-check `2` and `3 & 4`.
      - Type-checking `2` is trivial.
      - To type-check `3 & 4`, we need to evaluate its forward and type-check that; to do that, we need to type-check `3` and `4`.
        - Type-checking `3` is trivial.
        - Type-checking `4` is trivial. **(Type-checked 5 times)**
        - `3 & 4` forwards to `intAnd(3, 4)`
        - To type-check `intAnd(3, 4)`, we need to type-check `3` and `4`.
          - Type-checking `3` is trivial.
          - Type-checking `4` is trivial. **(Type-checked 6 times)**
      - `2 & (3 & 4)` forwards to `intAnd(2, 3 & 4)`.
      - To type-check `intAnd(2, 3 & 4)`, we need to type-check `2` and `3 & 4`.
        - Type-checking `2` is trivial.
        - To type-check `3 & 4`, we need to evaluate its forward and type-check that; to do that, we need to type-check `3` and `4`.
          - Type-checking `3` is trivial.
          - Type-checking `4` is trivial. **(Type-checked 7 times)**
          - `3 & 4` forwards to `intAnd(3, 4)`
          - To type-check `intAnd(3, 4)`, we need to type-check `3` and `4`.
            - Type-checking `3` is trivial.
            - Type-checking `4` is trivial. **(Type-checked 8 times)**

The typical solution to this, colloquially called "`decExpr` productions," is to have a production like:

```silver
production decExpr
Expr ::= child::Decorated Expr
{
  top.type = child.type;
  -- equations for any other host-language synthesized attributes too
  -- since child is a reference, it already has all its inherited attributes
}
```

Then, instead of writing `new` (or having it implicitly inserted for you), you write:

```
production andExpr
Expr ::= lhs::Expr  rhs::Expr
{
  forwards to
    case lhs.type, rhs.type of
    | boolType(), boolType() -> call(global("boolAnd"), exprsCons(decExpr(lhs), exprsCons(decExpr(rhs), exprsNil())))
    |  intType(),  intType() -> call(global( "intAnd"), exprsCons(decExpr(lhs), exprsCons(decExpr(rhs), exprsNil())))
    | _, _ -> errorExpr(...)
    end;
}
```

Now, when you type-check `1 & (2 & (3 & 4))`, `4` gets type-checked only once (using `{| child |}` as concrete syntax for `decExpr`):

- To type-check `1 & (2 & (3 & 4))`, we need to evaluate its forward and type-check that; to do that, we need to type-check `1` and `2 & (3 & 4)`.
  - Type-checking `1` is trivial.
  - To type-check `2 & (3 & 4)`, we need to evaluate its forward and type-check that; to do that, we need to type-check `2` and `3 & 4`.
    - Type-checking `2` is trivial.
    - To type-check `3 & 4`, we need to evaluate its forward and type-check that; to do that, we need to type-check `3` and `4`.
      - Type-checking `3` is trivial.
      - Type-checking `4` is trivial. **(Type-checked 1 time)**
      - `3 & 4` forwards to `intAnd({| 3 |}, {| 4 |})`
      - To type-check `intAnd({| 3 |}, {| 4 |})`, we need to type-check `{| 3 |}` and `{| 4 |}`.
        - Type-checking `{| 3 |}` can use the already-computed type, so it doesn't need to do the actual work!
        - Type-checking `{| 4 |}` can use the already-computed type, so it doesn't need to do the actual work!
    - `2 & (3 & 4)` forwards to `intAnd({| 2 |}, {| intAnd({| 3 |}, {| 4 |}) |})`
    - To type-check `intAnd({| 2 |}, {| intAnd({| 3 |}, {| 4 |}) |})`, we need to type-check `{| 2 |}` and `{| intAnd({| 3 |}, {| 4 |}) |}`
      - Type-checking `{| 2 |}` can use the already-computed type, so it doesn't need to do the actual work!
      - Type-checking `{| intAnd({| 3 |}, {| 4 |}) |}` can use the already-computed type, so it doesn't need to do the actual work!
  - `1 & (2 & (3 & 4))` forwards to `intAnd({| 1 |}, {| intAnd({| 2 |}, {| intAnd({| 3 |}, {| 4 |}) |}) |})`
  - To type-check `intAnd({| 2 |}, {| intAnd({| 3 |}, {| 4 |}) |})`, we need to type-check `{| 2 |}` and `{| intAnd({| 3 |}, {| 4 |}) |}`
    - Type-checking `{| 1 |}` can use the already-computed type, so it doesn't need to do the actual work!
    - Type-checking `{| intAnd({| 2 |}, {| intAnd({| 3 |}, {| 4 |}) |}) |}` can use the already-computed type, so it doesn't need to do the actual work!

## Problems of `decExpr` With Extensions

Let's say the compiler was structured so type-checking and most other semantic analyses are in the "core," but actual translations are implemented as translations.
(Or if that feels a bit far-fetched, pretend this is an extension that adds a new translation.)

To collect all the target language declarations, we have a threaded pair of attributes.
(This might be done instead of using a monoid attribute to allow for pure generation of fresh names.)

```silver
inherited attribute transDeclsIn::[Decl];
synthesized attributes transDecls::[Decl];
```

What would the aspect for `decExpr` look like?
It can't simply add additional attributes to the reference, so there's not really a better solution than:

```silver
aspect production decExpr
Expr ::= child::Decorated Expr
{
  production transDeclsChild::Decorated Expr with {decorate, transDeclsIn} = child;
  transDeclsChild.typeEnv = top.typeEnv;
  -- equations for any other host-language inherited attributes in transDecls' flowtype
  transDeclsChild.transDeclsIn = top.transDeclsIn;
  top.transDecls = transDeclsChild.transDecls;
}
```

However, this has the exponential redecoration problem again, for the `transDecls` attribute!
What we _want_ is some way to write:

```silver
aspect production decExpr
Expr ::= child::Decorated Expr
{
  child.transDeclsIn = top.transDeclsIn;
  top.transDecls = child.transDecls;
}
```

However, this isn't possible, since we don't know that `child` doesn't have existing attributes on it.
For example, another extension could add:

```silver
production fooExpr
Expr ::=
{
  local expr::Expr = litIntExpr(5);
  expr.transDeclsIn = [];
  forwards to
    if expr.someSynThatDependsOnTransDeclsIn
    then decExpr(expr)
    else errorExpr(...);
}
```

In this case, the `child.transDeclsIn` equation in `decExpr` can't work, since the process of computing `someSynThatDependsOnTransDeclsIn` might have put the `[]` somewhere that it might need to be removed.

*Put a footnote about "why not use flowtypes to remove attributes?" ?*

## Partial Decoration as a Solution

Partial Decoration makes it sound to write `child.transDeclsIn`, under some conditions.
It introduces a new variety of type, `PartiallyDecorated ... with ...`.

`PartiallyDecorated Expr` is a tree that *only* has the reference set of attributes, rather than having *at least* the reference set of attributes.
If you were to instead write:

*Check syntax here!*

```silver
aspect production decExpr
Expr ::= child::PartiallyDecorated Expr
{
  production transDeclsChild::PartiallyDecorated Expr with {decorate, transDeclsIn} = child;
  transDeclsChild.transDeclsIn = top.transDeclsIn;
  top.transDecls = transDeclsChild.transDecls;
}
```

This will reuse any thunks that are already present on `child`, like the original `decExpr` did.
However, `transDecls` now works!

## Caveats

Currently, partial decoration is implemented by *mutating* the tree!
This means that the following code is unsound:

```silver
production barExpr
Expr ::= child::PartiallyDecorated Expr
{
  production child1::PartiallyDecorated Expr with {decorate, barInh} = child;
  child1.barInh = true;

  production child2::PartiallyDecorated Expr with {decorate, barInh} = child;
  child2.barInh = true;

  top.barSyn = child1.barSyn || child2.barSyn;
}
```

In the future, some sort of substructural type system (e.g. linear types) or other use analysis may be implemented.

krame505:
Thanks for writing that Nathan!  Yeah, I think that looks mostly right. For the moment anyway, until I change my mind again :-P

One slight quibble: it isn't really true anymore that `PartiallyDecorated` only has the reference set, since someone else may have decorated it with more attributes.  Rather we are only allowed to provide its reference set at the decoration site where the reference was taken.

I just realized that I'm missing an MWDA check for autocopy attributes that introduce implicit equations that aren't in the reference set of the taken partially decorated ref.  Annoyingly, we don't have a way to turn off an autocopy equation that we don't want.  Another reason for getting rid of autocopy... fortunately this shouldn't really pose a problem as long as all auto-copied attributes are included in the reference set by the host language.  Autocopy shouldn't do anything for partially decorated children, anyway.  

Also I guess this PR is now technically new-jars, but it's pretty trivial to re-bootstrap.

krame505:
Ugh, actually there is no way to make this be modularly sound in the presence of autocopy, because we can have an autocopy attribute introduced on a nonterminal, and independently introduce a production on the same nonterminal that takes a partially decorated reference to a child, so there is no place where we can even know to flag the equation.  

It would maybe be possible to record in the presence of a partially decorated reference to a child in the prodelton, and not perform the copy in that case?  But this would also need to be reflected in the flow analysis.  I think for now we may just need to live with this being broken and prioritize removing autocopy as a project for the next hackathon.

grammars/silver/compiler/analysis/warnings/flow/OrphanedEquation.sv
```diff
@@ -127,6 +161,96 @@ top::ExprLHSExpr ::= attr::QNameAttrOccur
...
+    end;
+}
+aspect production lhsReference
+top::Expr ::= q::Decorated QName
+{
+  local finalTy::Type = performSubstitution(top.typerep, top.finalSubst);
+  top.errors <-
+    case finalTy of
+    | partiallyDecoratedType(_, _) when top.config.warnEqdef ->
+      [mwdaWrn(top.config, top.location, s"Cannot take a partially decorated reference of type ${prettyType(finalTy)} to the lhs tree.")]
```
krame505, line 226: Include the name of the variable in the error.

grammars/silver/compiler/definition/type/PrettyPrinting.sv
```diff
@@ -213,11 +213,17 @@ top::Type ::= t::Type i::Type
...
+{
+  top.typepp = s"PartiallyDecorated ${t.typepp} with ${i.typepp}";
+}
+
 aspect production ntOrDecType
 top::Type ::= nt::Type inhs::Type hidden::Type
 {
 -- Sometimes useful for debugging.
---  top.typepp = "Undecorable " ++ nt.typepp ++ " with  " ++ inhs.typepp ++ " specialized " ++ prettyTypeWith(hidden, []);
+--  top.typepp = "Undecorable " ++ nt.typepp ++ " with  " ++ inhs.typepp ++ " specialized " ++ prettyTypeWith(hidden, []));
```
krame505, line 226: Extra right paren

alexandra-hanson (approved):

PR #613, opened 2022-01-17
Fixing display of missing patterns for nested lists

RandomActsOfGrammar:
# Changes
In checking completeness of case expressions, we generate a pattern which is missing.  If the missing pattern is a non-empty list in a non-empty list, we need to correctly group the inner and outer lists.  For example, we want to see `(_::_)::_` rather than `_::_::_`.  This change makes it display as the former rather than the latter, where it was giving us the latter previously.

# Documentation
I added a comment explaining why we need the extra matching where I generate the missing pattern.  I also added a comment in the test file to try to explain why we don't need the formatting check for any other patterns.

This does not require any further documentation because it is a simple change which should only affect the code immediately around it and should not change anything users see, other than to make an incomplete pattern clear.


ericvanwyk:
Good write-up in the Documentation section - we should all be doing this.

krame505 (approved):

Issue #616, opened 2022-01-17
Implicit monad rewriting should happen on what the access production forwards to


krame505:
Currently the implicit monads extension aspectes the `access` production decorates `e` with `expectedMonad` and `monadicallyUsed`, using these attributes to override `mtyperep` and `monadRewritten`.  This is interfering, and also a barrier to fixing the flow errors using partially decorated references, since we only want to decorated `e` with these attributes once.  

Instead, the equations on `errorAccessHandler` should be updated to appropriately handle the cases in which we have encountered a type error during regular type checking, but a well-formed monadic rewrite does exist.

Issue #615, opened 2022-01-17
production ifThen in the implicit monads extension exceeds the allowed forward flow dependency set


krame505:
Similarly to #611, the `ifThen` production forward depends on `top.expectedMonad`.  I think it should be possible to generate the translation without knowing the expected monad, and only use that attribute for computing `merrors`.

RandomActsOfGrammar:
I put the wrong issue number in #619.

Issue #611, opened 2022-01-15
production mcaseExpr_c in the implicit monads extension exceeds the allowed forward flow dependency set


krame505:
In the process of fixing all the flow errors in Silver, there are a few in the implicit monads extension.  One is that the `mcaseExpr_c` forwards depending on `mDownSubst` and `expectedMonad`, which are not in the allow forwards dependency set for `Expr`.

I think it should be possible to generate the translation without knowing the expected monad or monadic typing context, and only use those for computing `merrors`.

RandomActsOfGrammar:
I put the wrong issue number in #619.

PR #617, opened 2022-01-17
Fix all remaining flow errors in the Silver compiler

krame505:
# Changes
This pull request is based on #609 and should be reviewed after that is merged.

Remaining TODOs before merging:

- [x] Fix flow errors in the implicit monads extension (#616, #615, #611 - assigned to @RandomActsOfGrammar)
- [x] Get #609 reviewed and merged
- [x] Add MWDA run in Silver's Jenkins pipeline

# Documentation
I added comments for why a few needed equations were added.  Otherwise the majority of the changes to comments consisted of deleting TODOs about flow errors that needed fixing.  

This pull request does not make any externally visible behavioral changes to Silver (besides fixing some possible corner cases that could cause a crash due to missing equations) and so no external documentation is needed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -80,6 +80,10 @@ melt.trynode('silver') {
     melt.archiveCommitArtifacts("jars/*.jar")
   }
 
+  stage("Modular Analyses") {
+    sh "./self-compile --clean --mwda"
+  }
+
   stage("Test") {
     // These test cases and tutorials are run as seperate tasks to allow for parallelism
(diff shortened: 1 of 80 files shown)


RandomActsOfGrammar (approved):
I find the change from `if cond then f1(args) else f2(args)` to `(if cond then f1 else f2)(args)` to be a strange choice.

I'm trusting you on some of these changes, but overall they seem reasonable.

krame505:
> I find the change from if cond then f1(args) else f2(args) to (if cond then f1 else f2)(args) to be a strange choice.

The reason for that is the restriction that we can only take a single partially decorated reference to any given decoration site.  I suppose a more sophisticated analysis could detect that the taking of a reference in the respective branches is mutually exclusive, but for the moment this is a flow error.

PR #618, opened 2022-01-17
Fix completeness checking for all whens

### LINKED ISSUE #546: Case completeness not checked when all clauses have whens
If every clause has a `when` condition, case completeness does not appear to trigger.  It should give a warning.

For example, this code
```
case 2 of                
| x when x == 3 -> true
end
```
should give a warning that the pattern matching is incomplete, but is accepted instead.

RandomActsOfGrammar:
# Changes
Fixes #546.  This checks if there are no patterns without `when` clauses but there are some match rules with `when` clauses. 
 In that case, it generates a warning for incomplete patterns.

# Documentation
I added a comment explaining the reason for the additional error check.  I also added a comment with the test to explain what it is checking.  This only affects the local code, so it doesn't need any further documentation elsewhere.  Its only effect visible to the user is in checking incomplete matching.

### DIFF
--- a/grammars/silver/compiler/extension/patternmatching/Case.sv
+++ b/grammars/silver/compiler/extension/patternmatching/Case.sv
@@ -544,9 +544,9 @@ function allConCaseCheckOverlapping
   We need the environment to look up any nonterminal matches and see
   if the nonterminal is closed or not.
 
-  We return nothing() if the patterns are complete, and just(plst) if
-  plst is an example of a missing pattern set (matching over multiple
-  values at once).
+  We return nothing() if there are some patterns and the patterns are
+  complete, and just(plst) if plst is an example of a missing pattern
+  set (matching over multiple values at once).
 -}
 function checkCompleteness
 Maybe<[Pattern]> ::= lst::[[Decorated Pattern]] env::Decorated Env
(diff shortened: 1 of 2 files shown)


krame505 (commented):
The comment on `checkCompleteness` should probably be updated - this function also returns `nothing()` if there are no (conditionless) patterns at all.

RandomActsOfGrammar:
I don't see anything to update for the `checkCompleteness` comment.  This new check is separate from that one.  When we have no conditionless rules, `checkCompleteness` returns `nothing()`, which corresponds to no error.  Do you have a specific suggestion of what should be added there?

krame505:
The comment says that `We return nothing() if the patterns are complete` - if there are no patterns, then they are clearly not complete.

RandomActsOfGrammar:
One could also argue that it is complete if there are no identifiable missing patterns, but I see your point and added the comment in the interest of clarity.

krame505:
I'm not sure how having no patterns whatsoever means that there are no identifiable missing patterns, since in that case any pattern would be a counterexample that isn't matched (unless the nonterminal being matched has no productions.)  So it still seems strange to me that `checkCompleteness` would return `nothing()` in that case, rather than choosing an arbitrary production as a counterexample.  

But it's probably fine as-is, I just found it slightly non-intuitive to handle the no-patterns case in this way.

krame505 (approved):

RandomActsOfGrammar:
What is the purpose of the `complete` argument to the `caseExpr` production?  Is it to prevent checking for completeness in certain cases, like the `strategyattr` extension?

krame505:
Yes, because rules in strategies are (almost always) incomplete matches, the translation inserts a default case that causes the strategy to fail.  We decorate a `caseExpr`  in order to perform normal error checking, and we still want to check for overlapping patterns, etc. however we don't want to error on incomplete matches.

PR #622, opened 2022-01-18
Bugfix/rewrite not access

### LINKED ISSUE #616: Implicit monad rewriting should happen on what the access production forwards to

RandomActsOfGrammar:
# Changes
Fixes #616.  This moves the monad rewriting out of the `access` production and into the productions to which it forwards.

# Documentation
This change only affects the productions in which it occurs and does not affect what users see, so it does not need any wider documentation.  It does not add any local comments because it is the clear local rewriting.

### DIFF
--- a/grammars/silver/compiler/extension/implicit_monads/Expr.sv
+++ b/grammars/silver/compiler/extension/implicit_monads/Expr.sv
@@ -456,8 +456,6 @@ aspect production access
 top::Expr ::= e::Expr '.' q::QNameAttrOccur
 {
   propagate mDownSubst, mUpSubst;
-  e.expectedMonad = top.expectedMonad;
-  top.merrors := forward.merrors;
 }
 
 aspect production errorAccessHandler
@@ -494,6 +492,9 @@ top::Expr ::= e::Decorated Expr  q::Decorated QNameAttrOccur
                                            " is neither")]
                  end;
 
+  --Why do we rewrite here, in an error production?  We can get here from the basic access
+  --   production based on normal typechecking failing even though our typechecking will
+  --   succeed, and we then need to be able to go back.
   local eUnDec::Expr =
         if ne.mtyperep.isDecorated
         then Silver_Expr{ silver:core:new($Expr {ne.monadRewritten}) }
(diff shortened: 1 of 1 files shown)


grammars/silver/compiler/extension/implicit_monads/Expr.sv
```diff
@@ -457,33 +457,53 @@ top::Expr ::= e::Expr '.' q::QNameAttrOccur
...
-                      | restrictedInhDcl(_, _, _) -> []
-                      | implicitSynDcl(_, _, _) -> []
-                      | implicitInhDcl(_, _, _) -> []
-                      | _ -> [err(top.location, "Attributes accessed in implicit equations must " ++
-                                                "be either implicit or restricted; " ++ q.unparse ++
-                                                " is neither")]
-                      end
-                 else [];
-  e.monadicallyUsed = false; --this needs to change when we decorate monadic trees
+  top.merrors := forward.merrors;
```
krame505, line 460: Equation isn't needed

grammars/silver/compiler/extension/implicit_monads/Expr.sv
```diff
@@ -457,33 +457,53 @@ top::Expr ::= e::Expr '.' q::QNameAttrOccur
 {
   propagate mDownSubst, mUpSubst;
   e.expectedMonad = top.expectedMonad;
```
krame505, line 459: Equation isn't needed

grammars/silver/compiler/extension/implicit_monads/Expr.sv
```diff
@@ -457,33 +457,53 @@ top::Expr ::= e::Expr '.' q::QNameAttrOccur
...
-                      | _ -> [err(top.location, "Attributes accessed in implicit equations must " ++
-                                                "be either implicit or restricted; " ++ q.unparse ++
-                                                " is neither")]
-                      end
-                 else [];
-  e.monadicallyUsed = false; --this needs to change when we decorate monadic trees
+  top.merrors := forward.merrors;
+}
+
+aspect production errorAccessHandler
```
krame505, line 463: I might add a comment explaining why we are implementing translation on the `errorAccessProduction`, since this might not be super intuitive.  I.e. that dispatching to this production happens based on the ordinary type checking, and that there may be type errors in expressions for implicit equations according to the ordinary typing rules, prior to the rewriting.

krame505 (changes requested):

RandomActsOfGrammar:
I removed the extraneous equations and added a comment explaining why we are rewriting on both error access productions.

krame505 (approved):

PR #624, opened 2022-01-21
Renaming IO types/functions - Part II

unironically:
# Changes
Removing unnecessary 'T' suffixes from functions in `IOToken` which do not conflict with functions in `IO`. Introduced new file `IOMisc` for miscellaneous IO functions which do not make use of the `IOToken` type. Introduced `touchFiles`, `deleteFiles`, `deleteDirFiles` to the `IO` type. Suffixed conflicting functions in `IOToken` with `T`.

# Documentation
No new documentation introduced. Any changes to existing documentation follow the refactoring described above.

### DIFF
--- a/grammars/silver/core/IOToken.sv
+++ b/grammars/silver/core/IOToken.sv
@@ -408,23 +408,6 @@ IOToken ::= files::[String] i::IOToken
 @@{- # IO Misc -}
 
 @{--
- - Die with the stated error message and a stack trace.  Note that Silver stacks
- - may be hard to read (it's a lazy language.)
- -
- - @warning Does not return! Does not do any cleanup!
- -
- - @param msg  The path to list the contents of.
- - @return  Does not return.
- -}
-function error
-a ::= msg::String
-{
-  return error("Not Yet Implemented: error"); -- lol
-} foreign {
-  "java" : return "common.Util.error(%msg%.toString())";
-}
-
-@{--
  - Create a bogus world-state token, for use with @link[unsafeTrace].
  -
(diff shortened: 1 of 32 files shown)


grammars/silver/core/IOToken.sv
```diff
@@ -320,7 +320,7 @@ IOVal<Boolean> ::= s::String i::IOToken
  - @param i  The "before" world-state token.
  - @return  true if all files are deleted successfully.  false otherwise.
  -}
-function deleteFilesT
+function deleteFiles
```
krame505, line 323: Wait, does this not have a corresponding monadic version? It probably should. I would leave the T suffix for this one.

grammars/silver/core/IOToken.sv
```diff
@@ -335,7 +335,7 @@ IOVal<Boolean> ::= s::[String] i::IOToken
  - @param i  The "before" world-state token.
  - @return  true if contents are deleted successfully.  false otherwise.
  -}
-function deleteDirFilesT
```
krame505, line 338: This should also have a monadic version, so I would leave the suffix.

grammars/silver/core/IOToken.sv
```diff
@@ -396,7 +396,7 @@ IOToken ::= file::String i::IOToken
  - @param i  The IOToken token.
  - @return The IOToken token. Errors are suppressed.
  -}
-function touchFilesT
+function touchFiles
```
krame505, line 399: Same as above

krame505 (approved):

PR #629, opened 2022-01-27
Expand and factor out random utilities into a separate library

krame505:
# Changes
This change moves most of the random utilities from silver:core to silver:util:random (excluding those that are build dependencies for the runtime, which consequently must stay in silver:core.)  Additionally more utilities are added, including a "token based" alternative to the `RandomGen` monad, analogous to the IO token.  This is advantageous for use with "threaded" attributes, e.g. if we want to forward based on a random choice.  

# Documentation
I have added doc comments for the new library features.  This change does not affect any core features of Silver.

### DIFF
--- a/grammars/silver/compiler/translation/java/core/ProductionDcl.sv
+++ b/grammars/silver/compiler/translation/java/core/ProductionDcl.sv
@@ -50,8 +50,8 @@ top::AGDcl ::= 'abstract' 'production' id::Name ns::ProductionSignature body::Pr
 
   local commaIfKids :: String = if length(namedSig.inputElements)!=0 then "," else "";
   local commaIfAnnos :: String = if length(namedSig.namedInputElements)!=0 then "," else "";
-  local commaIfKidsOrAnnos :: String = if length(namedSig.inputElements)!=0 || length(namedSig.namedInputElements)!=0 then "," else "";
-  local commaIfKidsAndAnnos :: String = if length(namedSig.inputElements)!=0 && length(namedSig.namedInputElements)!=0 then "," else "";
+  local commaIfContexts :: String = if length(namedSig.contexts)!=0 then "," else "";
+  local commaIfAny :: String = if length(namedSig.inputElements)!=0 || length(namedSig.namedInputElements)!=0 || length(namedSig.contexts)!=0 then "," else "";
 
   local contexts::Contexts = foldContexts(namedSig.contexts);
   contexts.boundVariables = namedSig.freeVariables;
@@ -89,7 +89,7 @@ ${namedSig.inhOccursIndexDecls}
 ${namedSig.childStatic}
     }
 
-    public ${className}(final NOriginInfo origin ${commaIfKidsOrAnnos} ${namedSig.javaSignature}) {
+    public ${className}(final NOriginInfo origin ${commaIfAny} ${namedSig.javaSignature}) {
         super(${if wantsTracking then "origin"++commaIfAnnos else ""}${implode(", ", map((.annoRefElem), namedSig.namedInputElements))});
 ${implode("", map(makeChildAssign, namedSig.inputElements))}
 ${contexts.contextInitTrans}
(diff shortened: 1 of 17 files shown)


grammars/silver/compiler/translation/java/core/ProductionDcl.sv
```diff
@@ -50,8 +50,8 @@ top::AGDcl ::= 'abstract' 'production' id::Name ns::ProductionSignature body::Pr
 
   local commaIfKids :: String = if length(namedSig.inputElements)!=0 then "," else "";
```
remexre, line 51: is... kids an abbreviation of children?

krame505: Yeah, I guess so.  Blame @602p

grammars/silver/compiler/translation/java/core/ProductionDcl.sv
```diff
@@ -89,7 +89,7 @@ ${namedSig.inhOccursIndexDecls}
 ${namedSig.childStatic}
     }
 
-    public ${className}(final NOriginInfo origin ${commaIfKidsOrAnnos} ${namedSig.javaSignature}) {
+    public ${className}(final NOriginInfo origin ${commaIfAny} ${namedSig.javaSignature}) {
```
remexre, line 92: would it possibly be cleaner to just put the origin in the javaSignature?

krame505: We also use `javaSignature` in functions which don't have an `origin` parameter.  Could maybe be cleaner, but I'm not refactoring the translation right now.

grammars/silver/util/random/RandomMonad.sv
```diff
@@ -0,0 +1,49 @@
...
+@{-
+  Run a RandomGen computation, using an arbitrary seed.
+  Warning: this function is nondeterministic (may vary between runs) and thus impure;
+  use at your own risk!
+  
+  @param r  The computation to run
+  @return  The result of the computation
+-}
+function runRandomGen
+a ::= r::RandomGen<a>
```
remexre, line 12: uh wait just noticing this now, but shouldn't this return IO<a>

krame505: See the doc comment - this function is impure like `genInt`.  Although maybe making it `IO` and forcing people to run it through `unsafeIO`/`unsafeTrace` would be reasonable.

remexre: yeah, that's what I meant; if it's for testing, i assume testing gives you an io token?

grammars/silver/core/Random.sv
```diff
@@ -1,68 +1,13 @@
...
-    end;
-}
-
--- Monad for computations involving random number generation
+-- Monad for computations involving random number generation.
+-- This must live in silver:core because the runtime depends on these productions.
+-- Associated utilities are defined in silver:util:random.
 nonterminal RandomGen<a>;
 
+-- Pure random "token" for use with threaded attributes, represented as just a seed value.
```
remexre, line 8: I assume these are safe to reuse, unlike IO tokens; is this worth mentioning?

remexre (approved):
Looks good overall. A couple general questions:

- Any reason not to define the monad as State<Token>?
- Should there be a pure interface to tokens, where you pass in a pair of integers as a seed and it converts them to a token?

krame505:
> Any reason not to define the monad as State<Token>?

Performance, I guess?  I haven't benchmarked it, but I suspect that interpreting a RandomGen "AST" is faster.  And we avoid creating lots of `java.util.Random` objects and using them only twice (which TBH is probably not a huge fraction of the number of objects we allocate, anyway...)

> Should there be a pure interface to tokens, where you pass in a pair of integers as a seed and it converts them to a token?

You can just do `runSeedRandomGen($seed, randomToken)`

remexre:
> You can just do `runSeedRandomGen($seed, randomToken)`

Well, not if that gets moved to `IO` for being impure, right? Though actually yeah, does that one need the impurity warning?

krame505:
Yeah, the `runSeedRandomGen` function isn't impure.  I'll fix the comment.

PR #631, opened 2022-02-01
Fix bug with lambda translation that could cause lazy arguments to be evaluated more than once

krame505:
# Changes
This fixes a bug I noticed with lambdas, after demanding an argument thunk we never update the original argument to record the result - potentially causing arguments that are referenced more than once to be evaluated more than once.  Probably not a huge performance issue most of the time since we would fairly soon reach thunks that already have been evaluated.

Also some minor, slightly related cleanup that fixes a few warnings about unchecked casts.  

# Documentation
None added, this is a minor cleanup/bug fix.

### DIFF
--- a/grammars/silver/compiler/modification/ffi/java/FunctionDcl.sv
+++ b/grammars/silver/compiler/modification/ffi/java/FunctionDcl.sv
@@ -31,7 +31,9 @@ top::FFIDef ::= name::String_t ':' 'return' code::String_t ';'
 function strictChildAccessor
 String ::= ns::NamedSignatureElement
 {
-  return "((" ++ ns.typerep.transType ++ ")common.Util.demand(" ++ ns.childRefElem ++ "))";
+  -- TODO: Slight bug here, if we refer to an argument more than once in an FFI string,
+  -- then we may evaluate argument thunks more than once!
+  return "common.Util.<" ++ ns.typerep.transType ++ ">demand(" ++ ns.childRefElem ++ ")";
 }
 
 function computeSigTranslation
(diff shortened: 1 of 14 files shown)


grammars/silver/compiler/translation/java/core/NamedSignature.sv
```diff
@@ -239,7 +239,9 @@ top::NamedSignatureElement ::= n::String ty::Type
   top.childDeclElem =
 s"""private Object child_${n};
   public final ${ty.transType} getChild_${n}() {
-    return (${ty.transType}) (child_${n} = common.Util.demand(child_${n}));
+    final ${ty.transType} result = common.Util.<${ty.transType}>demand(child_${n});
```
remexre, line 242: Isn't this "two levels of caching"?

remexre (approved):
not super confident on my understanding of this; is there an overview of how we translate functions written up somewhere?

grammars/silver/compiler/modification/ffi/java/FunctionDcl.sv
```diff
@@ -31,7 +31,9 @@ top::FFIDef ::= name::String_t ':' 'return' code::String_t ';'
 function strictChildAccessor
 String ::= ns::NamedSignatureElement
 {
-  return "((" ++ ns.typerep.transType ++ ")common.Util.demand(" ++ ns.childRefElem ++ "))";
+  -- TODO: Slight bug here, if we refer to an argument more than once in an FFI string,
+  -- then we may evaluate argument thunks more than once!
```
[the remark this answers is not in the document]
krame505: Yes

grammars/silver/compiler/translation/java/core/NamedSignature.sv
krame505: Uhhh not sure I understand what you mean by that?  

I just restructured the code here to avoid an unnecessary unchecked cast.

PR #639, opened 2022-02-10
Adds println, eprint, eprintln, and switches to use them in silver:compiler.

### LINKED ISSUE #633: Use stderr
Currently, there's no way to write to stderr besides opening `/dev/stderr` as a file.

Diagnostics (e.g. `Found foo:bar`) and errors should be written there, so they don't break piping stdout.

This is probably pretty easy; add `eprint`, `eprintln`, audit all the locations where Silver calls `print` (almost all of them should be `eprint`, I reckon).

I suppose technically this is a breaking change for scripts that parse Silver's output; they should change from `./self-compile | sed -e evil -e regexen | perl -pE fun` to `./self-compile |& sed -e evil -e regexen | perl -pE fun`.

remexre:
# Changes

Resolves #633; see it for a description.

# Documentation

Added doc comments.

Arguably https://github.com/melt-umn/melt-website/blob/62b8ae7062175c30fd524540e18d90e8436cdab8/content/silver/lib/io.md and https://github.com/melt-umn/melt-website/blob/62b8ae7062175c30fd524540e18d90e8436cdab8/content/silver/concepts/io.md should be updated as well?

The former could probably go away (and become a redirect to the generated docs?).

The latter could TBH use a rewrite:

- I think the Troubles section is obsolete, and there's other misc remarks that are no longer true (e.g. us not having a unit type? why would someone write that in the docs instead of just defining one?).
- > When we fix IO all this will be thrown out anyway.

  Sad...
- I think the pre-monad Haskell comment might be incorrect too; "Tackling the Awkward Squad" describes a very different model (one that maybe could've been more convenient in Silver than it was in Haskell? Almost definitely more awkward than tokens are though). *Current* GHC Haskell actually does this as a way to implement the IO monad: https://hackage.haskell.org/package/ghc-prim-0.8.0/docs/src/GHC-Types.html#line-233

I don't super-duper want to though lol

### DIFF
--- a/grammars/silver/compiler/analysis/warnings/exporting/Graph.sv
+++ b/grammars/silver/compiler/analysis/warnings/exporting/Graph.sv
@@ -37,7 +37,7 @@ abstract production dumpDepGraphAction
 top::DriverAction ::= specs::[Decorated RootSpec]
 {
   top.io = writeFileT("deps.dot", "digraph deps {\n" ++ generateDotGraph(specs) ++ "}",
-    printT("Generating import graph\n", top.ioIn));
+    eprintlnT("Generating import graph", top.ioIn));
 
   top.code = 0;
   top.order = 0;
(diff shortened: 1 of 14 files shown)


krame505:
Is there anything still blocking this PR?  Would be nice to have Silver diagnostic output on stderr.

remexre:
uh, blocked by me not having the energy to install nailgun; can look again tmrw early afternoon prolly

krame505:
Remind me what the issue with nailgun was - the ableC tests are passing on Jenkins?

remexre:
failing on jenkins in a way that wasn't obvious from the console, so wanted to run them locally

PR #640, opened 2022-02-16
Allow loading a grammar from only interface files when source grammar does not exist

krame505:
# Changes
Currently Silver ignores the interface file for a grammar if the sources for the grammar cannot be found.  We may wish to support binary library distribution that does not include source files; this is a step in that direction.

# Documentation
I added comments in the source code.  Users should not notice this change, except that importing a deleted grammar might not raise an error in a non-clean build.


remexre (approved):

ericvanwyk:
Does this also requires that the appropriate `.java` or `.class` files are there as well?

krame505:
Yes, it assumes that the generated code is being linked against generated code corresponding to the existing interface files, in some way.  This isn't currently relevant for the way Silver is typically installed where everything is dumped in a single generated directory, but is useful in integrating Silver with build tools like Bazel or Nix.

PR #641, opened 2022-02-19
Support building multiple unrelated grammars in a single invocation of silver

krame505:
# Changes
Allows multiple unrelated Silver grammars to be built (and included in a jar) in a single invocation of Silver, without the need for a dummy grammar that imports all of them.  This is helpful in integrating Silver with build tools such as Bazel.  This is also faster than just invoking Silver on every grammar to be built, as we avoid the need to start another JVM and re-parse interface files.

# Documentation
The command-line help usage string has been updated.

### DIFF
--- a/grammars/silver/compiler/composed/idetest/Analyze.sv
+++ b/grammars/silver/compiler/composed/idetest/Analyze.sv
@@ -26,7 +26,10 @@ IOVal<[Message]> ::= args::[String]  svParser::SVParser ioin::IOToken
   local envErrors :: [String] = case benvResult.iovalue of | right(s) -> s | _ -> [] end;
   
   -- Let's start preparing to build
-  local buildGrammar :: String = head(a.buildGrammars);
+  local buildGrammar :: String =
+    if length(a.buildGrammars) == 1
+    then head(a.buildGrammars)
+    else error("Exactly one build grammar expected in argument passed to ideAnalyze");
   local checkbuild :: IOVal<[String]> =
     checkPreBuild(benv, [buildGrammar], benvResult.io);
 
--- a/grammars/silver/compiler/composed/idetest/Analyze.sv
+++ b/grammars/silver/compiler/composed/idetest/Analyze.sv
@@ -70,7 +70,10 @@ IOVal<[Message]> ::= args::[String]  svParser::SVParser  ioin::IOToken
   local envErrors :: [String] = case benvResult.iovalue of | right(s) -> s | _ -> [] end;
   
   -- Let's start preparing to build
-  local buildGrammar :: String = head(a.buildGrammars);
+  local buildGrammar :: String =
+    if length(a.buildGrammars) == 1
+    then head(a.buildGrammars)
+    else error("Exactly one build grammar expected in argument passed to ideGenerate");
   local checkbuild :: IOVal<[String]> =
     checkPreBuild(benv, [buildGrammar], benvResult.io);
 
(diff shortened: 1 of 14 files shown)


grammars/silver/compiler/composed/idetest/Analyze.sv
```diff
@@ -26,13 +26,13 @@ IOVal<[Message]> ::= args::[String]  svParser::SVParser ioin::IOToken
   local envErrors :: [String] = case benvResult.iovalue of | right(s) -> s | _ -> [] end;
   
   -- Let's start preparing to build
-  local buildGrammar :: String = head(a.buildGrammar);
+  local buildGrammar :: String = head(a.buildGrammars);
```
remexre, line 29: Why are this (and the later) `head(a.buildGrammars)` correct? Isn't this throwing away other grammars now?

grammars/silver/compiler/modification/impide/BuildProcess.sv
```diff
@@ -13,10 +13,10 @@ import silver:util:cmdargs;
 --}
 
 aspect production compilation
-top::Compilation ::= g::Grammars  _  buildGrammar::String  benv::BuildEnv
+top::Compilation ::= g::Grammars  _  buildGrammars::[String]  benv::BuildEnv
 {
   -- The RootSpec representing the grammar actually being built (specified on the command line)
-  local builtGrammar :: [Decorated RootSpec] = searchEnvTree(buildGrammar, g.compiledGrammars);
+  local builtGrammar :: [Decorated RootSpec] = searchEnvTree(head(buildGrammars), g.compiledGrammars);
```
remexre, line 19: same `head` question here

grammars/silver/compiler/driver/util/FlowTypes.sv
```diff
@@ -9,7 +9,7 @@ import silver:util:graph as g;
 -- Hide all the flow type computation over here
 
 aspect production compilation
-top::Compilation ::= g::Grammars  r::Grammars  buildGrammar::String  benv::BuildEnv
+top::Compilation ::= g::Grammars  r::Grammars  buildGrammars::[String]  benv::BuildEnv
```
remexre, line 12: is there a reason why this one didn't get `_`'d like the others? (don't need to change it, just making sure I'm not missing smth)

grammars/silver/compiler/composed/idetest/Analyze.sv
krame505: Yeah, this is just for the eclipse IDE plugin though, which at the moment only currently supports specifying one build grammar in the plugin settings anyway.

remexre: should it emit an error when there's more than one, then?

grammars/silver/compiler/driver/util/FlowTypes.sv
krame505: I think because the regex I used for renaming in the first pass missed signatures containing _, and then the second pass of renaming I did replaced it with _ instead of `buildGrammars::[String]`

grammars/silver/compiler/composed/idetest/Analyze.sv
krame505: Ok, done

grammars/silver/compiler/modification/impide/BuildProcess.sv
krame505: Fixed

remexre (approved):

PR #645, opened 2022-03-01
Fix random tree generation to consider lexical precedence

krame505:
# Changes
This PR is based on #644 and should be reviewed after merging that.

Currently using Silver's random tree generation feature to generate concrete syntax trees has a bug: a terminal may have any lexeme generated that matches its regex, even if it is dominated by other identifiers; this may lead to the generation of unparsable ASTs.  This change considers lexical precedence relationships to filter out candidate lexemes that would lead to parse errors.

# Documentation
I added source comments as needed.  This fixes a bug in an experimental feature of Silver.

### DIFF
--- a/grammars/silver/compiler/extension/treegen/Arbitrary.sv
+++ b/grammars/silver/compiler/extension/treegen/Arbitrary.sv
@@ -29,8 +29,10 @@ top::AGDcl ::= 'generator' n::Name '::' t::TypeExpr '{' grammars::GeneratorCompo
       grammars.moduleNames, []);
   production specEnv::Decorated Env = newScopeEnv(med.defs, emptyEnv());
   
-  -- Override defs to suppress production attributes from flowing up,
-  -- to avoid a circularity with the environment.
+  -- Override defs to suppress production attributes from flowing up as a paDef,
+  -- to avoid a circularity as what production attributes are generated depends
+  -- on the environment. Practically this means that we can't aspect a generator
+  -- function and refer to its production attributes, which is probably good anyway.
   top.defs :=
     case forward of
     | functionDcl(_, _, ns, _) -> [funDef(top.grammarName, n.location, ns.namedSignature)]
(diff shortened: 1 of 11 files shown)


grammars/silver/compiler/extension/treegen/Arbitrary.sv
```diff
@@ -12,43 +13,55 @@ imports silver:compiler:modification:lambda_fn;
...
+  local med::ModuleExportedDefs =
+    moduleExportedDefs(
+      top.location, top.compiledGrammars, top.grammarDependencies,
+      grammars.moduleNames, []);
+  production specEnv::Decorated Env = newScopeEnv(med.defs, emptyEnv());
+  
+  -- Override defs to suppress production attributes from flowing up,
+  -- to avoid a circularity with the environment.
+  top.defs :=
+    case forward of
```
remexre, line 35: I don't quite grok how this works; if we're demanding the forward anyway, shouldn't there still be a circularity when pulling any attributes off of it? Or is this because namedSignature has a smaller flowtype than the forward flowtype? (Does it?)

krame505: The `functionDcl` production emits `paDef`s from production attributes in the function body.  However the function body is generated depending on what terminals/nonterminals/productions exist inside the specified grammars.  If the generator is declared inside one of its own specified grammars, then we have a circularity of `top.env -> top.defs -> forward.defs -> forward -> top.env`.  The solution to this is to override `defs` and avoid making it depend on the production attributes in the function body - this sort of issue and hacky solution is a consequence of shoving a bunch of independent namespaces inside a single environment defs attribute.

krame505: Note that pattern matching on the `functionDcl` production in the forward to pull out the signature does not cause the body to be demanded.

remexre: Isn't this throwing away the `localDef`s then?

krame505: Yeah, to the extent that production attribute `localDef`s don't bubble up as a `paDef`s, so you couldn't reference them if you aspected the function defined by a generator declaration.

grammars/silver/compiler/extension/treegen/Arbitrary.sv
```diff
@@ -220,15 +233,30 @@ ProductionStmt ::= loc::Location  env::Decorated Env  specEnv::Decorated Env  nt
 }
 
 function genTermLocalDecl
-ProductionStmt ::= loc::Location  env::Decorated Env  specEnv::Decorated Env  t::String
+ProductionStmt ::= loc::Location  env::Decorated Env  specEnv::Decorated Env  dominatingTerminals::EnvTree<Decorated SyntaxDcl> t::String
 {
   local te::TypeExpr = nominalTypeExpr(qName(loc, t).qNameType, location=loc);
+
+  -- Filter out cantidate lexemes by checking if they match dominating terminal regexes.
+  -- TODO: This a approach is somewhat infinite, and breaks if a terminal is totally dominated by another.
```
remexre, line 241: somewhat infinite == it might never give back a terminal? Is the bug here that for a grammar like:

```
terminal A 'x';
terminal B 'x' dominates A;
nonterminal C
concrete production d
C ::= B A
{}
```

even though contextual scanning means we can parse this, we'll never generate it? (If so, add an example, perhaps this one?)

krame505: Yeah.  That comment was supposed to say "somewhat inefficient, and breaks with infinite recursion if ..."

remexre (approved):

grammars/silver/compiler/extension/treegen/Arbitrary.sv
krame505: Actually for what it's worth contextual scanning doesn't apply in the above grammar, `B dominates A` means we disallow `A` from ever matching anything that matches `B`'s regex.  So the infinite recursion failure can only happen for terminals that are impossible to ever lex anyway.

remexre: Oh, right, contextual scanning would just mean that if I *didn't* write the `dominates` then it'd still be non-ambiguous, and be parsable. Okay, that's a much less bad failure case.

Issue #601, opened 2021-12-14
Annotation "record" update syntax


krame505:
As we somewhat recently discovered, Silver has decently good support for records via annotations (e.g. see [the refactor to `FlowEnv`](https://github.com/melt-umn/silver/pull/585/files#diff-653bee41354f2cc99d2ff0eb7f9e9c381b10c4d9bc31648dfc2ac24310fd69bf)).  The one thing we lack is a corresponding notion of a "record update" that constructs a ~~record~~ nonterminal with different values of some ~~fields~~ annotations and copies others unchanged - for example we might have a language with an environment structure:
```
import silver:util:treemap as map;

annotation valueItems::map:Map<String ValueItem>;
annotation typeItems::map:Map<String TypeItem>;
nonterminal Env with valueItems, typeItems;
production env top::Env ::= {}

global emptyEnv::Env = env(valueItems=map:empty(), typeItems=map:empty());

function addValueDefs
Env ::= defs::[(String, ValueItem)]  e::Env
{
  return env(
    valueItems=map:add(defs, e.valueItems),
    typeItems=e.typeItems);
}

function addTypeDefs
Env ::= defs::[(String, TypeItem)]  e::Env
{
  return env(
    valueItems=e.valueItems,
    typeItems=map:add(defs, e.typeItems));
}
```
Here we need to specify that we are copying the value of `typeItems` in `addValueDefs`, and `valueItems` in `addTypeDefs`.  If we added another namespace, we would need to modify every site where we construct an `env`.  Conversely in Haskell one could just write  `e {valueItems=add defs (valueItems e)}` without mentioning the fields that are being copied.  

I guess we could just copy Haskell's syntax and permit `e {valueItems=map:add(defs, e.valueItems}}`?  For consistency we would also maybe want to support `env {valueItems=..., typeItems=...}` for constructing productions that have annotations but no children.  

Note that this proposed syntax adds even more confusion to the situation as described in #243 even worse, with regard to another inconsistent use of curly braces in Silver.  But my recollection is that the one thing we all agreed on in that issue was that we should change the decorate syntax to be `decorate foo with {attr1=bar, attr2=baz}` , but that never actually happened.  So maybe this syntax is reasonable if we make that change, too.  I could be persuaded otherwise.

krame505:
Doing some major refactoring in ableC reiterated that single-production `data` declarations with annotations work pretty well as records in Silver:
```
annotation returnType :: Maybe<Type>;
[...]

PR #652, opened 2022-03-06
Records

remexre:
# Changes

For now just adds [record declarations](https://github.com/melt-umn/silver/blob/a0d2eca0dce39d6b006e8281928de99d2ac2c10a/test/silver_features/Records.sv).

Before un-marking-as-draft will add record updates.

Also has a few minor API additions and a script tweak.

# Documentation

- [ ] Will make PR to update language reference

### DIFF
--- a/grammars/silver/compiler/extension/records/Records.sv.md
+++ b/grammars/silver/compiler/extension/records/Records.sv.md
@@ -1,0 +1,370 @@
+# Records
+
+```silver imports
+grammar silver:compiler:extension:records;
+
+import silver:compiler:definition:core;
+import silver:compiler:definition:env;
+import silver:compiler:definition:type;
+import silver:compiler:definition:type:syntax;
+```
+
+The records extension is meant to codify a common pattern of defining product types that generates efficient code and has nice usability.
+
+## Declaration
+
+A record definition starts with the `record` keyword.
+
+```silver
+terminal RecordKwd_t 'record' lexer classes {KEYWORD};
+```
+
+A record definition looks basically like a nonterminal declaration followed by a list of fields.
+This forwards to a nonterminal, a single production with no children, and an annotation that occurs on the nonterminal for each field.
+
+```silver
+concrete production recordDecl
+this::AGDcl ::= quals::NTDeclQualifiers  'record'  name::IdUpper_t
+  modifiers::NonterminalModifiers  '{'  fields::RecordFields0  '}'  ';'
+{
+  this.unparse =
+    s"${quals.unparse}record ${name.lexeme} ${modifiers.unparse} {${fields.unparse}};";
+
+  local ntDcl::AGDcl =
+    nonterminalDcl(quals, 'nonterminal',
+      nameIdUpper(name, location=this.location),
+      botlNone(location=this.location), modifiers, ';', location=this.location);
+  local ntType::TypeExpr =
+    nominalTypeExpr(qNameTypeId(name, location=this.location),
+      location=this.location);
+
+  -- PARTIAL: head() and tail() are safe because names are nonempty.
+  local nameChars::[String] = explode("", name.lexeme);
+  local prodName::IdLower_t = terminal(IdLower_t,
+    implode("", toLower(head(nameChars)) :: tail(nameChars)),
+    name.location);
+
+  local prodLHS::ProductionLHS = productionLHS(
+    nameIdLower(terminal(IdLower_t, "this"), location=this.location), '::',
+    ntType, location=this.location);
+  local prodSig::ProductionSignature =
+    productionSignatureNoCL(prodLHS, '::=',
+      productionRHSNil(location=this.location), location=this.location);
+  local prodDcl::AGDcl =
+    productionDcl('abstract', 'production',
+      nameIdLower(prodName, location=this.location), prodSig, 
+      productionBody('{', productionStmtsNil(location=this.location), '}',
+        location=this.location),
+      location=this.location);
+
+  forwards to appendAGDcl(ntDcl,
+    appendAGDcl(prodDcl, fields.recordFieldAGDcl, location=this.location),
+    location=this.location);
+}
+```
+
+### `without declaring` Nonterminal Modifier
+
+Note that to support multiple records with the same field names, we don't always declare a new annotation for the field.
+The `without declaring` nonterminal modifier requests that an annotation not be declared.
+It creates an entry in the `withoutDeclaring` synthesized attribute for the annotation.
+
+```silver
+synthesized attribute withoutDeclaring::[String] occurs on NonterminalModifier,
+  NonterminalModifiers, NonterminalModifierList;
+
+terminal Without_kwd   'without'   lexer classes {KEYWORD};
+terminal Declaring_kwd 'declaring' lexer classes {KEYWORD};
+
+concrete production withoutDeclaringOne
+this::NonterminalModifier ::= 'without'  'declaring'  fieldName::IdLower_t
+{
+  propagate nonterminalModifiers;
+
+  this.errors := [];
+  this.unparse = s"without declaring ${fieldName.lexeme}";
+
+  this.withoutDeclaring = [fieldName.lexeme];
+}
+
+concrete production withoutDeclaringMany
+this::NonterminalModifier ::= 'without'  'declaring'  '{' 
+  fieldNames::FieldNames0  '}'
+{
+  propagate nonterminalModifiers;
+
+  this.errors := [];
+  this.unparse = s"without declaring {${fieldNames.unparse}}";
+
+  this.withoutDeclaring = fieldNames.names;
+}
+
+aspect default production
+this::NonterminalModifier ::=
+{ this.withoutDeclaring = []; }
+```
+
+We need the propagates for `nonterminalModifiers` because it's interfering...
+
+```silver imports
+import silver:compiler:definition:concrete_syntax only nonterminalModifiers;
+```
+
+We want to use our own `FieldNames0` nonterminal, because none of the built-in ones seem to be "just a plain list of `IdLower_t`s."
+In particular, `QNames` is a list of type-qualified `QName`s, not just plain names.
+
+```silver
+synthesized attribute names::[String];
+
+nonterminal FieldNames0 with location, unparse, names;
+nonterminal FieldNames1 with location, unparse, names;
+
+concrete production fieldNames0Empty
+this::FieldNames0 ::=
+{
+  this.unparse = "";
+  this.names = [];
+}
+
+concrete production fieldNames0Nonempty
+this::FieldNames0 ::= fieldNames::FieldNames1
+{
+  this.unparse = fieldNames.unparse;
+  this.names = fieldNames.names;
+}
+
+concrete production fieldNames0TrailingComma
+this::FieldNames0 ::= fieldNames::FieldNames1  ','
+{ forwards to fieldNames0Nonempty(fieldNames, location=this.location); }
+
+concrete production fieldNames1One
+this::FieldNames1 ::= name::IdLower_t
+{
+  this.unparse = name.lexeme;
+  this.names = [name.lexeme];
+}
+
+concrete production fieldNames1Cons
+this::FieldNames1 ::= init::FieldNames1  ','  last::IdLower_t
+{
+  this.unparse = s"${init.unparse}, ${last.lexeme}";
+  this.names = init.names ++ [last.lexeme];
+}
+```
+
+To make `withoutDeclaring` flow up through the nonterminal modifier lists, we need to add aspects for those nonterminals.
+
+```silver
+aspect production nonterminalModifiersNone
+this::NonterminalModifiers ::=
+{ this.withoutDeclaring = []; }
+
+aspect production nonterminalModifierSome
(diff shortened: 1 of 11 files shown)


ericvanwyk:
Is this a "proper" extension to Silver? Does it all forward to nonterminals, productions, and annotations?

remexre:
Yes, with the exception of the record creation/update syntax (which is not yet added to this PR, and having a weird infinite loop I haven't had time to debug) being non-MDA-clean, but we're not enforcing MDA for Silver anyway I think?

That syntax being `prodName { field1 = expr1, field2 = expr2 }` to create, `recordValue { field2 = expr2 }` to update.

ericvanwyk:
We are moving towards enforcing MDA and MWDA for Silver. So new work should not make that more difficult.

Is it a good idea to implicitly introduce the production name by lower-casing the type name?  Why not use the record type name for creation?

krame505:
Most of the existing Silver "extensions" would require syntax changes to pass the MDA - so I wouldn't compromise on syntax just to make this a pure syntactic extension.  At some point we should probably sort out the Silver extensions into pure ones that pass the MDA and MWDA, and "semantic" extensions that only pass the MWDA.  

I am also confused by that record syntax, though - isn't `prodName` a `QName` that would be ambiguous with `recordValue`?  When would we ever have more than one production for a record?

remexre:
Oh, that could be the type name; I was going off of

> For consistency we would also maybe want to support env {valueItems=..., typeItems=...} for constructing productions that have annotations but no children.

from #601 ; the type name might make more sense though.

In the not-yet-working one that does a prod name, it parses as `prodOrRecord::Expr  '{'  args::AnnoAppArgs  '}'` and forwards based on the type of `prodOrRecord`:
- if a `::=`-type, fwd to `$prodOrRecord($args)`
- if a record-like NT, fwd to `$recordProd($args, $otherArg1=$prodOrRecord.$otherArg1, etc)`
- else fwd to errorExpr

krame505:
I'll try to take a look tomorrow.

grammars/silver/compiler/extension/records/Records.sv.md
```diff
@@ -0,0 +1,370 @@
...
+        location=this.location),
+      location=this.location);
+
+  forwards to appendAGDcl(ntDcl,
+    appendAGDcl(prodDcl, fields.recordFieldAGDcl, location=this.location),
+    location=this.location);
+}
+```
+
+### `without declaring` Nonterminal Modifier
```
krame505, line 66: Uhhhh I really don't like this syntax.  I would strongly prefer to just look at fixing the env to allow us to avoid duplicating annotation declarations.

grammars/silver/compiler/extension/records/Records.sv.md
```diff
@@ -0,0 +1,370 @@
...
+
+```silver
+synthesized attribute withoutDeclaring::[String] occurs on NonterminalModifier,
+  NonterminalModifiers, NonterminalModifierList;
+
+terminal Without_kwd   'without'   lexer classes {KEYWORD};
+terminal Declaring_kwd 'declaring' lexer classes {KEYWORD};
+
+concrete production withoutDeclaringOne
+this::NonterminalModifier ::= 'without'  'declaring'  fieldName::IdLower_t
```
krame505, line 80: Why does this reuse NonterminalModifier? I don't think there are any other nonterminal modifiers that we would want to support on records?

grammars/silver/compiler/extension/records/Records.sv.md
```diff
@@ -0,0 +1,370 @@
...
+
+  this.withoutDeclaring = fieldNames.names;
+}
+
+aspect default production
+this::NonterminalModifier ::=
+{ this.withoutDeclaring = []; }
+```
+
+We need the propagates for `nonterminalModifiers` because it's interfering...
```
krame505, line 107: How is this interfering?  `NonterminalModifier` is a closed nonterminal.

grammars/silver/compiler/extension/records/Records.sv.md
```diff
@@ -0,0 +1,370 @@
...
+import silver:compiler:definition:concrete_syntax only nonterminalModifiers;
+```
+
+We want to use our own `FieldNames0` nonterminal, because none of the built-in ones seem to be "just a plain list of `IdLower_t`s."
+In particular, `QNames` is a list of type-qualified `QName`s, not just plain names.
+
+```silver
+synthesized attribute names::[String];
+
+nonterminal FieldNames0 with location, unparse, names;
```
krame505, line 119: Why is this factored into two seperate nonterminals?

grammars/silver/compiler/extension/records/Records.sv.md
```diff
@@ -0,0 +1,370 @@
...
+abstract production bar
+Bar ::=
+{}
+
+annotation field1<a>::a;
+annotation field1<Boolean> occurs on Bar;
+annotation field2<Float> occurs on Bar;
+```
+
+Type inference should always be able to sort out what the type of `thing.field1` is correctly, although we don't currently have a way to talk about "types with a `field1` with type `String`."
```
krame505, line 233: You should just be able to write `annotation field1<String> occurs on a` as a type constraint.  Could make an extension forwarding to that too, e.g. `record a has field field1::String`.  Although I really would like to see some more examples of how this would be useful...

grammars/silver/compiler/extension/records/Records.sv.md
```diff
@@ -0,0 +1,370 @@
...
+There can be zero fields, a single field, or a series of fields; fields are comma-separated, and a trailing comma is allowed.
+The grammar is:
+
+```
+RecordFields ::=
+              |  RecordField
+              |  RecordField ',' RecordFields
+```
+
+However, for parsing efficiency, we want to make this grammar left-recursive instead of right-recursive.
```
krame505, line 305: Is this actually a measurable performance impact?  Also I think this grammar allows a trailing comma, was that intended?

self-compile
```diff
@@ -11,6 +11,7 @@ SV_BUILD_TARGET=${SV_BUILD_TARGET:-"silver:compiler:composed:Default"}
 
 export ANT_OPTS=-Xss10M
 
+cd "$(realpath "$(dirname "${BASH_SOURCE[0]}")")"
```
krame505, line 14: What does this do?

krame505 (commented):
For the record (no pun intended) I wasn't proposing to add a separate record declaration syntax in #601, although I'm not opposed to it if it can be done cleanly - but the weirdness with redeclaring annotations certainly doesn't help.

grammars/silver/compiler/extension/records/Records.sv.md
```diff
@@ -0,0 +1,370 @@
...
+```silver
+terminal RecordKwd_t 'record' lexer classes {KEYWORD};
+```
+
+A record definition looks basically like a nonterminal declaration followed by a list of fields.
+This forwards to a nonterminal, a single production with no children, and an annotation that occurs on the nonterminal for each field.
+
+```silver
+concrete production recordDecl
+this::AGDcl ::= quals::NTDeclQualifiers  'record'  name::IdUpper_t
```
krame505, line 27: I would consider it a style guideline in the Silver compiler itself to use `top` for the LHS name, just to be consistent.

krame505:
Why the restriction on record-update syntax only working for "record like" nonterminals that have a single childless production?  It would make sense for record updates to work with any nonterminal type, with the semantics of "copy this term and update the values of the specified annotations".  

This wouldn't be doable as an extension, but if we wanted "record update" syntax to work with any nonterminal then it probably belongs as a separate modification or part of the host language anyway.  Shouldn't be too hard to implement, we would probably want to generate a helper function in every production class taking all the annotations as parameters, that simply copies the term with the new annotation values.

grammars/silver/compiler/extension/records/Records.sv.md
remexre: To have a left-recursive grammar with an optional trailing comma; I can add a comment about this?

grammars/silver/compiler/extension/records/Records.sv.md
remexre: > performance impact

From this extension alone, no, but my understanding was that in general, you want to make as many things as possible left-recursive because it matters for parsing perf.

> trailing comma

Yeah, that was intentional; I typically write records with trailing commas in other languages because then I don't need to worry about it when changing or reordering fields, and diffs don't have the extra "noise" of:

```diff
 record Foo {
     bar: String,
-    baz: String
+    baz: String,
+    quux: String
 }
```

self-compile
remexre: Oh, just realized the `realpath` is superfluous; but it's so I can run `../../self-compile` from e.g. `test/silver_features`

grammars/silver/compiler/extension/records/Records.sv.md
remexre: Ah, I'll fix that.

grammars/silver/compiler/extension/records/Records.sv.md
remexre: Shouldn't `silver:compiler:definition:concrete_syntax` define a default equation for it then? Or is it being a monoid attribute supposed to provide one? It doesn't look like it does; removing the propagates here and running the test suite does result in a crash from a missing equation.

I suppose I should've just added a default equation rather than a workaround here...

remexre:
> Why the restriction on record-update syntax only working for "record like" nonterminals that have a single childless production?

Because I didn't want to do the env plumbing to make it only work on ones declared with `record` :P 

IMO if we want a more general "replace the annotations on this term" syntax, that requires more thinking. And, especially, a use-case -- is the thinking that this might be used to update locations or something, using rewriting+reflection to find all the location-having terms? That feels specialized enough that we don't particularly need syntax for it.

grammars/silver/compiler/extension/records/Records.sv.md
krame505: Ah yes, makes sense.

grammars/silver/compiler/extension/records/Records.sv.md
krame505: `silver:compiler:definition:concrete_syntax` is an optional export of `silver:compiler:definition:core`, so `noterminalModifiers` is effectively a part of the "host language" defining the `NonterminalModifier` nonterminal - thus it doesn't need a default equation, and must be defined for all `NonterminalModifier` extension productions.

FYI if an extension defines a new attribute on a closed host nonterminal without a default equation, then the problem is categorized as a modular well-definedness issue along the same lines as an extension production lacking a forward, not an interference issue.

krame505:
I could imagine wanting to define an algebraic data type with multiple constructors, that also has some record fields shared by all constructors.  In Haskell the "data type" portion would need to be a seperate data type corresponding to a field in the record, but in Silver that maps nicely to a nonterminal with annotations and multiple productions.  I also think that there are more ways in which we can effectively use annotations in abstract syntax (e.g. type qualifiers as an annotation on types in ableC), where something like this could be useful.  

I don't really want people to think of nonterminals with annotations and records as separate features in Silver - I would prefer to keep "record update" syntax as a general-purpose feature for manipulating terms, and just have record declarations as a lightweight extension for declaring a nonterminal with multiple annotations and a single empty production.

grammars/silver/compiler/extension/records/Records.sv.md
krame505: As discussed, this should probably not be a NonterminalModifier - we don't want to permit this on regular nonterminal declarations.  And I don't see a good reason to declare any other kind of NonterminalModifier on a record nonterminal.

remexre:
If this could be done as a pure extension I'd agree, but unless we have a concrete use-case, I'd be kinda reluctant to make this a modification?

> In Haskell [...]

FWIW, Haskell lets you write:

```
data Foo
  = Bar { a :: Integer, b :: String }
  | Baz { a :: Integer, c :: Bool }

fooA :: Foo -> Integer
fooA = a
```

AFAIK all the optics libraries also generate a Lens for `a` as well.

> algebraic data type
> I don't really want people to think of nonterminals with annotations and records as separate features in Silver

/me mumbles about datatypes vs terms

grammars/silver/compiler/extension/records/Records.sv.md
remexre: Yeah, when I fix the env I'll just drop this section.

remexre:
I was proposing keeping it for record-like nonterminals only (and then as a pure extension) until we have a piece of Silver code where someone was like "darn, I really wish I had non-record-like annotation updates." Off the top of my head, I'm not remembering one, but I could of course be very easily missing one or ten.

> we can't know about all possible productions 

*datatype mumbling intensifies*

ericvanwyk:
So what is the fix that alleviates the mumbling? Is this record extension just a band-aid on a deeper problem? What is the idealized view of AGs that supports this more directly?

The need for "datatypes" is a legit one. We should articulate why that is. How does it relate to extensibility?  What about performance?  What about ergonomics of using Silver / AGs in this way?

krame505:
After a chat with @remexre, closing this in favor of #777.

ericvanwyk:
This sounds good to me.

PR #654, opened 2022-03-24
Fix for spawning a subprocess with arguments

RandomActsOfGrammar:
# Changes
This fixes the `spawnProcess` function in `silver:util:subprocess` so it doesn't crash when arguments are given for the subprocess being spawned.

# Documentation
I added a comment explaining why the `for` loop is there.  This doesn't need any further documentation because it is fixing the behavior to be the expected one and doesn't affect any code outside of the immediate function.

### DIFF
--- a/runtime/java/src/common/rawlib/RawProcessHandle.java
+++ b/runtime/java/src/common/rawlib/RawProcessHandle.java
@@ -207,6 +207,7 @@ public class RawProcessHandle {
     public static NIOVal spawnProcess(StringCatter cmd, ConsCell args, IOToken io) {
         List<StringCatter> argsList = ConsCell.toList(args);
         List<String> full_cmd = new ArrayList<String>();
+        //Convert all the args to regular Strings instead of StringCatter
         for (int i = 0; i < argsList.size(); i++) {
             full_cmd.add(i, argsList.get(i).toString());
         }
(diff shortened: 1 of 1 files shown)


runtime/java/src/common/rawlib/RawProcessHandle.java
```diff
@@ -204,7 +205,12 @@ private void waitOnEnd() {
      * <pre>IOVal<ProcessHandle> ::= cmd::String args::[String] io::IO</pre>
      */
     public static NIOVal spawnProcess(StringCatter cmd, ConsCell args, IOToken io) {
-        List<String> full_cmd = ConsCell.toList(args);
+        List<StringCatter> argsList = ConsCell.toList(args);
```
krame505, line 208: `common.javainterop.ConsCellCollection` provides a nicer way to iterate over the elements of a Silver list without converting to an intermediate `ArrayList`:
```java
for (StringCatter arg : new ConsCellCollection<StringCatter>(args)) {
  full_cmd.add(arg.toString);
}
```

remexre: `full_cmd.addAll(new ConsCellCollection<StringCatter>(args));` too

krame505: You also need to map `.toString()` over the args

remexre: whoops yeah, missed that.

RandomActsOfGrammar: I changed it to use Lucas's suggestion now.

krame505 (approved):

PR #657, opened 2022-04-25
Fix bug with strategy attribute translation let binding not being decorated

krame505:
# Changes
Fixes a minor bug in strategy attribute translation: when the production LHS name and the name given in a rule don't match, we introduce a let binding.  This should be a decorated reference to the actual LHS but instead was undecorated, leading to missing inherited equations when the copied LHS gets re-decorated with nothing.

# Documentation
None, this is a bug fix for an already-documented feature.

### DIFF
--- a/grammars/silver/compiler/extension/strategyattr/StrategyExpr.sv
+++ b/grammars/silver/compiler/extension/strategyattr/StrategyExpr.sv
@@ -796,7 +796,7 @@ top::StrategyExpr ::= id::Name ty::TypeExpr ml::MRuleList
     else if top.frame.signature.outputElement.elementName == id.name
     then res
     else Silver_Expr {
-      let $Name{id}::$TypeExpr{ty} = $name{top.frame.signature.outputElement.elementName}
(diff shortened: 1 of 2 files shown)


grammars/silver/compiler/extension/strategyattr/StrategyExpr.sv
```diff
@@ -796,7 +796,7 @@ top::StrategyExpr ::= id::Name ty::TypeExpr ml::MRuleList
     else if top.frame.signature.outputElement.elementName == id.name
     then res
     else Silver_Expr {
-      let $Name{id}::$TypeExpr{ty} = $name{top.frame.signature.outputElement.elementName}
+      let $Name{id}::Decorated $TypeExpr{ty} = $name{top.frame.signature.outputElement.elementName}
```
remexre, line 799: Didn't this originally tyck b/c of decoration inference :P

krame505: I guess so?  Either behavior may be useful in different situations.  At least I haven't changed anything with how `Decorated` works that affects this.

hypnoticOxbow (approved):

PR #658, opened 2022-04-26
Allow specifying the direction that children are traversed in threaded attributes

krame505:
# Changes
This allows specifying the default order in which children are traversed in propagating a threaded attribute pair, by allowing `direction = left to right` or `direction = right to left` after the type in the declaration.  The default if no direction is specified remains left to right.

# Documentation
I updated the docs about threaded attributes on the website.


ericvanwyk (approved):
Looks good - was less complicated than I expected.

PR #659, opened 2022-05-11
Add a utility for selecting a random element from a list

krame505:
# Changes
Added a utility function for selecting a random element from a list.

# Documentation
I added a doc comment for the new function.

### DIFF
--- a/grammars/silver/util/random/Util.sv
+++ b/grammars/silver/util/random/Util.sv
@@ -20,6 +20,21 @@ RandomGen<[a]> ::= elems::[a]
     };
 }
 
+@{--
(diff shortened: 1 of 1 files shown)


remexre (commented):
rip, do we not have an `nth :: (a ::= [a] Int)`? LGTM if not; though maybe worth noting that it's partial on an empty list? (Maybe I'll add `@partial` for doc comments at some point?)

remexre (approved):

krame505:
Yeah, doesn't look like we have `nth` but I improved the error message and added a comment for when the list is empty.

PR #660, opened 2022-05-17
Bug fixes and cleanup to attribute occurs-on contexts

krame505:
# Changes
* Fix a translation bug when there are multiple inherited occurs-on contexts for a single type variable; use `Math.max` for computing the size of the thunk array (was previously calling `max` which isn't globally defined)
* Add MWDA checks for places where a synthesized occurs-on context is solved (e.g. in calling a function with a syn occurs context) that the actual flow dependencies of the attribute are a subset of the flow type in the constraint.  I apparently forgot to implement/test this when implementing occurs-on contexts?
* Remove special-case workaround in the MWDA for lists from when they were a modification

# Documentation
This is a fix to an existing feature; no new documentation is added besides source code comments.  I added test cases for the new MWDA warnings.

### DIFF
--- a/grammars/silver/compiler/analysis/typechecking/core/Context.sv
+++ b/grammars/silver/compiler/analysis/typechecking/core/Context.sv
@@ -6,6 +6,8 @@ monoid attribute contextErrors::[Message] occurs on Contexts, Context;
 attribute downSubst, upSubst occurs on Contexts, Context;
 propagate contextLoc, contextSource, contextErrors, downSubst, upSubst on Contexts;
 
+flowtype contextErrors {contextLoc, contextSource, env, frame, grammarName, compiledGrammars, config} on Context;
+
 aspect production instContext
 top::Context ::= cls::String t::Type
 {
--- a/grammars/silver/compiler/analysis/warnings/flow/Inh.sv
+++ b/grammars/silver/compiler/analysis/warnings/flow/Inh.sv
@@ -210,21 +210,10 @@ function inhDepsForSynOnType
 {
   local contexts::Contexts = foldContexts(ns.contexts);
   contexts.env = env;
-  
-  local isList::Boolean =
-    case t of
-    | appType(listCtrType(), _) -> true
-    | decoratedType(appType(listCtrType(), _), _) -> true
-    | _ -> false
-    end;
 
   return
     if t.isNonterminal || (t.isDecorated && t.decoratedType.isNonterminal)
     then (just(inhDepsForSyn(syn, t.typeName, flow)), [])
-    -- TODO: Hack for lists, since these pretend not to be not be nonterminals yet use attributes in the nonspecialized implementation.
-    -- These attributes (i_headList, i_tailList, i_nullList, i_lengthList) have no inherited dependencies.
-    -- This can go away if we make lists a built-in/modification and scrap the unspecialized implementations.
-    else if isList then (just(mempty), [])
     else (
       map(set:fromList, lookup(syn, lookupAll(t.typeName, contexts.occursContextInhDeps))),
       concat(lookupAll(syn, lookupAll(t.typeName, contexts.occursContextInhSetDeps))));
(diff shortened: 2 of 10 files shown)


grammars/silver/compiler/analysis/typechecking/core/Context.sv
```diff
@@ -52,13 +57,16 @@ top::Context ::= attr::String args::[Type] atty::Type ntty::Type
...
     -- Check for ambiguous type variables.
-    -- Since we've already computed the final substitution, if t has any,
+    -- Since we've already computed the final substitution, if ntty has any,
     -- they could unify with something more specific in instance resolution here,
     -- and unify with something else in solving another instance later on.
     if !null(ntty.freeFlexibleVars)
     then map(
       \ tv::TyVar -> err(top.contextLoc, s"Ambiguous type variable ${findAbbrevFor(tv, top.freeVariables)} (arising from ${top.contextSource}) prevents the constraint ${prettyContext(top)} from being solved."),
       ntty.freeFlexibleVars)
+    -- atty should never have free type variables if ntty does not.
```
remexre, line 67: Is this enforced by the syntax for declaring occurrences, so an extension that caused this would be interfering?

(Also, any reason to use `error` instead of `[err(...)]`?)

grammars/silver/compiler/analysis/typechecking/core/Context.sv
```diff
@@ -78,16 +86,31 @@ top::Context ::= attr::String args::[Type] atty::Type inhs::Type ntty::Type
...
     else requiredContexts.contextErrors;
 
-  -- Not refining based on occurs-on constraints for now, since the appropriate inference should happen on the associated attribute access
-  top.upSubst = top.downSubst;
+  production substNtty::Type = performSubstitution(ntty, top.downSubst);
+  production substInhs::Type = performSubstitution(inhs, top.downSubst);
+  top.upSubst =
+    -- If the nonterminal type is known but the flow type inh set is unspecialized,
+    -- specialize it to the flow type of the attribute on the nonterminal.
+    -- This is a bit of a hack, since we don't properly support functional dependencies.
```
remexre, line 105: Is this situation like, the syntax `foo occurs on a` rather than `foo occurs on a { bar }`?

grammars/silver/compiler/analysis/typechecking/core/Expr.sv
```diff
@@ -66,6 +66,9 @@ top::Expr ::= e::Expr '(' es::AppExprs ',' anns::AnnoAppExprs ')'
   -- contexts in the environment.
   production infContexts::Contexts = foldContexts(e.contexts);
   infContexts.env = top.env;
+  infContexts.frame = top.frame;
+  infContexts.grammarName = top.grammarName;
+  infContexts.compiledGrammars = top.compiledGrammars;
```
remexre, line 71: was this previously an MWDA violation? Or are we demanding new attributes now

grammars/silver/compiler/analysis/warnings/flow/Inh.sv
```diff
@@ -757,11 +746,45 @@ Boolean ::= prod::ValueDclInfo  sigName::String  attrName::String  realEnv::Deco
     ignoreIfAutoCopyOnLhs(sigName, prod.namedSignature, realEnv, attrName); -- not autocopy (and on lhs)
 }
 
+-- In places where we solve a synthesized attribute occurs-on context,
+-- check that the actual deps for the attribute do not exceed the one specified for the context.
+aspect production synOccursContext
+top::Context ::= attr::String args::[Type] atty::Type inhs::Type ntty::Type
+{
+  -- oh no again
+  production myFlow :: EnvTree<FlowType> = head(searchEnvTree(top.grammarName, top.compiledGrammars)).grammarFlowTypes;
```
remexre, line 755: uhhhhhhhhhhhhhhhh production attributes are public API so this should get a much better doc comment........

grammars/silver/compiler/analysis/typechecking/core/Context.sv
remexre: Like, I could make an extension that gets the `error()` there, right? It'd just have to be interfering, which is why it's fine to do `error`?

grammars/silver/compiler/analysis/typechecking/core/Context.sv
krame505: No? I don't think I understand what you mean. 
If we have the function
```
synthesized attribute foo::String;
function getFoo
attribute foo i occurs on a =>
String ::= x::Decorate a with i
{ return x.foo; }
```
and we call it with a reference that is polymorphic in the set of inherited attributes
```
nonterminal Thing with foo, env;
flowtype Thing = foo {env};
production thing
top::Thing ::= t::Thing
{
  top.foo = getFoo(t);  -- t has type Decorated Thing with i
}
```
then we are trying to solve the constraint `attribute foo i occurs on Thing`, where `i` is a flexible type var.  That could be an error, but since we know the flow type of `foo` on `Thing` is `{env}`, we can instead just specialize `i` to be `{env}`.

grammars/silver/compiler/analysis/typechecking/core/Expr.sv
krame505: Demanding these attributes now, since we need flow info in performing the new MWDA check on context resolution.

grammars/silver/compiler/analysis/warnings/flow/Inh.sv
krame505: This is copy/pasted from every other place where we need to get the grammar's flow types, with that comment from Ted.  This should probably be refactored (maybe a hackathon project?) but I've just kept the comment to make grepping for all the places easier.

krame505:
Ok, merging this since this is semi-blocking me.

grammars/silver/compiler/analysis/typechecking/core/Context.sv
krame505: Ugh okay it looks like this makes type checking mutually circular with the flow analysis and breaks things.  Changing it to instead only look at manually specified flow types.

grammars/silver/compiler/analysis/typechecking/core/Context.sv
krame505: Yeah, I guess so?  Another extension could always construct a nonsensical `inhOccursContext` where the atty has type vars not in the ntty, but that is the fault of the extension.

PR #662, opened 2022-05-23
Add overloaded fromJson function to json library

krame505:
# Changes
Adds a type class method `fromJson` for converting arbitrary values from a json representation.

# Documentation
I updated a doc comment.

### DIFF
--- a/grammars/silver/json/Json.sv
+++ b/grammars/silver/json/Json.sv
@@ -9,33 +9,73 @@ nonterminal Json with jsonString;
 @{- Converts the JSON value to a string. -}
 synthesized attribute jsonString :: String;
 
-@{- ToJson represents conversion to JSON. -}
-class ToJson a {
+@{- JsonRepr represents conversion to and from JSON. -}
+class JsonRepr a {
   toJson :: (Json ::= a);
+  fromJson :: (Either<String a> ::= Json);
 }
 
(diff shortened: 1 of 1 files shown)


remexre:
Is there a reason not to make this a separate typeclass?

remexre:
I don't see any law books saying you can't write a law "if this type is also `FromJSON`, [... laws as you wrote them ...]" :P

The defacto standard libraries for Rust (serde) and Haskell (aeson) both separate these into two typeclasses; the top Google search results for `"$languageName" json` do:

- `"coq" json`: [coq-json](https://github.com/liyishuai/coq-json) has separate `JDecode`/`JEncode`
- `"purescript" json`: [purescript-simple-json](https://pursuit.purescript.org/packages/purescript-simple-json/8.0.0) has separate `ReadForeign`/`WriteForeign`
- `"scala" json`: [Play Framework's ScalaJson](https://www.playframework.com/documentation/2.8.x/ScalaJson) has separate `Reads`/`Writes`

(The first result for Agda, Coalton, and Mercury did not typeclasses for converting to/from JSON.)

IMO that's enough prior art going the same direction to say that we should have a reason for diverging?

krame505:
Fair, I'll change it.

remexre (approved):

PR #664, opened 2022-06-01
Re-enable support for getting parsed terminals on a ParseResult, without reflection hack

krame505:
# Changes
This re-adds support for getting the parsed terminals as `TerminalDescriptor`s on a parse result, which was previously removed due to the use of an awful hack using reflection to call a `getTokens` method on the generated Copper parser class, which was not included in the `CopperParser` interface.  

Instead this adds a new `HasTokens` interface with the appropriate `getTokens` method, and makes the generated Copper parsers implement this interface.  This avoids the need for reflection.

Also (unrelated) un-reserve `Function` and `Production` as keywords - these used to be part of the function type syntax but have been deprecated for so long that there isn't really a good reason to keep recognising that syntax, AFAIK.

# Documentation
Added doc comments as appropriate.


grammars/silver/compiler/extension/deprecation/TypeExpr.sv
```diff
@@ -1,20 +0,0 @@
-grammar silver:compiler:extension:deprecation;
```
remexre, line 1: TIL this syntax existed!

remexre (approved):
LGTM

grammars/silver/compiler/extension/deprecation/TypeExpr.sv
krame505: Yeah, I guess at one point functions and productions had different types...

PR #665, opened 2022-06-05
Implement a language server for Silver that provides semantic tokens

krame505:
# Changes
* Implement a language server for Silver that provides semantic tokens
* Implement a VS code extension using the language server 
* Add lexer classes for specifying LSP semantic token types on Silver's terminals
* Adds a new `insert semantic token` parser action for specifying semantic tokens to be returned by the parser
* Bug fix, super classes were missing from the list of lexer classes on terminal objects

There is probably more that can be done in terms of semantic tokens in Silver.  We probably want some way of looking outside of the currently parsed file in contributing additional semantic tokens from the decorated syntax tree - if we have access to the environment then we can e.g. can distinguish what is a standard library function or not.  Trying to do this with an attribute gets a bit weird though, as extensions that introduce concrete syntax shouldn't want the semantic tokens of what they forward to.  Would be solved by a CST/AST split, I guess.

The VS code plugin I made for testing this out is dumped here under `support/vs-code`.  That still requires quite a bit of cleanup, and I have no idea whether putting a VS code extension under a subfolder of a bigger repo is recommended/acceptable practice.  Some restructuring is possibly needed to actually publish the extension, anyway.

The `build.sh` script in the `language-server/` folder requires you to first run `mvn install` in wherever you have Copper checked out.  We should figure out what it takes to publish Copper on Maven Central to avoid this step.  Shouldn't be too hard I think, since it should already be packaged appropriately?

This PR is new-jars because it introduces and uses the new syntax for specifying semantic tokens.

# Documentation
I added some doc comments in the LSP server implementation and accompanying Silver library.
Docs for the `insert semantic token` parser action are at https://github.com/melt-umn/melt-website/pull/42.


language-server/langserver/src/main/java/edu/umn/cs/melt/silver/langserver/CopperSemanticTokenEncoder.java
```diff
@@ -0,0 +1,140 @@
+package edu.umn.cs.melt.silver.langserver;
```
ericvanwyk, line 1: Can you add a few more comments here about what this is meant to do? Or is there additional documentation somewhere to explain this? I see some docs for `insert semantic token` but we (eventually) will want some guide on what one needs to do to make an LSP for some other language.

support/vs-code/silverlsp/README.md
```diff
@@ -0,0 +1,70 @@
+# silverlsp README
```
ericvanwyk, line 1: Strictly speaking, one can run LSP with editors other than VS Code. Do you envision some other organization to reflect this? I know this is early days, but this may be worth considering.

ericvanwyk (approved):
Looks good. This will be nice to have.  A few comments about organization to consider.

support/vs-code/silverlsp/README.md
krame505: This is the VS code extension for Silver that makes use of the Silver language extension. I have named the extension `silverlsp`, it lives under `support/vs-code` for the moment.  The README file here is auto-generated and needs to be updated once this is beyond an early prototype.  The actual reusable language server implementation is under the `language-server` folder.

PR #666, opened 2022-06-09
Misc LSP changes

krame505:
# Changes
Fixing some minor issues I encountered when setting up the silver LSP extension on another computer.

# Documentation
None needed


krame505:
I can't seem to install it on my Chromebook, anyway. I'm just bumping it for the language server implementation for now, I guess silver itself still technically supports java 8 for now, but I don't think there is a strong reason to continue to care about that once we ditch the eclipse ide plugin.

PR #667, opened 2022-06-13
Implement monad-related utilities

krame505:
# Changes
Some stuff I needed in refactoring the Silver driver to be monadic:
* Implement the `EitherT` monad transformer version of `Either`
* Implement a `MonadFix` type class, for recursive monadic computations
* Implement recursive `mdo` notation, as supported in Haskell (https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/recursive_do.html)
* Implement `unsafeInterleaveIO`, as found in Haskell (and associated audit of where `unsafeTrace` is needed in `IO` monad plumbing)

# Documentation
I added doc comments on the stdlib additions.  Docs for `mdo` added in https://github.com/melt-umn/melt-website/pull/43

### DIFF
--- a/grammars/silver/core/IO.sv
+++ b/grammars/silver/core/IO.sv
@@ -1,5 +1,13 @@
 grammar silver:core;
 
+@{-
+ - Representation of a monadic IO action.
+ -
+ - The stateIn/stateOut threading exists to ensure that IO actions happen in
+ - the proper order, thus the invariant that stateOut should be demanded before stateVal.
+ - Note that unsafeInterleaveIO intentionally violates this, causing IO actions to be evaluated when
+ - their value is demanded.
+ -}
 closed nonterminal IO<a> with stateIn<IOToken>, stateOut<IOToken>, stateVal<a>;
 
 abstract production bindIO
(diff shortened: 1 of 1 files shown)


grammars/silver/compiler/modification/lambda_fn/java/Lambda.sv
```diff
@@ -56,6 +56,6 @@ aspect production lambdaParamReference
 top::Expr ::= q::PartiallyDecorated QName
 {
   top.translation = s"common.Util.<${finalType(top).transType}>demandIndex(lambda_${toString(q.lookupValue.dcl.lambdaId)}_args, ${toString(q.lookupValue.dcl.lambdaParamIndex)})";
-  top.lazyTranslation = top.translation;
+  top.lazyTranslation = wrapThunk(top.translation, top.frame.lazyApplication);
```
remexre, line 59: Is this a bugfix? And what was the bug? (it looks like it adds more laziness than was previously present unless in an action block?)

remexre (approved):
lgtm, though kinda wondering if we could choose a better name than mdo... probably doesn't matter, and the consistency with haskell makes up for the non-obviousness of the name...

krame505:
Yeah, I guess I also considered `do rec` but Haskell has a different recursive do notation that uses the `rec` keyword differently, which I figured would be even more confusing...

PR #668, opened 2022-06-13
Highlight aspect signature names as semantic tokens

krame505:
# Changes
This adds aspect production and aspect default production signature names to the `sigDefs` parser attribute, so they show up as "parameter" semantic tokens.  This required refactoring to the syntax of aspect default productions.  This change is new-jars due to the resulting bootstrapping, so I would appreciate getting this merged quickly.  

# Documentation
None needed, this is purely an internal change/refactor that improves semantic highlighting.


remexre (approved):
nothing jumps out as ridiculous, but i think i need to read thru the LSP stuff to grok it all. since it sounds urgent that it get merged, approving

PR #670, opened 2022-06-22
Refactor build process to transitively rebuild dependent grammars

### LINKED ISSUE #36: Build fails without --clean meta-bug

krame505:
# Changes
This fixes #36.  This PR is based on #669 and should be reviewed after that is merged.

When building a grammar causes changes to the interface file, this change now causes the driver to rebuild any needed grammars that depend on the changed grammar, and repeat that process iteratively.  Comparing interface files is a bit more nuanced than just comparing the serialized files, since we don't care about locations, and also type variables can differ between builds due to the use of `genInt()`.  Instead equality attributes on the various `DclInfo` nonterminals allow the appropriate type unification to be done.

Note that there is still a bit of an issue with comparing flow defs between builds, as `genInt` is used for the names of anonymous flow vertices.  For the moment I am just ignoring flow defs in the comparison, which means that a change to a grammar that affected the inferred flow types without causing any other interface changes would cause dependent grammars to not be rechecked.  This isn't a translation issue, and we usually run the flow analysis with `--clean` to ensure that the results are complete anyway, so I don't think that this is much of an issue.

I left parser generation as a separate phase from java translation for the moment.  With this change, whenever concrete syntax included in a parser is changed, we will now recompile the grammar containing the parser declaration.  So in theory we could now merge parser generation into the java translation step, which would significantly simplify some things. However there are some changes to a dependency grammar (e.g. adding a new attribute) that don't affect the concrete syntax at all, and so we shouldn't need to regenerate the parser.  So the existing optimization that dumps the syntax AST and compares it before generating a parser is still useful and should be kept, IMO.  

# Documentation
Added source comments as needed.  https://melt.cs.umn.edu/silver/concepts/origins/#build-issues can be deleted once this is merged.


grammars/silver/compiler/analysis/typechecking/core/Context.sv
```diff
@@ -64,10 +64,10 @@ top::Context ::= attr::String args::[Type] atty::Type ntty::Type
     then map(
       \ tv::TyVar -> err(top.contextLoc, s"Ambiguous type variable ${findAbbrevFor(tv, top.freeVariables)} (arising from ${top.contextSource}) prevents the constraint ${prettyContext(top)} from being solved."),
       ntty.freeFlexibleVars)
-    -- atty should never have free type variables if ntty does not.
-    else if !null(atty.freeFlexibleVars)
+    -- atty should never have free type variables if ntty does not, except in case of errors elsewhere.
+    else {-if !null(atty.freeFlexibleVars)
```
ericvanwyk, line 68: What is the reason for leaving this commented out if-expression in the code?

grammars/silver/compiler/driver/BuildProcess.sv
```diff
@@ -13,180 +13,188 @@ type SVParser = (ParseResult<Root> ::= String String);
...
+    | right(a) -> do {
+      -- Figure out build env from environment and args
+      benv::BuildEnv <- determineBuildEnv(a);
+      -- Because we want printing the version to work even if the environment is messed up
+      -- we premptively handle that here. This is slightly unfortunate.
+      -- Ideally, version printing would be just another thing we could have the command
+      -- line decide to go do, but currently it's hard to re-use code if we do that.
+      if a.displayVersion then
+        throwRunError(127, -- error code so 'ant' isnt run
+          "Silver Version 0.4.5-dev\n" ++
```
ericvanwyk, line 67: Should the version number really be hard coded like this?

ericvanwyk (approved):
looks ok to me. Seemed like some overlap between this one and the other one on monadifying the build process.

grammars/silver/compiler/driver/BuildProcess.sv
krame505: Probably not, but it was previously, and I don't know of a better way.  Any ideas?

grammars/silver/compiler/analysis/typechecking/core/Context.sv
krame505: Might be useful to uncomment for testing/debugging purposes.

PR #671, opened 2022-06-26
Fix bugs in implicit monad case rewriting

RandomActsOfGrammar:
# Changes
This fixes a couple of bugs which I have introduced over the past due to not running enough testing, both of which have to do with matching on implicit expressions.  The expressions on which the case expression matched were not being rewritten, leading to errors if they used monads implicitly.  They were also being bound in, then the new `bind` applications were being rewritten, leading to typing errors.

I also simplified the rewriting in fixing this.

I plan on getting the Caml Light implementation set up on Jenkins which will test this.

# Documentation
I added a comment explaining the rewriting.  This does not affect anything outside the production as it is fixing a single production's behavior to the intended behavior, so no doc comments or website documentation are needed.


grammars/silver/compiler/extension/implicit_monads/Case.sv
```diff
@@ -178,30 +162,21 @@ function monadicMatchTypesNames
...
          | _, [] -> pair([], elst)
          | e::etl, t::ttl ->
-           let ety::Type = decorate e with {env=env; mDownSubst=sub; frame=f; grammarName=gn;
-                                            downSubst=sub; finalSubst=sub;
-                                            compiledGrammars=cg; config=c; flowEnv=fe;
-                                            expectedMonad=em; isRoot=iR; originRules=oR;}.mtyperep
+           let decE::Decorated Expr with {downSubst, finalSubst, frame, grammarName, isRoot,
+                                          originRules, compiledGrammars, config, env, flowEnv,
+                                          expectedMonad, mDownSubst} =
+               decorate e with {env=env; mDownSubst=sub; frame=f; grammarName=gn;
```
krame505, line 168: Wouldn't it be much simpler and more efficient to pass in a list of `Decorated Expr with {downSubst, finalSubst, frame, grammarName, isRoot, originRules, compiledGrammars, config, env, flowEnv, expectedMonad, mDownSubst}` for `elist`, than to re-decorate them all here?

RandomActsOfGrammar: I changed it and it's a bit simpler, though not as much as I would like.  It still needs a number of the arguments for other reasons.

grammars/silver/compiler/extension/implicit_monads/Case.sv
```diff
@@ -73,26 +65,36 @@ top::Expr ::= 'case' es::Exprs 'of' vbar::Opt_Vbar_t ml::MRuleList 'end'
...
   local attribute monadStuff::([(Type, Expr, String)], [Expr]);
-  monadStuff = monadicMatchTypesNames(es.rawExprs, ml.patternTypeList, top.env, ml.mUpSubst, top.frame,
-                                      top.grammarName, top.compiledGrammars, top.config, top.flowEnv, [],
-                                      top.location, 1, top.expectedMonad, top.isRoot, top.originRules);
+  monadStuff = monadicMatchTypesNames(decExprs, ml.patternTypeList, [], top.env,
+                                      ml.mUpSubst, top.expectedMonad, top.location, 1);
+  local decExprs::[Decorated Expr with {downSubst, finalSubst, frame, grammarName,
+                                        isRoot, originRules, compiledGrammars, config,
+                                        env, flowEnv, expectedMonad, mDownSubst}] =
+        map(\ e::Expr ->
```
krame505, line 73: Is there still a reason why the expressions from `es` need to be re-decorated here?  Why not just make a new attribute on `Exprs` of type `[Decorated Expr with {downSubst, finalSubst, frame, grammarName, isRoot, originRules, compiledGrammars, config, env, flowEnv, expectedMonad, mDownSubst}]` that extracts the appropriately-decorated `Expr` children?  

Also you could write a type alias
```
type MonadInhs = {
  downSubst, finalSubst, frame, grammarName, isRoot, originRules,
  compiledGrammars, config, env, flowEnv, expectedMonad, mDownSubst
};
```
and then instead just write the decorated type as `Decorated Expr with MonadInhs`

krame505: In fact, I think there is a type inference bug here since all the expressions get the same `mDownSubst`, it might be possible for a type variable to be unified in two different ways in two expressions in `es`.  And it's a bit hard to follow but I think the `mUpSubst`s from the expressions decorated here are ignored and don't make it in to the `finalSubst`? 
 Decorating the entire `Exprs` tree at once and extracting the needed `Decorated Expr`s from that should make this a bit cleaner to fix.

Hmm, I guess that decorating `es` might not  quite work if `downSubst` is needed on the decorated expressions, since there isn't a `downSubst` equation for `es` on `caseExpr_c`, and so this would end up being an orphaned equation.  You could define a new production attribute equal to `es` and decorate that instead.

krame505 (approved):

krame505:
Looks like the build was failing because caml-light doesn't have a `develop` branch - I changed the default branch to `develop` to fix this.

RandomActsOfGrammar:
Lucas, can you take a look at the `caml-light` repository?  The build is failing there because the testing can't find the Jar file, even though the build stage succeeds.

krame505:
I don't see anything obviously wrong, but running each test in parallel a separate task can be tricky and seems like overkill in this case, since I am guessing the tests should just take a few seconds each?  

I would suggest maybe making a script that runs the tests, and have the Jenkinsfile just call that.

Issue #594, opened 2021-11-27
Infix Operator Tracking Issue


remexre:
(Making a tracking issue so I don't forget!)

- [ ] Create typeclasses for all the typical infix operators (#451)
- [ ] Make infix operator syntax forward to the typeclass methods
- [ ] Change infix operators to use a single production that resolves associativity and precedence while forwarding ("custom operator infrastructure")
- [ ] Make infix operators a non-interfering extension?
- [ ] Expose custom infix operators to users? (Discuss me!)
- [ ] Idiom brackets? (`[| f x1 x2 x3 |]` => `ap(ap(map(f, x1), x2), x3)`) Not hugely related, but replaces some of the need for:
- [ ] Traditional operators for Haskellism (`<$>`, `<*>`, `>>=`, `>=>`, `$>`, `*>`, etc.)
  - If we add these, I promise to write a doc page about them, with a spoken name (e.g. `applyFirst`), the definition in terms of typeclass methods, the definition in terms of `do` where possible, and an example of use; someone else should come up with a better term than "Haskellism," though.

PR #676, opened 2022-07-11
Algebra Typeclasses

remexre:
See #451.


grammars/silver/core/DivisionRing.sv
```diff
@@ -0,0 +1,30 @@
+grammar silver:core;
+
+@{- A typeclass for rings that have a reciprocal operation, which turns
```
krame505, line 3: Incomplete comment?

grammars/silver/core/BooleanAlgebra.sv
```diff
@@ -0,0 +1,10 @@
+grammar silver:core;
+
+@{- This typeclass guarantees that a Heyting algebra behaves exactly like the
+  - booleans. It doesn't provide any operators, but may be useful for algorithms.
+  -
+  - Laws:
+  -
+  - * Excluded Middle: `disj(x, not(x)) = true`
```
krame505, line 8: Should `not(not(x)) = x` also be a law here?

krame505 (approved):
Looks good, approving with a couple of minor comments. 
Are you planning to change all the built-in operator productions to forward to these functions in a subsequent pull request?

grammars/silver/core/DivisionRing.sv
remexre: Fixed in 715424caab3e4c9a5cf9808c8f338446d882f47e.

grammars/silver/core/BooleanAlgebra.sv
remexre: Hm, PureScript only has the one, and in a sufficiently strong logic, one can certainly derive double negation elimination from the law of excluded middle; I'll see if I can put one of the standard proofs that those are equivalent into term-rewrite form on my way to campus.

If so, do you think it'd make sense to add to the doc comment that double negation elimination is implied by this (possibly with the proof)?

remexre:
> Are you planning to change all the built-in operator productions to forward to these functions in a subsequent pull request?

I was planning on it (#594), though if the performance suffers, I might not. (This is also very as-time-allows.)

grammars/silver/core/BooleanAlgebra.sv
remexre: Okay yeah, couldn't derive it so I added the `conj` "inverse law" for that, and double negation elimination in d843fd780432d202c0e6436df21497adc3c60465.

PR #677, opened 2022-07-14
Fix translation bug with decorated type variable types

krame505:
# Changes
Fixes a bug causing invalid java code to be generated for productions with children like `Decorated a` where there is an inherited attribute occurs-on type constraint for `a`.

# Documentation
None, this is a bug fix.

### DIFF
--- a/grammars/silver/compiler/translation/java/core/NamedSignature.sv
+++ b/grammars/silver/compiler/translation/java/core/NamedSignature.sv
@@ -245,16 +245,17 @@ s"""private Object child_${n};
   }
 """;
 
+  local ntType::Type = if ty.isDecorated then ty.decoratedType else ty;
+
   top.childTypeVarElem =
     if lookupBy(typeNameEq, ty, top.sigInhOccurs).isJust
-    then s"type_${ty.transTypeName}"
+    then s"type_${ntType.transTypeName}"
     else "-1";
   
-  local ntType::Type = if ty.isPartiallyDecorated then ty.decoratedType else ty;
   top.childStaticElem =
-    if lookupBy(typeNameEq, ntType, top.sigInhOccurs).isJust
-    then s"\t\tchildInheritedAttributes[i_${n}] = new common.Lazy[count_inh__ON__${ty.transTypeName}];\n"
-    else if ntType.isNonterminal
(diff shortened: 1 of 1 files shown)


grammars/silver/compiler/translation/java/core/NamedSignature.sv
```diff
@@ -245,16 +245,18 @@ s"""private Object child_${n};
...
     else "-1";
   
-  local ntType::Type = if ty.isPartiallyDecorated then ty.decoratedType else ty;
   top.childStaticElem =
-    if lookupBy(typeNameEq, ntType, top.sigInhOccurs).isJust
-    then s"\t\tchildInheritedAttributes[i_${n}] = new common.Lazy[count_inh__ON__${ty.transTypeName}];\n"
-    else if ntType.isNonterminal
+    if lookupBy(typeNameEq, ty, top.sigInhOccurs).isJust
+    then s"\t\tchildInheritedAttributes[i_${n}] = new common.Lazy[count_inh__ON__${ntType.transTypeName}];\n"
+    else if ty.isNonterminal || ty.isPartiallyDecorated && ntType.isNonterminal
```
remexre, line 259: these associate `a || (b && c)`, right?

krame505: yeah

grammars/silver/compiler/translation/java/core/NamedSignature.sv
```diff
@@ -245,16 +245,18 @@ s"""private Object child_${n};
   }
 """;
 
+  local ntType::Type = if ty.isDecorated then ty.decoratedType else ty;
+  ntType.boundVariables = ty.boundVariables;
+
   top.childTypeVarElem =
     if lookupBy(typeNameEq, ty, top.sigInhOccurs).isJust
-    then s"type_${ty.transTypeName}"
+    then s"type_${ntType.transTypeName}"
```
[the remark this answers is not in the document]
krame505: What this stores is a unique id for the child's type variable if the child is an `a`, `Decorated a` or `PartiallyDecorated a`, where `a` has inherited occurs-on contexts.  The variable storing this id should be named purely based on the type var, not the entire type, which previously led to the translation bug.

remexre (approved):

PR #678, opened 2022-07-17
Include partially decorated children in propagated monoid attribute equations

krame505:
# Changes
It seems that in almost every case, the desired behavior for monoid attributes on partially decorated children is to include the value of the attribute on the child.  For example in host languages that have non-forwarding productions with partially decorated children, we probably want to include the error messages from the partially decorate children.  In the rare cases where this is not the desired behavior, one can simply not propagate the attribute for the production.

This allows for a significant cleanup in Silver, removing a number of `<-` equations for partially decorated children.  It also seems to be much more straightforward for `exprRef` to include error messages from its child in its errors; this allows for the removal of a number of interfering equations.

# Documentation
The behavior of partially decorated children with respect to automatic attributes seems to have been previously undocumented, however I have added some notes about this at https://github.com/melt-umn/melt-website/tree/feature%2Fmonoid-partial-dec-propagate

### DIFF
--- a/grammars/silver/compiler/definition/core/Expr.sv
+++ b/grammars/silver/compiler/definition/core/Expr.sv
@@ -26,7 +26,9 @@ flowtype forward {} on Exprs, ExprInhs, ExprInh, ExprLHSExpr;
 
 flowtype errors {decorate} on Exprs, ExprInhs, ExprInh, ExprLHSExpr;
 
-propagate errors, freeVars on Expr, Exprs, ExprInhs, ExprInh, ExprLHSExpr;
+propagate errors on Expr, Exprs, ExprInhs, ExprInh, ExprLHSExpr
+  excluding terminalAccessHandler;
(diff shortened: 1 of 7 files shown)


grammars/silver/compiler/definition/core/Expr.sv
```diff
@@ -394,10 +382,6 @@ top::Expr ::= e::Expr '.' q::QNameAttrOccur
   top.unparse = e.unparse ++ "." ++ q.unparse;
   propagate freeVars;
   
-  -- We don't include 'q' here because this might be a terminal, where
-  -- 'q' shouldn't actually resolve to a name!
-  top.errors := e.errors ++ forward.errors;
```
remexre, line 399: Was it correct to remove this equation too?

remexre (approved):

remexre:
Is this the only PR in this set of changes, or are there downstream PRs to look at too?

krame505:
The only Silver PR, for now.  I just made one for the docs.

PR #679, opened 2022-07-19
Be more conservative about inlining strategy attributes

krame505:
# Changes
We were being a bit too aggressive about inlining strategy attributes, which led to some unexpected behavior.  For example
```
strategy attribute optimize = repeat(optimizePass)
  occurs on Root;
partial strategy attribute optimizePass = someTopDown(rule on Expr of addOp(e, const(0)) -> e end)
  occurs on Root, Expr;
propagate optimize on Root;
propagate optimizePass on Root, Expr;
```
Here `optimize` only occurs on `Root`, while `optimizePass` occurs on both `Root` and `Expr`.  When generating equations for `optimize` on `Root`, we would inline `optimizePass` into `optimize`, and see that since the rule for `Expr` always fails on `Root`, the whole `optimize` strategy reduces to just `id`.

To fix this, we need to be a bit more careful about when we allow inlining.  The conservative rule I choose was that we only inline attribute `foo` into `bar` if `foo` occurs on all the nonterminals mentioned in `foo`'s strategy expression as rules/congruences.  

# Documentation
This is a bug fix, really, so no new documentation needed.

### DIFF
--- a/grammars/silver/compiler/extension/strategyattr/StrategyExpr.sv
+++ b/grammars/silver/compiler/extension/strategyattr/StrategyExpr.sv
@@ -714,6 +714,8 @@ top::StrategyExpr ::= n::Name s::StrategyExpr
     else [pair(sName, s)];
   top.freeRecVars := remove(n.name, s.freeRecVars);
 
+  -- Decorate s assuming that the bound strategy is total, in order to check for totality.
+  -- See Fig 4 of the strategy attributes paper (https://www-users.cse.umn.edu/~evw/pubs/kramer20sle/kramer20sle.pdf)
   local s2::StrategyExpr = s;
   s2.recVarTotalEnv = pair(n.name, true) :: s.recVarTotalEnv;
   s2.recVarTotalNoEnvEnv = pair(n.name, true) :: s.recVarTotalNoEnvEnv;
--- a/grammars/silver/compiler/extension/strategyattr/StrategyExpr.sv
+++ b/grammars/silver/compiler/extension/strategyattr/StrategyExpr.sv
@@ -28,7 +28,7 @@ monoid attribute freeRecVars::[String];
 monoid attribute partialRefs::[String];
 monoid attribute totalRefs::[String];
 monoid attribute matchesFrame::Boolean with false, ||;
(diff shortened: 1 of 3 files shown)


krame505:
After some discussion with @remexre I realized that this actually still isn't conservative enough - it is possible that `optimizePass` could traverse over a third nonterminal (e.g. `Stmt`) that isn't mentioned at all in the strategy expression, such that inlining would still result in different behavior.  

So to be totally safe, we just shouldn't inline any strategy that contains a traversal combinator.  This isn't really a big loss though, since the case where inlining helps the most often is when the inlined strategy is just a collection of rules.

grammars/silver/compiler/extension/strategyattr/StrategyExpr.sv
```diff
@@ -696,37 +707,39 @@ top::StrategyExpr ::= n::Name s::StrategyExpr
...
-      recVarTotalEnv = pair(n.name, true) :: s.recVarTotalEnv;
-      env = s.env; config = s.config; grammarName = s.grammarName; recVarNameEnv = s.recVarNameEnv; outerAttr = s.outerAttr;
-    }.isTotal;
-  top.isTotalNoEnv =
-    decorate s with {
-      recVarTotalNoEnvEnv = pair(n.name, true) :: s.recVarTotalNoEnvEnv;
-      config = s.config; grammarName = s.grammarName; recVarNameEnv = s.recVarNameEnv; outerAttr = s.outerAttr;
-    }.isTotalNoEnv;
+
+  local s2::StrategyExpr = s;
```
remexre, line 717: Comment or better name for s2? Is the point of it to like, be the base case of a fixed point for totality checking?

krame505: Sort of like a fixed-point I guess, but we only need to decorate it twice in total.  See Fig 4 in the [strategy attributes paper](https://www-users.cse.umn.edu/~evw/pubs/kramer20sle/kramer20sle.pdf)

I just refactored this from using a `decorate` expression, but I can add a comment.

PR #684, opened 2022-09-08
Delete Eclipse project files for Silver runtime

krame505:
# Changes
The presence of Eclipse project files in the Silver runtime project was causing some issues in editing the runtime using VS Code.  Code apparently has some limited support for Eclipse projects, but the project classpath referencing a `SILVER_HOME` variable was creating issues with resolving the generated sources referenced in the runtime.  Since presumably nobody is still using Eclipse for working on the Silver runtime, deleting the project files means VS Code will treat them as unmanaged sources.  Long term, if we move to building the Silver runtime with Maven, we will avoid all these IDE classpath issues.

Also there were 2 apparently different Eclipse projects at different directory levels, one of which had somehow gained a hard coded absolute path (?!)

# Documentation
The now-irrelevant documentation for editing the runtime project in Eclipse was deleted.


ericvanwyk (approved):
lgtm

PR #691, opened 2022-09-12
Rename HasTokens interface to SilverCopperParser and make it extend SilverParser

krame505:
# Changes
The Copper parsers generated by Silver have a `getTokens` method, not present in the `CopperParser` interface; this method is defined in a separate interface `HasTokens`.  So every place that can accept any Copper parser needs to have a generic parameter `PARSER extends CopperParser<ROOT, CopperParserException> & HasTokens`, which is a bit of a mouthful.  

Instead, we can just define an interface `SilverCopperParser` with the `getTokens` method, and make it extend `CopperParser`. The generated Copper parser classes can then just implement this interface, and we can use `SilverCopperParser<ROOT>` as the type of any Copper parser with a `hasToken` method.

Also, fix a minor Copper build bug - parsers weren't being re-generated with `--clean` if the Copper spec is unchanged.

# Documentation
I updated the javadoc comment on the interface.


krame505:
(Note that this refactor is a prerequisite for dynamically loading different parsers in the LSP server.)

alexandra-hanson (approved):

PR #692, opened 2022-09-13
Add support for dynamically loading a parser from a jar

krame505:
# Changes
This PR is based on #691, so review this after that is merged.

This adds support for dynamically loading a parser (and its dependencies) from a jar specified in the settings.

# Documentation
Added javadoc/comments for the new utility functions.


language-server/langserver/src/main/java/edu/umn/cs/melt/silver/langserver/Util.java
```diff
@@ -119,4 +128,71 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
...
+
+        // Sanity check: make sure it's actually a Silver-declared Copper parser
+        if (!SilverCopperParser.class.isAssignableFrom(parserClass)) {
+            throw new ReflectiveOperationException("Loaded class is not a Silver-generated Copper parser");
+        }
+
+        // Hacky way of determining the actual start nonterminal type of the parser, from its superclass generic type args.
+        Type[] genericParams = ((ParameterizedType)parserClass.getGenericSuperclass()).getActualTypeArguments();
+        if (genericParams.length != 2) {
+            throw new ReflectiveOperationException("Could not determine parser root type");
```
alexandra-hanson, line 168: This shouldn't be possible to reach, right? Since the superclass would be CopperParser and we know this from the last check. Is this another sanity check? Either way maybe the exception message could be clearer?

krame505: Essentially a sanity check.  This could break if there was a change to the super class used by Copper.

grammars/silver/compiler/definition/concrete_syntax/ast/LexerClassModifiers.sv
```diff
@@ -53,15 +53,10 @@ top::SyntaxLexerClassModifier ::=
 abstract production lexerClassExtends
 top::SyntaxLexerClassModifier ::= super::[String]
 {
-  local superRefsL :: [[Decorated SyntaxDcl]] = lookupStrings(super, top.cstEnv);
-  production superRefs :: [Decorated SyntaxDcl] = map(head, lookupStrings(super, top.cstEnv));
+  -- Lexer classes not included in this parser are ignored
```
alexandra-hanson, line 56: Maybe just add why this is to the comment for later reference (and/or maybe why it's ok to do so?)

alexandra-hanson (approved):
Maybe expand that one other comment I just noted but otherwise seems good.

PR #695, opened 2022-09-16
Refactor reusable lsp4j utilities into a seperate library

krame505:
# Changes
This PR is based on #692 and should be reviewed after that is merged.

This just pulls the utilities for interoperation between lsp4j and Silver/Copper into a seperate library that could be reused in LSP implementations for other languages.

# Documentation
Added a brief README.md file.

Issue #690, opened 2022-09-12
Tabs are always treated as 8 characters in reported column indices


krame505:
Terminals that span multiple lines are split into separate tokens, one on each line in the Silver LSP implementation.  However there seems to be a bug in this encoding logic (in `CopperSemanticTokenEncoder.java`), as seen with multi-line string templates:
![image](https://user-images.githubusercontent.com/8467850/189575198-72519f93-9bf7-4aaa-9d5a-978ce504f69b.png)

krame505:
It seems that Copper always reports column numbers with tabs padded to multiples of 8 characters.  This isn't compatible with the LSP spec, which uses absolute character indices.  A proper fix would require making the desired tab width configurable when invoking a generated Copper parser, so that we could set it as 1.

PR #697, opened 2022-09-17
Fix issue with semantic tokens for lines containing tabs.

krame505:
# Changes
This PR is based on #695 and should be reviewed after that is merged.

This is a workaround for #690 - just replace tabs with spaces before running the parser, for now.  Also fix a bug with reporting semantic tokens for terminals that begin or end with a newline.

# Documentation
Added a comment about the workaround.


ericvanwyk:
Does VS Code have settings so that on the editor side, all tabs are replaced by spaces?

RandomActsOfGrammar:
Is there a compelling reason for the Silver behavior of treating tabs as eight spaces, rather than as a single character?  I'm not sure the fix for this is changing the tabs to spaces before parsing, rather than changing the behavior of Silver to treat tabs as a single character.

krame505:
> Does VS Code have settings so that on the editor side, all tabs are replaced by spaces?

There are editor commands to substitute tabs for spaces or vice versa, but I'm not aware of a way to treat tabs as spaces for the purposes of semantic highlighting, though?  

> Is there a compelling reason for the Silver behavior of treating tabs as eight spaces, rather than as a single character? I'm not sure the fix for this is changing the tabs to spaces before parsing, rather than changing the behavior of Silver to treat tabs as a single character.

This is really an issue of how Copper reports locations, not Silver.  

In most editors the column numbers are computed assuming that tabs pad to multiples of some width - so padding is generally the desired behavior, to get error messages to line up with the actual editor location.  In emacs the default is tab width is 8, I'm not sure about other editors.  

But with LSP the token encoding is based on the number of characters, not the visual column index.  Replacing tabs with spaces keeps the character count aligned with the column index computed by Copper, for now.  Properly fixing this would require a change to the API of Copper-generated parsers, which I don't feel like tackling at the moment.

ericvanwyk:
What do white-space sensitive languages like Haskell and Python do for tabs? The language definition must assign them some fixed width, right?

It also seems like `tab` is no longer treated as a character to insert - many times `tab` directs the editor to indent the line in some appropriate way based on what is above it, inserting spaces, I presume, to make that happen. So maybe this problem will/should go away for some editors/languages.

krame505:
Haskell also seems to treat tabs as padding multiples of 8 in reported error column numbers:
```haskell
module Test where

x :: Integer
x =
	"hello"
```

```
Test.hs:5:9: error:
    • Couldn't match expected type ‘Integer’ with actual type ‘[Char]’
    • In the expression: "hello"
      In an equation for ‘x’: x = "hello"
  |
5 |         "hello"
  |         ^^^^^^^
```

Python doesn't report column numbers in error messages. only the line number.

We definitely _should_ support tabs in source files.  We don't use them much in Silver, but they do show up in the java translation in multi-line string templates.

ericvanwyk:
If Haskell treats a tab as equal to 8 spaces then it doesn't seem wrong for us to do the same. I agree with Dawn that the right solution is to handle tabs properly and adjust the Copper API as you say. But just open a ticket for this on Copper's repo as something to be fixed later.

krame505:
Turns out that properly fixing this doesn't actually require any Copper API changes - the `setTabWidth` on the `virtualLocation` in the parser class does what is required.

ericvanwyk:
Cool.

unironically (approved):

PR #703, opened 2022-09-27
Fix another MWDA crash case caused by aspecting a non-existent production

krame505:
# Changes
When we aspect an undefined production and don't find the flow graph, create an anonymous flow graph to use in the body MWDA checks, instead of crashing.

# Documentation
Added a source comment.


unironically:
Should aspecting a non-existing production be an error case here? Do we allow aspect productions which reference non-existing productions?

(Maybe I'm missing if/where this is picked up elsewhere during compilation).

unironically (approved):

PR #710, opened 2022-10-10
Remove usage of autocopy from Silver

krame505:
# Changes
* Replace all autocopy attributes with inherited attributes, and add propagate declarations as needed.
* Fix some incorrect equations for `isRoot`, which probably should never have been an autocopy in the first place.
* Improve strategy attribute flow types by removing some undesirable autocopy equations.

# Documentation
None needed, really?  This change makes the code more explicit.

### DIFF
--- a/grammars/silver/compiler/extension/doc/core/TypeClassDcls.sv
+++ b/grammars/silver/compiler/extension/doc/core/TypeClassDcls.sv
@@ -5,6 +5,7 @@ attribute docForName occurs on ClassBodyItem, InstanceBodyItem;
 
 @{- What to prefix 'child' declaration names with in docs and for hyperlinking. -}
 inherited attribute scopeName :: String occurs on ClassBody, ClassBodyItem, InstanceBody, InstanceBodyItem;
+propagate scopeName on ClassBody, ClassBodyItem, InstanceBody, InstanceBodyItem;
 
 aspect production typeClassDcl
 top::AGDcl ::= 'class' cl::ConstraintList '=>' id::QNameType var::TypeExpr '{' body::ClassBody '}'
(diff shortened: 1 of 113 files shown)


grammars/silver/compiler/extension/doc/core/TypeClassDcls.sv
```diff
@@ -70,15 +71,14 @@ top::ClassBodyItem ::= comment::DocComment_t item::ClassBodyItem
...
+  top.upDocConfig <- parsed.upDocConfig;
   top.docErrors <- parsed.errors;
 
-  local realDclDocs::[CommentItem] = filter((\x::CommentItem->!x.stub), item.docs);
+  local realDclDocs::[CommentItem] = filter((\x::CommentItem->!x.stub), forward.docs);
   local isDoubleComment::Boolean = length(realDclDocs) != 0;
   top.docs := if isDoubleComment
                 then [standaloneDclCommentItem(parsed)] ++ realDclDocs
-                else [dclCommentItem(top.scopeName ++ "." ++ item.docForName, item.docUnparse, item.grammarName, item.location, parsed)];
+                else [dclCommentItem(top.scopeName ++ "." ++ forward.docForName, forward.docUnparse, forward.grammarName, item.location, parsed)];
```
hypnoticOxbow, line 81: Is there an advantage to demanding from the forward as opposed to demanding from the child (item, in this case)?

The semantics seem similar here.

hypnoticOxbow (approved):
LGTM! :thumbsup:

grammars/silver/compiler/extension/doc/core/TypeClassDcls.sv
krame505: We are forwarding to `item` here, so there is no reason to decorate `item` twice.

Issue #716, opened 2022-10-17
Associativity defaults to non-associative, when left unspecified


RandomActsOfGrammar:
Consider this Silver grammar:
```
grammar err;

lexer class KEYWORD dominates {Name_t};

terminal Name_t   /[a-z][a-zA-Z0-9_]*/;

terminal If_t     'if'     lexer classes {KEYWORD};
terminal Else_t   'else'   lexer classes {KEYWORD};

terminal Semi_t   ';';
terminal Eq_t     '=';


nonterminal Stmt;
nonterminal Expr;

concrete productions top::Stmt
| 'if' e::Expr th::Stmt 'else' el::Stmt
{ }
| 'if' e::Expr th::Stmt
{ }
| name::Name_t '=' e::Expr ';'
{ }

concrete productions top::Expr
| name::Name_t
{ }


parser p::Stmt {
   err;
}
```
This should have an ambiguity for the dangling `else`, but this is what Silver outputs:
```
Found err
	[/home/micha576/research/melt/silver/generated/src/err/Silver.svi]
Compiling err
	[../err/]
	[Concrete.sv]
Found silver:core
	[/home/micha576/research/melt/silver/generated/src/silver/core/Silver.svi]
Checking For Errors.
Generating Translation.
	[err]
Generating parser err:p.
-------------
Final report:
-------------
Parser and scanner specification
   has 6 terminals, 3 nonterminals, 5 productions,
   and 0 disambiguation functions declared,
   producing 12 unique parse states
   and 12 unique scanner states.
0 useless nonterminals.
0 malformed regexes.
0 unresolved parse table conflicts:
   1 shift/reduce conflict, 0 reduce/reduce, 1 resolved.
No lexical ambiguities detected.
Parser code output to /home/micha576/research/melt/silver/generated/src/err/Parser_err_p.java.
Buildfile: /home/micha576/err/build.xml

init:

grammars:
    [javac] Compiling 14 source files to /home/micha576/research/melt/silver/generated/bin
    [javac] Ignoring source, target and bootclasspath as release has been set

jars:
      [jar] Building jar: /home/micha576/err/err.jar

dist:

BUILD SUCCESSFUL
Total time: 1 second
```

PR #719, opened 2022-10-18
Change Default Associativity Behavior

### LINKED ISSUE #718: Support association = none
We currently support `association = left` and `association = right` for resolving shift/reduce conflicts between operators of the same precedence.  I thought we supported `association = none` for when we want to ban both the shift and reduce (i.e. make `x + y + z` a syntax error if `'+'` has `association = none`), but it seems we don't have syntax for this in Silver.  Copper does support this already.  

This could easily be fixed at the same time as #716.

keltono:
# Changes
* Adds `none` option for associativity
* Changes default behavior to not set associativity if not specified 
# Documentation
* No new documentation required, mostly just a modification of existing code

closes #718 and #716 .


krame505:
> No new documentation required, mostly just a modification of existing code

https://melt.cs.umn.edu/silver/ref/decl/terminals/#terminal-association should probably be updated.

krame505:
Also, looks like the tests are now failing

keltono:
> Also, looks like the tests are now failing

Yeah I noticed this. I was going to ask about it in our meeting in a few minutes, since it's working on my machine :tm:..

krame505 (approved):
I _think_ the only blocker on this is that some of the downstream projects are failing?  That should be easy to fix.

Issue #553, opened 2021-09-22
Make the list "extension" a modification


krame505:
Currently lists in Silver are an "extension"; they forward to a nonterminal type, and cons/null/head/tail/length are all implemented as functions in silver:core. Consequently, this feature currently lives under `silver:compiler:extension:list`.  

However, this "extension" violates our rules for extensions in a number of ways.  First of all, we override the java translation of lists to a more efficient implementation.  While this in itself may appear to be acceptable for some notion of non-interference, another extension to Silver that doesn't explicitly handle list types can inadvertently discard the "list-ness" of a list type, leading to an incompatible translation.  There are a few other similar issues:
* The pretty-print of list types should always be `[a]` and not `silver:core:List<a>`
* We try to hide the "nonterminal-ness" of list types - we don't want to allow `Decorated [a]` or `attribute foo occurs on [a];` for example.  This gives rise to some other unfortunate workarounds where we do want to treat lists as nonterminals in their default implementation, e.g. https://github.com/melt-umn/silver/blob/605720a0e997a9fb3c4956af19e19768037b0554/grammars/silver/compiler/analysis/warnings/flow/Inh.sv#L214

For these reasons I suggest moving lists to be a modification rather than an extension and make the `listCtrType` production non-forwarding.  This does mean doing away with the default implementation completely; however if someone wants to add a new back-end for Silver, they should probably provide a more efficient implementation of lists anyway.

ericvanwyk:
Yes, moving it to the `modifications` directory, or somewhere else that indicates the same thing, may be the right thing to do for now.

But it sure seems like lists should be a nice extension to AGs. That fact that it isn't feels like we have done something else wrong somewhere. Maybe this goes on the Silver 2.0 list of things to consider.

PR #720, opened 2022-10-18
Misc cleanup in origins translation and runtime

krame505:
# Changes
* Make the types of the `Tracked` methods more specific
* Move implementation of `duplicate(OriginContext)` to the `Tracked` interface
* Better use of `implode` in translation
* Other misc cleanup

# Documentation
I updated some outdated documentation on the website.

### DIFF
--- a/grammars/silver/compiler/translation/java/core/Origins.sv
+++ b/grammars/silver/compiler/translation/java/core/Origins.sv
@@ -43,7 +43,7 @@ function makeOriginContextRef
 String ::= top::Decorated Expr --need .frame anno
 {
   local rulesTrans :: [String] = (if top.config.tracingOrigins then [locRule] else []) ++ map((.translation), top.originRules);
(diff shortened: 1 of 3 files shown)


grammars/silver/compiler/translation/java/core/Origins.sv
```diff
@@ -74,9 +74,7 @@ function getSpecialCaseNoOrigins
     -- These are forced to be untracked to prevent circularity
     "silver:core:OriginInfo",
     "silver:core:OriginInfoType",
-    "silver:core:OriginNote",
-    -- List is special(TM) because of it's special(TM) quasi-extension translation specialization
-    "silver:core:List"
```
remexre, line 79: Does this no longer hold, then? Was there a refactor at some point that allows lists to have them?

krame505: `silver:core:List` no longer exists, since #553

remexre (approved):

Issue #550, opened 2021-09-16
Remove support for autocopy


remexre:
Per Monday's discussion, it might be possible to add a special case in orphaned-equation-checking for propagating inherited attributes, then to remove `autocopy` from Silver altogether.

- [x] Remove `autocopy` usage from Silver
- [x] Remove `autocopy` usage from projects written in Silver
- [x] Add `propagate` special case
- [x] Add docs about the workaround?
- [ ] Remove `autocopy` support from Silver

krame505:
Preserving this explaination from Slack of why autocopy is a blocker for my partial remote decoration (or whatever we should call it) work...

If I have a production of type `Expr ::= e::Expr`, where `e` has an autocopy attribute `env`, and I want to take a reference to `e` of type `PartiallyDecorated Expr with {}`, then to avoid duplicate equations we want to ban any inherited equations on `e`.  However `e` implicitly gets an equation for `env` anyway due to autocopy, and there isn't currently a way to specify that we don't want a particular autocopy equation.
So if the partially decorated reference we took later gets decorated with `env`, we end up with a duplicate equation, and some attributes on `e` might be computed from the wrong `env`.
It might be possible to work around this and disable autocopy equations on children that have had a partially decorated reference taken, but that will just add more complexity.

PR #722, opened 2022-10-25
Implement `undecorates to` and decoration site expressions, and use them in Silver

krame505:
This is based on #550 and should be reviewed after that is merged.  Otherwise this branch is at a reasonable checkpoint, renaming `PartiallyDecorated` -> `Decorated!`, finishing the uniqueness analysis, and implementing decoration site projections will come in future PRs.

# Changes
This adds support for `undecorates to` in production bodies to specify how a tree containing unique reference can be undecorated, and `@e` syntax for wrapping a unique reference to `e` with no attributes as an undecorated term that, when decorated, gives the same tree as `e` refers to.  A very minimal version of the uniqueness analysis is also added, to the extent needed in the translation.  For now, multiple decorations of a unique reference with the same inherited attribute gives a runtime error.

# Documentation
Documentation is partially written, but has been partially outdated due to the subsequent renaming and work on uniqueness analysis.  As such it doesn't make sense to focus on updating the docs until this sequence of PRs is merged.


ericvanwyk (approved):
Looks OK to me.

PR #726, opened 2022-11-04
Adds QNames antiquote production.

hypnoticOxbow:
# Changes

# Documentation

The changes made by this PR are short, and do not require specific documentation as they are small additions to the metatranslation antiquotes.

### DIFF
--- a/grammars/silver/compiler/extension/convenience/Lists.sv
+++ b/grammars/silver/compiler/extension/convenience/Lists.sv
@@ -51,6 +51,14 @@ top::QNames ::= id1::QNameWithTL ',' id2::QNames
   top.qnames = [id1] ++ id2.qnames;
 }
 
+abstract production qNamesError
+top::QNames ::= msg::[Message]
+{
+  top.unparse = "<err>";
+  top.qnames = [];
+
+}
+
 --------------------------------------------------------------------------------
 
 function makeOccursDcls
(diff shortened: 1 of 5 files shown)


krame505 (approved):

grammars/silver/compiler/extension/convenience/Lists.sv
```diff
@@ -51,6 +51,14 @@ top::QNames ::= id1::QNameWithTL ',' id2::QNames
   top.qnames = [id1] ++ id2.qnames;
 }
 
+abstract production qNamesError
```
krame505, line 54: This "error" production doesn't actually flow up the errors in `msg`, since `errors` doesn't occur on `QNames`.  This production isn't needed anyway, see my comment below.

grammars/silver/compiler/extension/silverconstruction/Syntax.sv
```diff
@@ -117,6 +118,15 @@ top::QName ::= '$QName' '{' e::Expr '}'
       location=top.location);
 }
 
+concrete production antiquoteQNames
+top::QNames ::= '$QNames' '{' e::Expr '}'
+{
+  top.unparse = s"$$QNames{${e.unparse}}";
+  forwards to
+    qNamesError(
```
krame505, line 126: This can be `qNamesSingle(qNameWithTL(qNameError(...), botlNone())` to avoid adding a new "error" production that is only useful here.

krame505 (changes requested):

PR #730, opened 2023-01-17
Renaming of partially decorated references to unique references

krame505:
# Changes
* "Partially decorated references" are now referred to as "unique references"
* The type expression syntax `PartiallyDecorated` is changed to `Decorated!` (suggested by @RandomActsOfGrammar)

# Documentation
This will be reflected in the updated docs once feature/uniqueness-analysis is ready to merge.

### DIFF
--- a/grammars/silver/compiler/analysis/typechecking/core/Expr.sv
+++ b/grammars/silver/compiler/analysis/typechecking/core/Expr.sv
@@ -19,7 +19,7 @@ top::Expr ::=
 }
 
 aspect production productionReference
-top::Expr ::= q::PartiallyDecorated QName
+top::Expr ::= q::Decorated! QName
 {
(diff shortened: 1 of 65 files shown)


ericvanwyk:
Can you indicate the motivation for `Decorated!`?  Does this come from some other language or system?

ericvanwyk (approved):
Look good - pretty straightforward.

But add something about where the ! comes from. I presume some other system or language uses this symbol for a similar reason.

krame505:
According to @RandomActsOfGrammar:

> In logic, sometimes people write ∃! to mean "there exists a unique item such that".  That's why I suggested it.  I don't know if any other software systems use it.

ericvanwyk:
Something like this should be part of the documentation. Maybe find out more what systems/logics do this.  Thanks.

PR #736, opened 2023-02-20
Implements find references for silver LSP

### LINKED ISSUE #699: Implement "Find References" LSP request
A good starting point for work on the Silver LSP implementation might be to implement the [Find references request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_references).  This should be able to mostly reuse / copy from the "Go to definition" implementation, except that upon resolving the symbol, instead of looking up its source location in the environment, we would like to collect all the other references to the same name.

alexandra-hanson:
# Changes
Implements the "find references" feature of LSP for Silver. Adds capability & overrides references implementation in lsp4j and adds helper functions to the langserver grammar. Closes #699.

# Documentation
Comments describing helper functions were added.


grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -195,6 +202,19 @@ map:Map<String (Location, Decorated RootSpec, a)> ::= accessor::([(Location, a)]
     rs));
 }
 
+-- Create a map from a reference's full name to its path & location
```
alexandra-hanson, line 205: This won't include the actual declaration of the symbol, but will include all other references. Not sure if this is desired behavior.

krame505: It seems like the behavior in other language server implementations is to include the declaration.  I think that should be pretty easy to add though?

alexandra-hanson: Done in [78da06f](https://github.com/melt-umn/silver/pull/736/commits/78da06f65cede94027b294bca8deb693a329ef53), but I've also realized that some references aren't included when I try to use the feature, or occasionally a reference will be misidentified (e.g. when looking for "RootSpec" an unrelated symbol in a comment is included as a reference). I am not sure why this is the case: I was wondering if it has to do with using just the value, type, and attribute reference locations, but this also wouldn't explain why some references are misidentified.

grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -256,3 +280,26 @@ function findDeclLocation
...
+
+-- Looks up all references to symbol at the given location
+-- Returns a list of all reference locations
+-- Input is filename, line & col number, & decl map to resolve the symbol
+-- Uses refs map to lookup the reference paths & locations from the symbol fullName  
+function lookupReferenceLocations
+attribute fullName {} occurs on a =>
+[Location] ::= fileName::String line::Integer col::Integer decls::map:Map<String (Location, Decorated RootSpec, a)> refs::map:Map<String (String, Location)>
+{
+  return foldr(\a::[Location] b::[Location] -> a ++ b, [], map(\ item::(Decorated RootSpec, a) ->
```
krame505, line 292: Can just be `concat(...)` instead of `foldr(\a::[Location] b::[Location] -> a ++ b, [], ...)`. 

And  `concat(map(...))` can be  just `flatMap(...)`.

grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -195,6 +202,23 @@ map:Map<String (Location, Decorated RootSpec, a)> ::= accessor::([(Location, a)]
...
 
+-- Create a map from a reference's full name to its path & location
+function buildAllRefs
+annotation sourceLocation occurs on a,
+attribute fullName {} occurs on a =>
+map:Map<String (String, Location)> ::= accessor::([(Location, a)] ::= Decorated RootSpec) rs::[Decorated RootSpec]
+{
+  return directBuildTree(flatMap(\ r::Decorated RootSpec ->
+    (map(\item::(Location, a) ->
+      (item.2.fullName, r.grammarSource ++ item.1.filename, item.1),
```
alexandra-hanson, line 213: I'm thinking the bug I mentioned above has to do with using the grammarSource of the RootSpec in the path? But I'm not sure what path should be used as an alternative.

alexandra-hanson: I pushed what I tried most recently as described above for debugging purposes.

grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -256,3 +287,26 @@ function findDeclLocation
...
+-- Returns a list of all reference locations
+-- Input is filename, line & col number, & decl map to resolve the symbol
+-- Uses refs map to lookup the reference paths & locations from the symbol fullName  
+function lookupReferenceLocations
+attribute fullName {} occurs on a =>
+[Location] ::= fileName::String line::Integer col::Integer decls::map:Map<String (Location, Decorated RootSpec, a)> refs::map:Map<String (String, Location)>
+{
+  return flatMap(\ item::(Decorated RootSpec, a) -> 
+    map( \loc::(String, Location) -> updateLocPath(loc.1, loc.2), 
+    (map:lookup(item.2.fullName, refs))),
```
krame505, line 301: Minor nit, extra parentheses here

krame505 (commented):
I just tried it out again and it seems to be mostly working now.   The one main remaining issue is that finding references (and declaration) at the actual declaration location doesn't work.  This is because declarations aren't currently being included in the `*RefLocs` attributes.  This could be done by adding a bunch more aspects for these attributes, but a cleaner way is probably to do it in `buildFileRefs` by getting appropriate sort of `DclInfo` from the `defs` from the passed-in `RootSpec`s.  Let me know if you want to take a stab at this and I can give you some more detailed pointers on Silver's environment plumbing, or if you just want me to take it over so you can move on.  

Another minor issue I noticed is that the collected ref locs aren't treating names from qualified imports properly.  But that's a separate issue that can be addressed in another pull request.

grammars/silver/compiler/langserver/ReferenceLocations.sv
alexandra-hanson: Done in [7339972](https://github.com/melt-umn/silver/pull/736/commits/7339972b564f90afdf050fd5929e74378ffd5dd3)

grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -174,27 +174,70 @@ synthesized attribute valueFileRefLocs::map:Map<String (Location, Decorated Root
...
   return directBuildTree(flatMap(\ r::Decorated RootSpec ->
     map(\ item::(Location, a) ->
       (r.grammarSource ++ item.1.filename, item.1, head(map:lookup(item.2.sourceGrammar, r.compiledGrammars)), item.2),
-      accessor(r)),
+      accessor(r)) ++
+    flatMap(\def::Def ->
+      map(\item::EnvItem<a> ->
+        (grammarToPath(item.dcl.sourceGrammar) ++ (item.dcl.sourceLocation.filename), 
+        item.dcl.sourceLocation, 
+        head(map:lookup(item.dcl.sourceGrammar, r.compiledGrammars)),
```
krame505, line 210: Can't this just be `r`?

grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -174,27 +174,70 @@ synthesized attribute valueFileRefLocs::map:Map<String (Location, Decorated Root
...
+  top.allTypeRefs = buildAllRefs((.typeRefLocs), g.grammarList);
+  top.allAttributeRefs = buildAllRefs((.attributeRefLocs), g.grammarList);
 }
 
 function buildFileRefs
+annotation sourceLocation occurs on a,
 annotation sourceGrammar occurs on a =>
-map:Map<String (Location, Decorated RootSpec, a)> ::= accessor::([(Location, a)] ::= Decorated RootSpec) rs::[Decorated RootSpec]
+map:Map<String (Location, Decorated RootSpec, a)> ::= 
+accessor::([(Location, a)] ::= Decorated RootSpec)
```
krame505, line 198: Style note, when params spill onto another line they should be indented.

grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -174,27 +174,70 @@ synthesized attribute valueFileRefLocs::map:Map<String (Location, Decorated Root
...
+accessor::([(Location, a)] ::= Decorated RootSpec)  
+accessList::([EnvItem<a>] ::= Def)
+rs::[Decorated RootSpec]
 {
   return directBuildTree(flatMap(\ r::Decorated RootSpec ->
     map(\ item::(Location, a) ->
       (r.grammarSource ++ item.1.filename, item.1, head(map:lookup(item.2.sourceGrammar, r.compiledGrammars)), item.2),
-      accessor(r)),
+      accessor(r)) ++
+    flatMap(\def::Def ->
```
krame505, line 206: Should probably also add a comment here that we are adding the declaration sites of all global defs as reference locations here, as a shortcut instead of adding aspects for them in `*RefLocs`.

grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -170,31 +170,118 @@ aspect attributeRefLocs on StrategyExpr using := of
 | totalRef(a) -> if attrDclFound then [(a.location, attrDcl)] else []
 end;
 
+-- Productions
```
alexandra-hanson, line 173: Wondering if it makes sense to split all the aspects into another file?

grammars/silver/compiler/langserver/ReferenceLocations.sv
alexandra-hanson: Done in [e884604](https://github.com/melt-umn/silver/pull/736/commits/e884604aaabce4eca736c22c087c1424d93cece1)

grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -170,31 +170,118 @@ aspect attributeRefLocs on StrategyExpr using := of
...
+  top.allValueRefs = buildAllRefs((.valueRefLocs), g.grammarList);
+  top.allTypeRefs = buildAllRefs((.typeRefLocs), g.grammarList);
+  top.allAttributeRefs = buildAllRefs((.attributeRefLocs), g.grammarList);
 }
 
 function buildFileRefs
+annotation sourceLocation occurs on a,
 annotation sourceGrammar occurs on a =>
-map:Map<String (Location, Decorated RootSpec, a)> ::= accessor::([(Location, a)] ::= Decorated RootSpec) rs::[Decorated RootSpec]
+  map:Map<String (Location, Decorated RootSpec, a)> ::=
```
krame505, line 239: Style nit, the first line of the signature with the return type shouldn't be indented - just the parameters that spill onto additional lines.

grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -256,3 +343,26 @@ function findDeclLocation
...
+-- Returns a list of all reference locations
+-- Input is filename, line & col number, & decl map to resolve the symbol
+-- Uses refs map to lookup the reference paths & locations from the symbol fullName  
+function lookupReferenceLocations
+attribute fullName {} occurs on a =>
+[Location] ::= fileName::String line::Integer col::Integer decls::map:Map<String (Location, Decorated RootSpec, a)> refs::map:Map<String (String, Location)>
+{
+  return flatMap(\ item::(Decorated RootSpec, a) -> 
+    map( \loc::(String, Location) -> updateLocPath(loc.1, loc.2), 
+    map:lookup(item.2.fullName, refs)),
```
krame505, line 357: The indentation here is confusing, since the `map:lookup(...)` is a parameter in the call to `map` in the previous line but is indented the same.

grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -170,31 +170,118 @@ aspect attributeRefLocs on StrategyExpr using := of
...
+-- Productions
+-- LHS
+aspect valueRefLocs on top::ProductionLHS using <- of
+| productionLHS(id, _, _) -> map(\dcl :: ValueDclInfo -> (id.location, dcl), getValueDcl(id.name, top.env))
+end;
+
+aspect valueRefLocs on top::AspectProductionLHS using <- of
+| aspectProductionLHSTyped(id, _, _) -> map(\dcl :: ValueDclInfo -> (id.location, dcl), getValueDcl(id.name, top.env))
+| aspectProductionLHSId(id) -> map(\dcl :: ValueDclInfo -> (id.location, dcl), getValueDcl(id.name, top.env))
+end;
```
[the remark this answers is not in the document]
krame505: We definitely just want to report references that are actually referring to that specific binding.

Hmm, the problem is that the "full name" of a signature name binding isn't unique.  I think the fix for this is to
* Put the `*DclInfo` as the values in the `all*Refs`  map attributes and the result of `buildAllRefs`, instead of the `(String, Location)` pair.
* In `lookupReferenceLocations`, filter the result of the lookup from `refs` to just the items that have the same `sourceGrammar`/`sourceLocation` as `item.2`
* Compute the location with the full path in `lookupReferenceLocations` instead of when we build the refs map.

grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -170,31 +170,118 @@ aspect attributeRefLocs on StrategyExpr using := of
...
 end;
 
+-- Productions
+-- LHS
+aspect valueRefLocs on top::ProductionLHS using <- of
+| productionLHS(id, _, _) -> map(\dcl :: ValueDclInfo -> (id.location, dcl), getValueDcl(id.name, top.env))
+end;
+
+aspect valueRefLocs on top::AspectProductionLHS using <- of
+| aspectProductionLHSTyped(id, _, _) -> map(\dcl :: ValueDclInfo -> (id.location, dcl), getValueDcl(id.name, top.env))
```
krame505, line 180: These are both forwarding productions - you just need a single aspect for `aspectProductionLHSFull`.

grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -170,31 +170,118 @@ aspect attributeRefLocs on StrategyExpr using := of
...
+| aspectDefaultProductionSignature(_,_,t,_) -> t.typeRefLocs
+end;
+
+--RHS
+aspect valueRefLocs on top::ProductionRHSElem using <- of
+| productionRHSElem(id, _, _) -> map(\dcl :: ValueDclInfo -> (id.location, dcl), getValueDcl(id.name, top.env))
+end;
+
+aspect valueRefLocs on top::AspectRHSElem using <- of
+| aspectRHSElemTyped(id, _, _) -> map(\dcl :: ValueDclInfo -> (id.location, dcl), getValueDcl(id.name, top.env))
```
krame505, line 204: Ditto above, the aspect should be on `aspectRHSElemFull`

grammars/silver/compiler/langserver/ReferenceLocations.sv
```diff
@@ -170,31 +170,118 @@ aspect attributeRefLocs on StrategyExpr using := of
...
+
+aspect typeRefLocs on ProductionLHS using <- of
+| productionLHS(_, _, t) -> t.typeRefLocs
+end;
+
+aspect typeRefLocs on AspectProductionLHS using <- of
+| aspectProductionLHSTyped(_, _, t) -> t.typeRefLocs
+end;
+
+attribute typeRefLocs occurs on AspectDefaultProductionSignature;
```
krame505, line 192: Maybe just move this occurs-on/propagate to the big list at the top of this file.

grammars/silver/compiler/langserver/ReferenceLocations.sv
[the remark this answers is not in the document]
krame505: Also, if we support finding references for signature items, we probably also want to add aspects for locals/prod attrs.

krame505 (approved):

PR #740, opened 2023-03-16
Make fetch-jars default to the latest commit on the current branch that has jars

krame505:
# Changes
The most common workflow when switching to a new-jars branch is to run `./fetch-jars HEAD`.  However this fails if the latest commit on the branch doesn't build, or the build hasn't finished yet, etc. in which case one needs to walk back to find the most recent commit.  

This changes the default behavior when `./fetch-jars` is run without arguments to find the most recent commit on the branch that has jars.    If the chosen commit matches develop, the stable jars are downloaded as before.  An explicit revision (e.g. `./fetch-jars develop`) can also still be specified.

This change is also needed for new-jars branches to work with Github's new merge queue feature, since that works by creating new temporary branches, and we must determine the appropriate jars to use from when the commit was previously built.  Technically, this also obsoletes some of the logic in Silver's Jenkinsfile for getting jars from a previous build's artifacts, but I'm leaving that for now as it still helpfully annotates which jars were used for the build. Although we could probably pull the logic for choosing a commit out into a separate script, and just have Jenkins use that instead for annotating the build.  

# Documentation
I updated the comments in the script.

### DIFF
--- a/fetch-jars
+++ b/fetch-jars
@@ -20,18 +20,15 @@ if [[ $* != *--copper* && $# -gt 0 || $# -gt 1 ]]; then
   rev=$(git rev-parse $1)
 fi
 
+# Get the hash of the latest commit on develop.
+# This always gives the latest commit regardless of if the repo is up to date,
+# and works in a Jenkins checkout.
+DEVELOP=$(git ls-remote https://github.com/melt-umn/silver develop | cut -f1)
+
 # Look for jars in the current commit and its parents.
 # First, check that we are inside a git repo.
-if git rev-parse --is-inside-work-tree 1> /dev/null 2> /dev/null && [[ -z $rev ]]; then
-  # List all cantidate commits
-  if git rev-parse develop 1> /dev/null 2> /dev/null; then
-    # We see develop, only list commits not on develop
-    commits=$(git rev-list HEAD ^develop)
-  else
-    # We don't see develop (checked out by Jenkins?), list all commits
-    commits=$(git rev-list HEAD)
-  fi
-  for commit in $commits; do
+if [[ -z $rev ]] && git rev-parse --is-inside-work-tree 1> /dev/null 2> /dev/null; then
+  for commit in $(git rev-list HEAD ^$DEVELOP); do
     if has_jars $commit; then
       echo "Found jars from past commit"
       rev=$commit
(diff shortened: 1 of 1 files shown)


remexre:
Might be worth running shellcheck; there are a bunch of warnings, mostly around quoting.

fetch-jars
```diff
@@ -3,26 +3,62 @@
...
+COMMIT_ARTIFACTS="https://foundry.remexre.xyz/commit-artifacts/"
 
+function has_jars {
+  wget --spider -q "$COMMIT_ARTIFACTS/$1"
+}
+
+rev=""
+
+# Look for jars in the specified revision
 if [[ $* != *--copper* && $# -gt 0 || $# -gt 1 ]]; then
```
remexre, line 19: Is this check right? seems like it'd ignore the rev if you did `./fetch-jars v0.4.4 --copper`

remexre: ...except that it's missing parens too, I think?

fetch-jars
```diff
@@ -3,26 +3,62 @@
...
-    rev=$(git rev-parse $1)
-    echo "Warning: Fetching unstable jars!  (commit $rev)"
+  rev=$(git rev-parse $1)
```
remexre, line 20: Is `./fetch-jars --copper v0.4.4` intended to be supported? might make more sense to do to parse the args and provide usage up-front, then do the rest of the script...... or rewrite it in python and use argparse lol

fetch-jars
```diff
@@ -3,26 +3,62 @@
...
+# This always gives the latest commit regardless of if the repo is up to date,
+# and works in a Jenkins checkout.
+DEVELOP=$(git ls-remote https://github.com/melt-umn/silver develop | cut -f1)
+
+# Look for jars in the current commit and its parents.
+# First, check that we are inside a git repo.
+if [[ -z $rev ]] && git rev-parse --is-inside-work-tree 1> /dev/null 2> /dev/null; then
+  for commit in $(git rev-list HEAD ^$DEVELOP); do
+    if has_jars $commit; then
+      echo "Found jars from past commit"
```
remexre, line 33: Maybe log which one it is, at least if it's not HEAD

fetch-jars
krame505: `./fetch-jars v0.4.4 --copper` should work? Since the number of args is greater than 1. I didn't update this logic, though.

fetch-jars
krame505: That gets printed with git show on line 46

fetch-jars
krame505: Yeah that doesn't work at the moment. I don't really feel like rewriting it in Python, and bash getopts is ugly/feels like overkill. I might try to fix that issue though.

remexre (approved):

PR #742, opened 2023-03-21
Add --unstable option to fetch-jars, to force fetching commit artifacts

krame505:
# Changes
When --unstable is specified, always fetch the jars from foundry.  This is useful when debugging issues with deployment, in case the stable jars on the website and the jars from the latest commit on develop don't match.

# Documentation
Updated the comment in the script.

### DIFF
--- a/fetch-jars
+++ b/fetch-jars
@@ -10,7 +10,7 @@ set -eu
 
 # Parse command line argumants.
 # Hacky, but getopts seems like overkill for this...
-rev_name=develop
+rev_name=""
 for arg in "$@"; do
   if [[ $arg != -* ]]; then
     rev_name=$arg
--- a/fetch-jars
+++ b/fetch-jars
@@ -2,20 +2,21 @@
 
 set -eu
 
-# Usage: ./fetch-jars [rev-name] [--copper]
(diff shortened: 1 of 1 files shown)


fetch-jars
```diff
@@ -26,8 +27,8 @@ function has_jars {
 rev=""
 
 # Look for jars in the specified revision
-if [[ -n $REV_NAME ]]; then
-  rev=$(git rev-parse "$REV_NAME")
+if [[ -n $rev_name ]]; then
```
remexre, line 30: doesn't this fail shellcheck? if I call `./fetch-jars "x || 1 = 1"` does that break things?

krame505: shellcheck isn't complaining.  As I would expect that fails with
```
fatal: ambiguous argument 'x && 0 = 1': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

remexre: Oh, weird, first time I've ever experienced shellcheck not complaining _enough_ about a variable without `"`'s around it lol

idk, if we care, put `"`s around the vars; if not, I guess a user who puts spaces in a ref name is kinda asking for trouble anyway...

remexre (approved):

PR #745, opened 2023-03-24
Add support for threaded collection attributes

### LINKED ISSUE #744: Threaded collection attributes
@remexre suggested a use case for threaded collection attributes, for assigning unique identifiers.  Currently one would need to write something like
```
threaded attribute idIn, idOut :: Integer occurs on Expr;
production foo
{
  e1.idIn = top.idIn + 1;
  thread idIn, idOut on e1, e2, top;

  local id::Integer = top.idIn;
}
```
With a collection attribute, one could instead specify a global propagate and write
```
threaded attribute idIn, idOut :: Integer with + occurs on Expr;
propagate idIn, idOut on Expr;
production foo
{
  e1.idIn <- 1;

  local id::Integer = top.idIn;
}
```
This pattern may be useful in other sorts of analyses where we would like to collect things that come before a certain node in the tree.

I'm not really sure about the syntax.  Note that we don't need an empty value, as with monoid attributes.  
```
threaded attribute idIn, idOut :: Integer with +;
```
would be consistent with the existing collection attribute syntax.  It might be nice to make the append operation implicit if the attribute type has a `Semigroup` instance (e.g. lists), as with monoid attributes.  But
```
threaded monoid attribute things :: [Thing];
```
isn't any less verbose than just writing
```
threaded attribute things :: [Thing] with ++;
```
So I'm leaning towards the latter syntax.

krame505:
# Changes
Fixes #744.

# Documentation
https://github.com/melt-umn/melt-website/pull/50

### DIFF
--- a/grammars/silver/compiler/extension/autoattr/DclInfo.sv
+++ b/grammars/silver/compiler/extension/autoattr/DclInfo.sv
@@ -161,45 +161,71 @@ top::AttributeDclInfo ::= inh::String synPartial::String syn::String
 }
 
 abstract production threadedInhDcl
-top::AttributeDclInfo ::= inh::String syn::String bound::[TyVar] ty::Type rev::Boolean
+top::AttributeDclInfo ::= inh::String syn::String bound::[TyVar] ty::Type o::Maybe<Operation> rev::Boolean
 {
   top.fullName = inh;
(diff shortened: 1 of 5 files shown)


ericvanwyk (approved):
Looks great.

PR #747, opened 2023-04-03
Properly check that implicit forward copy eqs for host syn attributes don't exceed flow types

### LINKED ISSUE #746: Forward inh equations can cause excess dependencies missed by MWDA
Consider the following code:
```
inherited attribute env1, env2 :: [String];
synthesized attribute errors :: [Message];
nonterminal Expr with env1, env2, errors;
flowtype Expr = forward {env1, env2}, errors {env1};

production fwrdEqSiteExceedsFT
top::Expr ::= e::Expr
{
  forward.env1 = top.env2;
  forwards to e;
}
```
Here the forward inh equation means that `forward.env` depends on `top.env2`.  This creates a transitive dependency of `top.errors` on `top.env2`, which exceeds the flow type, however this isn't currently caught by the flow analysis.  

Currently, we just check for forward inh equations that $deps(\text{forward}.i)$ don't exceed $FT[\text{forward}] \cup \lbrace top.i \rbrace$.  (Cool, LaTeX works in github markdown?)  However this breaks for synthesized attributes $s$ that depend on $i$ where $FT[s] \subset FT[\text{forward}]$.

I'm not positive what the solution is, at the moment.  Might be as simple as seeding the flow graph with $deps[\text{forward}.i]$?  

I just noticed there's currently a TODO comment on `forwardInh`, `TODO: we need to figure out how to introduce any new lhsinh deps to the forward flow type automatically.`

krame505:
# Changes
Fix #746. Ensure that the transitive deps of every host syn attribute on the forward don't exceed their flow types, rather than checking that every host syn flowtype is a subset of the forward dependencies.

# Documentation
Updated comment, added a test.


ericvanwyk (approved):
LGTM

PR #748, opened 2023-04-03
Replace ExprVertexInfo with Maybe<VertexType>

krame505:
# Changes
Some basic refactoring.  `ExprVertexInfo` is basically a `Maybe` specialized to `VertexType`.  As best as I can tell, this was just done for legacy reasons back when we had textual interface files and attributes for serialization on everything.  Having it as a `Maybe` is neater, especially since I will be introducing a second sort of flow vertex associated with expressions in the future.

# Documentation
Updated comments.


ericvanwyk (approved):
Looks good. Sorry I missed this earlier. All looks quite straightforward.

Issue #555, opened 2021-09-23
Make type signatures on lambda parameters optional


krame505:
Currently we require explicit type signatures on all lambda arguments.  While in some cases this helps readability, at times always writing these can make code more cumbersome, e.g. https://github.com/melt-umn/ableC-string/blob/32bad7651e28f53b565904940b2d92f58ed9e0cf/grammars/edu.umn.cs.melt.exts.ableC.string/abstractsyntax/Type.sv#L56  Also note that lambda parameter type signatures are optional in most other functional languages, e.g. Haskell.

Currently lambdas reuse the `ProductionRHS` nonterminal for their parameter list.  Making this change would involve duplicating that nonterminal in the lambda modification, and adding a production with no type signature that uses a fresh type variable as the type.  

We may also want to support `_` in lambda signatures, as is allowed in Haskell and also in Silver aspect productions.

alexandra-hanson:
@krame505 I have looked at `ProductionRHS`, and the `freshType()` function and what you have described sounds simple enough. It looks like `freshType()` constructs a fresh type variable of kind star. One thing I am wondering that is maybe a dumb question: it seems that there would be cases where the lambda argument is a type of kind arrow, and not just of kind star.. if that's the case, I probably wouldn't want to use the `freshType()` function, right? Or am I thinking about this the wrong way/misunderstanding something?

PR #750, opened 2023-04-29
Refactor doc extension error handling to avoid flow error

krame505:
# Changes
feature/dec-site-projections includes some bug fixes to flow analysis for collection contribution equations.  This fixes a flow error that was found as a result of that, where the contribution equation for `errors` on `root` adds transitive deps on `top.downDocConfig`.  

Instead, we need to collect doc errors further up the tree, and add a way to inject them at the `RootSpec` level.

# Documentation
This fixes a flow bug in the documentation extension.

### DIFF
--- a/grammars/silver/compiler/driver/util/RootSpec.sv
+++ b/grammars/silver/compiler/driver/util/RootSpec.sv
@@ -129,9 +129,15 @@ top::RootSpec ::= g::Grammar  oldInterface::Maybe<InterfaceItems>  grammarName::
   top.declaredName = g.declaredName;
   top.moduleNames := nub(g.moduleNames ++ ["silver:core"]); -- Ensure the prelude is in the deps, always
   top.allGrammarDependencies := actualDependencies;
-  top.grammarErrors = g.grammarErrors;
+
+  top.grammarErrors = filter(\ fe::(String, [Message]) -> !null(fe.2), top.allFileErrors);
   top.parsingErrors = [];
-  top.allFileErrors = g.allFileErrors;
+
+  production attribute extraFileErrors::[(String, [Message])] with ++;
+  extraFileErrors := [];
+  top.allFileErrors = map(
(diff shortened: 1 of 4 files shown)


grammars/silver/compiler/driver/util/RootSpec.sv
```diff
@@ -129,9 +129,15 @@ top::RootSpec ::= g::Grammar  oldInterface::Maybe<InterfaceItems>  grammarName::
...
-  top.grammarErrors = g.grammarErrors;
+
+  top.grammarErrors = filter(\ fe::(String, [Message]) -> !null(fe.2), top.allFileErrors);
   top.parsingErrors = [];
-  top.allFileErrors = g.allFileErrors;
+
+  production attribute extraFileErrors::[(String, [Message])] with ++;
+  extraFileErrors := [];
+  top.allFileErrors = map(
+    \ fe::(String, [Message]) -> (fe.1, fe.2 ++ concat(lookupAll(fe.1, extraFileErrors))),
```
ericvanwyk, line 139: It would be nice if we did not use this `.1` and `.2` form. Isn't a `case` better so that names can be given to these pieces? Or do we have a `let` binding for irrefutable patterns so that one could write `let (filename, errs) = fe in ...`   Or better `\ (filename,errs) :: (String, [Message]) -> ...`

ericvanwyk (approved):
LTGM if the requested comments are added. Thoughts on the comment about using `case` instead of `.1` and `.2`.  As Silver gets more complicated it seems we need even more self-explanatory code.

grammars/silver/compiler/driver/util/RootSpec.sv
krame505: We don't have irrefutable pattern syntax for let/lambdas, but we probably should.  #555 is probably a prerequisite for that.

Issue #751, opened 2023-05-01
Simplifying Silver's possibly-decorated type inference semantics


krame505:
The current state of affairs is that whether a reference to a decorable child or local[^local] gets a decorated type is inferred through a "fun trick" in the type system.  This can be confusing[^decinf], and we also make it far too easy IMO to un-decorating a tree that has already been decorated.  Since we want to make only decorating something in one place the new normal, one should need to be explicit by writing `new` when they actually do wish to copy a tree.

[^local]: By local I really mean local or production attribute in this writeup.
[^decinf]: Citation: @remexre complaining about "decoration inference", which is really just part of the type system and not its own thing, but I digress.

# Background
As a quick refresher, a type is decorable when it is:
* a nonterminal type
* a unique reference type
* a type constant with a occurs-on type constraint.

The "base type" of a decorable type is the `nt` in `Decorated! nt with i` for a unique reference type, or else just the nonterminal/type constant.

The way it currently works is that a use of a child or local declared with a reference type gets a possibly-decorated type (`ntOrDecType` is the name of the production, but it doesn't have syntax or appear in error messages.)  A `ntOrDecType` for base type `nt` is permitted to unify with `nt`, `Decorated nt with i` or `Decorated! nt with i`, and after unification the type gets replaced with what it was specialized to - the `ntOrDecType` constructor has an extra hidden type variable to track this.  `ntOrDecType`s that are still unspecialized after unification default to `Decorated nt with {}`.  

This does pose a problem when we write e.g. `x == y` where `x` and `y` are both possibly-decorated `nt`s, where there is an instance for `Eq nt`.  Since `eq :: (Boolean ::= a a)` is polymorphic, we never specialize `x` and `y` to be undecorated, and get the default `Decorated nt with {}`, which doesn't have an `Eq` instance.  As a gross workaround for this, I made function application check if the applied function type has any type constraints that are only satisfiable if specialized to an undecorated type, and if so does the specialization.  

We also have special semantics for "lexical local variables", which are anything bound in a let/pattern match.  Anything bound with a decorated type is permitted to implicitly undecorate, meaning these also get an `ntOrDecType` type.  We do want to disallow casting between `Decorated` types with different sets of inherited attributes, thus the `ntOrDecType` constructor also has an `InhSet` type member to control what the reference set should effectively be when unifying with a `Decorated`/`Decorated!` type.

[...]

PR #753, opened 2023-05-09
Add support for decoration site projections

krame505:
# Changes
This adds support for decoration-site flow projections.  

The major additions required for this are:
* A new inherited attribute `decSiteVertexInfo :: Maybe<VertexType>`, containing the flow vertex corresponding to the decoration site where this expression will ultimately be decorated.
* A new inherited attribute `alwaysDecorated`, determining whether this expression is decorated unconditionally.  Note that there are cases where an expression has a known decoration site, but we cannot rely on it always being decorated. 
 For example in `forwards to if cond then @foo else errorExpr(...);` we can't project inherited equations supplied to `foo` through forwarding, but we can say that `foo.env` may depend on `top.env` when checking for hidden transitive dependencies.
* A new `FlowVertex`/`VertexType`/`FlowDef`, `subtermVertex`/`subtermVertexType`/`subtermDecEq`, for expressions that occur as decorable arguments to a production call that has a `decSiteVertexInfo`.
* New `FlowDef`s for tracking unique references to children/locals in places that have a `decSiteVertexInfo`.
* Added flow checks for supplying explicit equations to a child/local that has a unique ref decoration site, to avoid introducing hidden transitive dependencies that exceed the dependencies of the projected equation.
* Changes to the translation for taking non-unique references to a child or local that has a unique reference taken, where we might be relying on the child/local having been decorated there first.

The actual projection machinery in the flow inference driver was already there for pattern var flow projections, and could be reused pretty much unchanged.  

See https://github.com/melt-umn/silver/compare/feature/dec-site-projections...feature/use-dec-site-projections for some changes using this feature in the Silver compiler; this will be submitted in a subsequent pull request.  

# Documentation
This feature isn't documented yet.  I'm unsure how to proceed here, as an explanation of this is tied up in a larger discussion of unique references (which may be subject to change with #751) and approaches to forwarding.  This will be laid out in my thesis proposal/a potential SLE paper, but I'm not sure if I should write something up for the website before merging this.

Comments have been added in the implementation as needed.


ericvanwyk (approved):
I cannot say that I follow all of this. But it is OK to merge.

PR #754, opened 2023-05-13
Use decoration site projections for misc cleanup

krame505:
# Changes
Make use of decoration site projections to avoid writing some annoying inherited equations.  

# Documentation
Added a couple comments.


ericvanwyk (approved):
Looks good.

PR #758, opened 2023-06-10
Avoid introducing superfluous append when collection antiquote is the last in a list

krame505:
# Changes
When an a collection antiquote is the last item in the list, avoid generating an append operation.  For example:
```
ableC_Decl {
  int foo(int a, float b, $Parameters{params}, int c, $Parameters{moreParams}) {
    ...
    int res = foo(1, 2, $Exprs{args}, 3, ${moreArgs});
    ...
  }
}
```
With `params` and `args` above, we need to use `appendParams` and `appendExprs` to append extra items to the antiquoted expressions, but with `moreParams` and `moreArgs` we can just place them end of the cons chain instead of `nilParameters()`/`nilExpr()`.

This change is needed in my experiments with using decoration-site expressions in ableC extensions, since functions like `appendParameters` transform/re-decorate the operands and aren't currently compatible with decorated tree sharing.

# Documentation
Corresponding changes in silver-ableC, and docs updated there: https://github.com/melt-umn/silver-ableC/compare/feature/collection-antiquote-tail.

### DIFF
--- a/grammars/silver/compiler/metatranslation/Translation.sv
+++ b/grammars/silver/compiler/metatranslation/Translation.sv
@@ -76,8 +76,8 @@ top::AST ::= prodName::String children::ASTs annotations::NamedASTs
   
   -- "Collection" antiquote productions
   -- Key: antiquote production name
-  -- Value: (nonterminal short name, cons production name, append production name)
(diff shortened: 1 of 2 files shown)


ericvanwyk (approved):
looks ok to me

PR #759, opened 2023-06-10
Fix bug with check for missing remote equations

krame505:
# Changes
This fixes an issue with chaining multiple levels of decoration through references across multiple productions:
```
production foo
top::Expr ::= e::Expr
{
  top.errors = ... e.env ...;
  forwards to bar(@e);
}

production bar
top::Expr ::= e::Expr
{
  forwards to baz(@e);
}

production baz
top::Expr ::= e::Expr
{
  e.env = top.env;
}
```
`foo` depends on `e.env`, but this equation is in `baz`, not `bar`.  We need to find all decoration sites for `e`, and ensure that there is an `env` equation at any of them.  

# Documentation
Added source comments and a test case.


krame505:
I discovered that there is a related issue in the translation for this scenario that requires a bit of refactoring - will be another PR shortly.

PR #761, opened 2023-06-14
Implement forward production attributes

krame505:
# Changes
This adds support for forward production attributes, as proposed previously.  For example
```
production listComp
top::Expr ::= n::String l::Expr res::Expr cond::Expr
{
  local localErrors::[Message] =
     l.errors ++ res.errors ++ cond.errors ++
     case l.typerep of
     | listType(_) -> []
     | _ -> [err(l, "Input expected a list")]
     end ++
     case cond.typerep of
     | boolType() -> []
     | _ -> [err(cond, "Condition expected a bool")]
     end;
  
  forward fwrd =
    stmtExpr(
      seqStmt(
        varDecl("result", emptyList(res.typerep)),
        forEach(n, @l, ifStmt(@cond, appendList("result", @res)))),
      var("result"));

  forwards to if null(localErrors) then @fwrd else errorExpr(localErrors);
}
```
Here we want to decorate the children through forwarding, but conditionally forward to an error production if there are local errors.  Instead we can decorate the forward tree under a local production attribute.  Forward production attributes avoid the need to write copy equations for all the inherited attributes on `fwrd` by decorating the prod attr with the current node as its forward parent.

(I thought there was a github issue for this opened, but maybe not?)

# Documentation
The above example and some explanation are included in [this writeup](https://github.umn.edu/krame505/thesis-proposal/blob/main/decoration-and-forwarding.tex), and we may want to mention forward prod attrs in the forthcoming paper on forwarding.  This should eventually all get written up for the website docs, too.

### DIFF
--- a/grammars/silver/compiler/analysis/typechecking/core/ProductionBody.sv
+++ b/grammars/silver/compiler/analysis/typechecking/core/ProductionBody.sv
@@ -3,7 +3,7 @@ grammar silver:compiler:analysis:typechecking:core;
 
 attribute upSubst, downSubst, finalSubst occurs on ProductionStmt, ForwardInhs, ForwardInh, ForwardLHSExpr;
 propagate upSubst, downSubst on ProductionStmt, ForwardInhs, ForwardInh, ForwardLHSExpr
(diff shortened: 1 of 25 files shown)


ericvanwyk (approved):
LGTM

PR #763, opened 2023-06-20
Finish long-standing cleanup splitting flow vertices into syn/inh

krame505:
# Changes
FInish some cleanup to flow vertices that Ted started a while back, before making some big changes for translation attributes.

Also add an option to dump out the graph of grammar exports, needed for debugging an MWDA issue.  

# Documentation
Updated doc comments.

### DIFF
--- a/grammars/silver/compiler/translation/java/driver/BuildProcess.sv
+++ b/grammars/silver/compiler/translation/java/driver/BuildProcess.sv
@@ -96,6 +96,9 @@ top::Compilation ::= g::Grammars  _  buildGrammars::[String]  benv::BuildEnv
   production attribute keepFiles :: [String] with ++;
   keepFiles := [];
 
+  -- Seed flow deps with {config}
+  keepFiles <- if false then error(hackUnparse(top.config)) else [];
+
   top.postOps <-
     [genBuild(buildXmlLocation, buildXml)] ++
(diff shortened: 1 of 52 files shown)


krame505:
Hmm, I guess this fixes a bug in the flow analysis, where we weren't checking for equations for inh attributes that aren't in scope.

krame505:
Ugh, this does seem to make the Silver compiler's memory usage a bit worse...

grammars/silver/compiler/translation/java/driver/BuildProcess.sv
```diff
@@ -96,6 +96,9 @@ top::Compilation ::= g::Grammars  _  buildGrammars::[String]  benv::BuildEnv
   production attribute keepFiles :: [String] with ++;
   keepFiles := [];
 
+  -- Seed flow deps with {config}
+  keepFiles <- if false then error(hackUnparse(top.config)) else [];
```
remexre, line 100: debug code? or was this part of some bootstrapping thing?

remexre (approved):

grammars/silver/compiler/translation/java/driver/BuildProcess.sv
krame505: Seeding the flow deps, we should have  nicer syntax for this but not an issue.

PR #764, opened 2023-06-28
Add a top-level script to clean rebuild the silver language server and package the VS code extension

krame505:
# Changes
Add a script to make clean rebuilding and packaging the language server VS code extension easier for local development.  Also use it in the Jenkinsfile.

# Documentation
Added a comment in the script.


make-silverlsp-vsix
```diff
@@ -0,0 +1,33 @@
...
+# and VS Code extension.
+
+set -eu
+
+SILVERLSP_VSIX="silverlsp-latest.vsix"
+
+if [ $0 != "./make-silverlsp-vsix" ]; then
+  echo "Run as ./make-silverlsp-vsix"
+  exit 1
+fi
```
remexre, line 13: any reason not to do `cd "$(dirname "${BASH_SOURCE[0]}")"` instead?

remexre (approved):

make-silverlsp-vsix
krame505: Ehh I just copied that bit from all the other top-level utility scripts.  I could change it I guess?

PR #767, opened 2023-07-14
Error immediately on cycles in attribute evaluation

krame505:
# Changes
This tracks whether an attribute has been demanded on a `DecoratedNode`, to immediately report an error when encountering a cycle, rather than crashing with a `StackOverflowError`.  

# Documentation
Added doc comments.


krame505:
Turns out that checking for a cycle on every thunk is actually more efficient than doing it for syn/inh accesses on DecoratedNode. This should also catch cycles that don't involve attribute accesses (e.g. `local x::Expr = neg(x);`.)

runtime/java/src/common/Thunk.java
```diff
@@ -25,11 +31,37 @@ public Thunk(final Evaluable<T> e) {
...
+	@SuppressWarnings("unchecked")
+	private void doEval() {
+		demanded++;
+		if(demanded > 1) {
+			handleCycleError();
+		}
+		try {
+			o = ((Evaluable<T>)o).eval();
+		} catch(Throwable t) {
+			traceCycleError(t);
```
remexre, line 47: eh, i don't _love_ the name `traceCycleError` when it handles all throwables... Though, the logic here feels weird; shouldn't it just be catching a `CycleException` then reraising a `CycleTraceException` if `demanded>1`? (and in particular, not catching e.g. an IOException in every thunk.)

krame505: We would get a `CycleException` wrapped in a whole bunch of `TraceException`s here, not just a raw `CycleException`.   The convention seems to be to catch any `Throwable` and wrap it in a `TraceException`, so one might have an `IOException` inside a `TraceException` anyway.  But the only way that `demanded > 1` is if we already threw a `CycleException`, so we don't need to look at the root cause.

remexre: hm, okay; my gut still says it'd be better to be more specific in what we catch (L62 doesn't spark joy), but if stuff's already getting wrapped by TraceExceptions anyway, I guess it doesn't matter.

...now I'm kinda wondering if we might be better off using the stack inspection machinery to collect traces instead... that's something to think about a different day, though.

remexre (approved):

runtime/java/src/common/Thunk.java
krame505: Yeah, agreed.

I was wondering why we catch all `Throwable`s rather than just `RuntimeException`s - but I guess `StackOverflowError` isn't a runtime exception and we want to get a trace for that.

PR #772, opened 2023-07-20
Avoid build congestion by waiting for a free executor to spawn downstream builds

krame505:
# Changes
This should (hopefully) fix the issue with multiple Silver builds started close to the same time causing all executor slots to get filled with builds waiting for other downstream builds to complete.

# Documentation
Added comments in Jenkinsfile and jenkins-lib.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -101,6 +101,8 @@ melt.trynode('silver') {
     sh "./self-compile --clean --mwda --dont-translate"
   }
 
+  waitUntil { melt.isExecutorAvailable() }
+
   stage("Test") {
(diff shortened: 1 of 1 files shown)


remexre (approved):
is there still a race here, where multiple builds try to claim the same executor? Still, if this empirically fixes it for now, fine

krame505:
Yeah there still might technically be a race here, but `waitUntil` backs off to poll every 15 seconds, so the chances of multiple builds starting their downstream phase at once are small.  And in the worst case we fall back to the previous behavior, of just having downstream tasks from multiple builds in the queue.

PR #769, opened 2023-07-18
Implement data nonterminals

krame505:
# Changes
Fixes #556.

Depends on #768. 

# Documentation
TBD

PR #773, opened 2023-07-24
Use data nonterminals in the Silver compiler

krame505:
# Changes
Depends on #769. 

Change a number of data structures in the standard library and various portions of the compiler to be data nonterminals.

# Documentation
Updated comments.


grammars/silver/compiler/definition/type/Unification.sv
```diff
@@ -109,7 +109,7 @@ top::Type ::= fn::String ks::[Kind] data::Boolean tracked::Boolean
   top.unify = 
     case top.unifyWith of
     | nonterminalType(ofn, oks, odata, otracked) ->
-        if fn == ofn && data == odata && tracked == otracked  -- Mismatched data/tractness can happen when comparing interface files
+        if fn == ofn --&& data == odata && tracked == otracked  -- Mismatched data/tractness can happen when comparing interface files
```
ericvanwyk, line 112: Change this back?

ericvanwyk (approved):
Very nice

grammars/silver/compiler/definition/type/Unification.sv
krame505: Will do in a future PR, see #804

Issue #556, opened 2021-09-27
Datatypes


remexre:
Datatypes, which are not nonterminals (like lists kinda sorta are).

- No forwarding
- No attributes (maybe reuse attribute syntax for UFCS?)
- No decoration
- Maybe Haskell-like syntax?

krame505:
What does UFCS mean?  

I could maybe support an extension for more concisely defining a nonterminal and productions with no attributes with Haskell-like syntax, that just forwards to normal nonterminal/production decls.  But anything more than that seems like too much added complexity to be worthwhile.

ericvanwyk:
Do define UFCS.

Would a "closed and we mean it" notion on nonterminal work?  This would be one where when you define, say, `List<a>` you can say it is closed so that no additional attributes from outside of this  grammar can be added.  If one doesn't define any attributes for it, then there are no attributes for it.

remexre:
UFCS would basically be, we forward `x.f(y, z)` to `f(x, y, z)`.

I think the "it doesn't make sense to decorate this" is definitely part of what I want. "closed and we mean it" might work too, but even having built-in attributes feels potentially weird for e.g. a red-black tree -- the memory usage of operations on it should be stack usage, not heap usage.

krame505:
If the goal is improving performance for specialized data structures, might I suggest another version of pattern matching that ignores forwarding?  

Introducing another category of nonterminals with different restrictions seems undesirable to me.  I don't see why the ability to decorate a `List<a>` or `Tree<k v>` with new attributes is particularly bad, if the commonly used operations were implemented efficiently as functions.  

https://en.wikipedia.org/wiki/Uniform_Function_Call_Syntax seems like a separate issue to me.  I could imagine maybe permitting it on nonterminals (we already overload dot notation for syn/inh/annotation...) but I would like to see some good example use cases first.

remexre:
> I would like to see some good example use cases first.

Silver with #555:

```
sum(map(\xyz -> xyz.1.1 * xyz.1.2 * xyz.2,
        zip(map(\x -> (x, x + 1), filter(is_prime, foo)), bar)))
```

Silver with UFCS:

```
foo.filter(is_prime)
   .map(\x -> (x, x + 1))
   .zip(bar)
   .map(\xyz -> xyz.1.1 * xyz.1.2 * xyz.2)
   .sum()
```

Rust:

```
foo.iter()
   .filter(is_prime)
   .map(|x| (x, x + 1))
   .zip(bar)
[...]

PR #774, opened 2023-07-24
Deriving mechanism for Eq/Ord

krame505:
# Changes
Add support for deriving for `Eq` and `Ord`.  This is needed for #556 since datatypes don't have inh attributes, so equality/ordering attributes are no longer an option.  Also this handles polymorphic nonterminals like `Maybe` or `Pair`, where additional constraints are needed on the generated instance, where equality/ordering attributes were not an option.

# Documentation
https://github.com/melt-umn/melt-website/pull/51, also added tests.

### DIFF
--- a/grammars/silver/compiler/extension/autoattr/Ordering.sv
+++ b/grammars/silver/compiler/extension/autoattr/Ordering.sv
@@ -80,7 +80,7 @@ top::ProductionStmt ::= inh::String keySyn::String syn::Decorated! QName
             else
               foldr1(
                 \ e1::Expr e2::Expr ->
-                  Silver_Expr { if $Expr{e1} == 0 then $Expr{e2} else $Expr{e1} },
+                  Silver_Expr { let res::Integer = $Expr{e1} in if res == 0 then $Expr{e2} else res end },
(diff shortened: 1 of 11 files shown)


grammars/silver/compiler/extension/autoattr/Ordering.sv
```diff
@@ -80,7 +80,7 @@ top::ProductionStmt ::= inh::String keySyn::String syn::Decorated! QName
             else
               foldr1(
                 \ e1::Expr e2::Expr ->
-                  Silver_Expr { if $Expr{e1} == 0 then $Expr{e2} else $Expr{e1} },
+                  Silver_Expr { let res::Integer = $Expr{e1} in if res == 0 then $Expr{e2} else res end },
```
remexre, line 83: Are there macro-hygiene-flavored issues here if a child is named `res`?

krame505: Shouldn't be, since the derived instances are purely generated code with all variable names determined here.

ericvanwyk (approved):

PR #775, opened 2023-07-27
Change Pair to use fst/snd annotations

krame505:
# Changes
Depends on #773. 

We discussed this change a while ago but never actually did it - this changes the definition of `Pair` from 
```
synthesized attribute fst<a> :: a;
synthesized attribute snd<a> :: a;
data nonterminal Pair<a b> with fst<a>, snd<b>;
production pair
top::Pair<a b> ::= x::a y::b
{
  top.fst = x;
  top.snd = y;
}
```
to
```
annotation fst<a> :: a;
annotation snd<a> :: a;
data nonterminal Pair<a b> with fst<a>, snd<b>;
production pair
top::Pair<a b> ::=
{}
```

This gives a slight performance improvement, as we don't need to create a `DecoratedNode` to access an annotation.  This is also a nice demo of Silver's support for "records" via annotations (which works out a bit nicer now with the addition of deriving support in #774.)  

# Documentation
https://github.com/melt-umn/melt-website/pull/52


krame505:
This is currently failing to build due https://github.com/melt-umn/silver/issues/776

ericvanwyk (approved):

ericvanwyk:
Can one still pattern match on pairs if annotations are used instead of regular child terms?

krame505:
Yes, one can still write 
```
case foo of
| pair(fst=42, y=a) -> ...
end
```
or
```
case foo of
| (42, a) -> ...
end
```
using tuple syntax.

ericvanwyk:
Of course - the tuples extension is what we'd use in most applications. Thanks.

PR #783, opened 2023-08-08
Add a couple misc utilities

krame505:
# Changes
Add list sum/product and instance MonadFail IO

# Documentation
Doc comments


remexre (approved):
LGTM; these should probably go on Foldable when we get that. (or just use a foldable instead of [])

krame505:
Yeah, there's a bunch of utils specialized to `[]` at the moment that could be `Foldable` eventually.

PR #794, opened 2023-09-01
Add heading for tests in pull request template

RandomActsOfGrammar:
# Changes
This adds a heading to pull requests requiring a description of what testing is written for the changes.  I think we have had problems in the past with tests not being written.  This change will bring tests to mind before any change is incorporated into Silver and help ensure the reviewer also considers the tests written, or the lack thereof, before approving.

# Documentation
This does not require any documentation because it will show up whenever someone tries to make a pull request and does not affect anything else.


ericvanwyk (approved):
Good idea!

krame505 (approved):

PR #785, opened 2023-08-09
Misc refactoring to origin tracking implementation and libraries

krame505:
Depends on #775 #790 

# Changes
* Add utilities for reflectively accessing attributes
* Add `genericPP` for pretty-printing arbitrary values.  This is a general-purpose improved replacement to `hackUnparse` for most purposes; this function uses the pp/unparse attributes when available, falling back to showing a term representation with reflection.
* Improved utilities for handling origin info.  This includes improved pretty-printing of origin info chains with origin notes.
* Add  `extensionGenerated` note to flag code generated by an extension that should not have errors, include diagnostic message with origin info chain when the note appears in origins for an error message.
* Add `logicalLocationFromOrigin` note to indicate that the logical location of a term is the parsed origin location of another term.
* Include locations from origin tracking and origin notes in stack traces
* Make some standard library nonterminals tracked, including regexes and term rewriting
* Refactor `OriginInfo` to use annotations for `originNotes` and `originType`
* Misc fix to `rebuild-runtime` script to work after a `deep-clean`.

# Documentation
Added doc comments to new utilities.

PR #803, opened 2023-09-22
Change the Silver compiler to use origin tracking

krame505:
Depends on #785

# Changes
Major overhaul of the Silver compiler to replace uses of the location annotation with origin tracking.

# Documentation
Added comments as needed.

# Testing
Hard to test, but verified for a number of error messages that error locations are in a reasonable place.


grammars/silver/compiler/analysis/warnings/flow/MwdaFlag.sv
```diff
@@ -47,3 +47,16 @@ Message ::= config::Decorated CmdArgs l::Location m::String
     then err(l, m)
     else wrn(l, m);
 }
+
```
ericvanwyk, line 50: Do we have `silver/core` library code for origins and messages? Maybe functions like this could live there and lead to more use of origins and be easier to document for other projects to use. Could we do something like that?

grammars/silver/compiler/extension/constructparser/Construct.sv
```diff
@@ -16,64 +16,50 @@ top::Root ::= gdcl::GrammarDcl  mStmts::ModuleStmts  is::ImportStmts
...
           productionRHSCons(
-            productionRHSElem(name("ioIn",  top.location), '::',
-              nominalTypeExpr(qNameTypeId(terminal(IdUpper_t, "IOToken"), location=top.location), location=top.location),
-              location=top.location),
-            productionRHSNil(location=top.location),
-            location=top.location
-          ),
-          location=top.location
-        ), location=top.location),
+            productionRHSElem(name("ioIn"), '::',
```
ericvanwyk, line 48: We should remember this specific diff and the collection of them for writing something about how much origin tracking saves us. How many characters/lines shorter is the code after this change?

grammars/silver/compiler/extension/deriving/Derive.sv
```diff
@@ -90,108 +88,95 @@ top::AGDcl ::= nt::Decorated! QName
...
-                          location=top.location),
-                        location=top.location),
-                      Silver_Expr {true},
-                      location=top.location),
-                    location=top.location),
-                  includedProds),
-                top.location),
-              Silver_Expr {silver:core:error("Unexpected production in derived Eq instance!")},
-              location=top.location),
+                              range(0, length(prod.namedSignature.inputElements)))))),
```
ericvanwyk, line 176: Wow - this one is good too.

grammars/silver/compiler/extension/doc/core/DocConfig.sv
```diff
@@ -4,67 +4,34 @@ grammar silver:compiler:extension:doc:core;
...
+  | fileNoDocsConfig v::Boolean
+  | tocConfig v::Boolean
+  ;
 
 @{-
  - Is this @link[DocConfigSetting] local to the file (e.g. @@title) or to the
  - grammar (e.g. @@grammarTitle)?
  -}
 synthesized attribute fileScope::Boolean occurs on DocConfigSetting;
-
```
ericvanwyk, line 14: I guess we see data nonterminals here too - so any calculation of shortening lines/chars is due to more than just origin tracking....

ericvanwyk (approved):

grammars/silver/compiler/analysis/warnings/flow/MwdaFlag.sv
krame505: Yeah, we have `errFromOrigin`, `wrnFromOrigin`, etc. in `silver:langutil`.  This is a specific helper for MWDA warnings that should be reported as errors with the `--mwda` flag.

grammars/silver/compiler/extension/constructparser/Construct.sv
krame505: We saved 437 lines across the whole Silver compiler - the line count didn't change as much as you might expect, since most of the locations are on the same line as something else.  The char count is a bit more significant, saving 52876 chars.

grammars/silver/compiler/extension/doc/core/DocConfig.sv
krame505: Yes I suppose so - so the figures in my previous comment might not be accurate.

ericvanwyk: "might not" -> "are" :)   Seems like a big majority were origin tracking though.

Issue #812, opened 2023-10-09
Dispatch productions and tree sharing


krame505:
I am considering a pretty major change to tree sharing, related to dispatching productions.  

## Background
One might currently define an overloaded operator as
```
production negOp
top::Expr ::= e::Expr
{
  e.env = top.env;
  forwards to e.typerep.negProd(e);
}
production negInt
top::Expr ::= e::Decorated! Expr with {env}
{
  undecorates to negOp(new(e));
  top.errors = e.errors;
  top.type = e.type;
  ...
}
production negBool
top::Expr ::= e::Decorated! Expr with {env}
{
  undecorates to negOp(new(e));
  top.errors = e.errors;
  top.type = e.type;
  ...
}

synthesized attribute negProd::(Expr ::= Decorated! Expr with {env});
nonterminal Type with negProd;
production intType
top::Type ::=
{
  top.negProd = negInt;
}
production boolType
top::Type ::=
{
  top.negProd = negBool;
}
```

This makes use of tree sharing through unique references, such that the operand `e` gets decorated with `env` prior to forwarding, and any other inherited attribute not needed to determine the forward tree are supplied after forwarding.

## Problem
The problem with this approach is evident when an extension wants to introduce an analysis involving a new inherited attribute on `Expr`, and wants to define the inherited attribute on the forwarding production:
```
production implies
top::Expr ::= a::Expr  b::Expr
{
  top.inline = implies(a.inline, b.inline);
  forwards to orOp(negOp(@a), @b);
}

inherited attribute subs::[(String, Expr)] occurs on Expr;
functor attribute inline occurs on Expr;
propagate subs, inline on Expr;

aspect production negOp
top::Expr ::= e::Expr
{
  e.subs = top.subs;
}
...
```
Here `implies` decorates `a` and `b` through forwarding with tree sharing, and the extension writer wants the `inline` attribute to preserve implications.  More realistically, `implies` and `inline` could be defined in different extensions, one of which is built with knowledge of the other.  

To access `inline` on `a`, `a` must first be decorated with `subs`.  This requires `negOp` to provide an equation for `subs`, which depends on `top.subs`[^1].  However, this attribute isn't in the reference set of the reference taken in `negOp`, potentially leading to hidden, unexpected transitive dependencies when accessing `subs` on a reference.  This could lead to a missing equation error in the following (a bit contrived) example of another extension providing an overload for `negOp`:
```
production negThing
top::Expr ::= e::Decorated! Expr with {env}
{
[...]

PR #817, opened 2023-10-29
Change over remaining built-in operators to use algebra type classes

### LINKED ISSUE #802: Using type class methods for arithmetic operators
See #451.  This project was started a while back, to change the various host-language operators into forwarding productions.  This saves a significant amount of boilerplate in the Silver compiler, as specifying various analyses for these operators (type checking, flow analysis, implicit monad translation, java translation) can be avoided.  

This change was done for the `++` and various equality operators, but wasn't quite finished for the remaining Boolean and arithmetic operators.  

Supporting overloading for these operators is useful for a few reasons.  One possibility is that one can write instances for abstract syntax operator productions in their language, providing a lightweight version of concrete object language syntax - one could write something like `var(a) + var(b) * intConst(2)`.  In simple languages, or extensions that introduce a new sort of expression, it may not be feasible to set up an extended version of Silver with quotation for the new expression nonterminal.  
Note that this previously wasn't feasible with the need to supply a location annotation when applying these productions, but this is no longer an issue when using origin tracking.  

Another motivation is that we currently only support `Integer` and `Float` as numeric types.  There may be reasons to desire other representations of numbers, e.g. building an interpreter for a language that supports other sorts of numbers.  At the moment changing the type checking logic for these operators would be quite annoying, but would be trivial with type classes. I can recall wanting an unlimited-width integer type at one point (but I don't quite remember why.)  

One of @ericvanwyk's complaints is that `add`, `sub`, `neg`, etc. are taking names that should be available as the names of operator productions.  I personally find `addOp`, `subOp`, etc. to be more descriptive names anyway.  Or `plus`, `minus`, etc. are also options.  

One can still define productions with the same name as functions in `silver:core` - however in other grammars importing the grammar defining these productions, one must `import silver:core hiding add, sub, mul, ...;`.  We may consider improvements to Silver's module system, or name resolution semantics, to be able to state that one imported module should take precedence over another for name conflicts (should explicit imports always shadow `silver:core`?)

krame505:
Fixes #802.

# Changes
Finish changing over the built-in operators to use the algebra type class methods.  This makes those operators overloadable, but more significantly, allows the removal of a massive amount of code in type checking and implicit monads.

There may be a slight performance regression here, due to primitive operations like && now involving a function call in the translation.  We should still probably special-case the translation of these operators for common primitive types, but that is currently impractical due to #812 - see the comments I added on the translation.

# Documentation
This is mostly deleting code and making a few productions forward - the algebra type classes themselves are already fairly well documented.

# Testing
No new features added.

Checked that errors for operators are still reasonable when operands have the wrong types.


grammars/silver/compiler/extension/doc/driver/BuildProcess.sv
```diff
@@ -95,7 +95,7 @@ top::DriverAction ::= a::Decorated CmdArgs  specs::[Decorated RootSpec]
       | grammarRootSpec(g, _, _, _, _, _) ->
           if (length(g.documentedNamed)+length(g.undocumentedNamed))!=0
           then [s" - [${g.grammarName}]: ${toString(length(g.documentedNamed))}" ++
-                s"/${toString(toInteger(length(g.undocumentedNamed)+length(g.documentedNamed)))} " ++
```
remexre, line 98: toInteger can be `Integer ::= Integer`? :eyes:

That seems warning-worthy...

krame505: Yeah, it's a method of type class `ConvertablePrim`.  Not sure how to make it a warning nicely.

grammars/silver/compiler/translation/java/core/Expr.sv
```diff
@@ -540,7 +540,9 @@ top::Expr ::= 'false'
   top.translation = "false";
   top.lazyTranslation = top.translation;
 }
-
+{- TODO: We should re-enable the specialized translations here for primitive types,
```
remexre, line 543: We're intentionally leaving both blocks in this file intentionally commented out until this is resolved, right?

grammars/silver/core/BooleanAlgebra.sv
```diff
@@ -10,3 +10,5 @@ grammar silver:core;
   - * Double Negation Elimination: `not(not(x)) = x`
   -}
 class HeytingAlgebra a => BooleanAlgebra a {}
+
+instance BooleanAlgebra Boolean {}
```
remexre, line 14: oops lmao

grammars/silver/compiler/translation/java/core/Expr.sv
krame505: Yeah

grammars/silver/core/DivisionRing.sv
```diff
@@ -14,6 +14,14 @@ class Ring a => DivisionRing a {
   recip :: (a ::= a);
 }
 
+instance DivisionRing Integer {
+  recip = \ a -> 1/a;
+}
```
remexre, line 19: This isn't lawful, is it?

krame505: Oh, good catch.  Technically Float doesn't either, but close enough I guess?

grammars/silver/core/EuclideanRing.sv
```diff
@@ -19,9 +19,47 @@ class CommutativeRing a => EuclideanRing a {
...
+function modInteger
+Integer ::= a::Integer b::Integer
+{
+  return error("Foreign function");
+} foreign {
+  "java": return "(%a% % (int)%b%)";
+}
+
+instance EuclideanRing Float {
+  degree = \n -> if n < 0.0 then toInteger(-n) else toInteger(n);
```
remexre, line 44: This should be `\_ -> 1`, I think, since there's a Field instance?

grammars/silver/core/EuclideanRing.sv
```diff
@@ -19,9 +19,47 @@ class CommutativeRing a => EuclideanRing a {
...
+  div = divFloat;
+  mod = modFloat;
```
remexre, line 46: okay so bad news... the Field laws say this should be `\_ -> 0`... Which like, yeah, that's the solution to make `a = (a / b) * b + (a % b)` hold, but... maybe not what want? Well, what even are the semantics of % on floats? Do we like those semantics for some reason?

krame505: I'm not familiar with what degree means in a Field.  I don't particularly care about the semantics for `%` on floats, but it's easiest to match the Java semantics if feasible?

grammars/silver/core/Field.sv
```diff
@@ -12,3 +12,6 @@ grammar silver:core;
   - * Non-Zero Multiplicative Inverse: `mod(a, b) = zero`
   -}
 class DivisionRing a, EuclideanRing a => Field a {}
+
+instance Field Integer {}
```
krame505, line 16: Should Integers be a Field? I thought they were but they don't satisfy the law for DivisionRing?

remexre, line 16: ehhhhhhhhhhhh this is a suspect instance IMO. ℤ is not a field, since e.g. 2 has no inverse. `ℤ mod n` can be a field because of overflow (e.g. `(2 ≡ 3⁻¹) mod 5` since `(2 * 3 = 1) mod 5`), but I don't _think_ that's true of signed 32-bit integers (because 2^32 isn't prime; though I'm not sure if `ℤ mod n` groups for non-prime n are necessarily not fields?)

Also, this file's missing a trailing newline.

krame505: Yeah, makes sense

remexre (approved):

PR #818, opened 2023-10-30
Add genericShow function in silver:core

krame505:
# Changes
This adds `genericShow :: (String ::= a)`, as a means to turn arbitrary values into strings.  This is an improved replacement for `hackUnparse`, making use of `genericPP` from `silver:langutil:reflect`. However `genericShow` can be used from the runtime and generated code without fear of adding a dependency on the langutil library - the implementation of `genericShow` calls `silver:langutil:reflect:genericPP` and `silver:langutil:pp:show` via Java reflection, falling back to `hackUnparse` if the langutil library is not availalble.  

This function is primarally intended for debugging, diagnostics and error reporting in Silver extension-generated code.  Pretty-printing values of known types should be done using the `Show` type class.

The Silver compiler has been updated to use `genericShow` in place of `hackUnparse`.

# Documentation
Added a doc comment and updated the doc comment for `hackUnparse`.  Also, deleted a number of now-outdated TODO comments.

# Testing
Updated the silver_features and silver_construction test suites to use `genericShow` in place of `hackUnparse`.  Also, change the testing framework itself to use `genericShow` in reporting failures.


grammars/silver/compiler/extension/testing/EqualityTest.sv
```diff
@@ -151,8 +151,8 @@ ag::AGDcl ::= kwd::'equalityTest'
         attributeDef(concreteDefLHS(qNameId(tref)), '.', qNameAttrOccur(qNameId(msgref)), '=',
           foldStringExprs([
             strCnst("Test at " ++ getParsedOriginLocationOrFallback(ag).unparse ++ " failed.\nChecking that expression\n   " ++
-              stringifyString(value.unparse) ++ "\nshould be same as expression\n   " ++
-              stringifyString(expected.unparse) ++ "\nActual value:\n   "),
+              value.unparse ++ "\nshould be same as expression\n   " ++
+              expected.unparse ++ "\nActual value:\n   "),
```
remexre, line 155: other stringifyStrings became escapeStrings -- should these too?

krame505: No, these weren't actually needed and were getting escaped twice.

remexre (approved):

PR #831, opened 2024-03-12
Initial support for signature sharing and dispatch productions

krame505:
# Changes
See #812.  

This adds initial syntax, type checking and translation for marking signature items as shared, and defining dispatch signatures.  The corresponding MWDA changes are still pending, but it will be more straightforward to make those changes concurrently with removing support for unique references.  However unique references are still used in the Silver compiler, so we need to change over to this new dispatch sharing mechanism before they can be removed.  Bootstrapping that change will be much less tedious by merging the incomplete feature, now.

This also makes giving a name for the production LHS optional, which is more consistent with signature labels being optional on the RHS.  Currently we always syntactically require a name like `top` for the LHS, even when it is never referenced in the production body.  I am making this change now because dispatch signature declarations reuse the `ProductionSignature` nonterminal, and it is not particularly helpful to require a name for the LHS in a signature declaration - one should be able to write
```
dispatch UnaryOp = Expr ::= @e::Expr;
```
and not
```
dispatch UnaryOp = top::Expr ::= @e::Expr;
```

# Documentation
https://github.com/melt-umn/melt-website/pull/60

# Testing
Updated tests are added in https://github.com/melt-umn/silver/pull/837


ericvanwyk (approved):
Looks good! Thanks.

PR #847, opened 2024-05-25
Fix bugs with occurs-on constraints

### LINKED ISSUE #846: Crashing function with type constraints
If I have
```
our_own_output ++ showDoc(width, state.pp) ++ "\n"
```
it works fine.  If I have the following (that should be equivalent, since none of the inherited attributes are being accessed)
```
our_own_output ++
decorateAndShow(state, top.proverState.knownTypes,
   top.proverState.knownRels,
   top.proverState.knownConstrs, width, false) ++ "\n"
```
where we have
```
function decorateAndShow
attribute typeEnv occurs on a,
attribute relationEnv occurs on a,
attribute constructorEnv occurs on a,
attribute fromAbella<a> {typeEnv, relationEnv, constructorEnv} occurs on a,
attribute pp {} occurs on a =>
String ::= a::a ty::Env<TypeEnvItem> rel::Env<RelationEnvItem>
           cons::Env<ConstructorEnvItem> width::Integer
           doFromAbella::Boolean
{
  a.typeEnv = ty;
  a.relationEnv = rel;
  a.constructorEnv = cons;
  return if doFromAbella
         then showDoc(width, a.fromAbella.pp)
         else showDoc(width, a.pp);
}
```
it crashes with the eventual message
```
(extensibella.main.run.PdecorateAndShow in PdecorateAndShow.java:222): Error while evaluating function extensibella:main:run:decorateAndShow
(silver.langutil.pp.PshowDoc in PshowDoc.java:147): Error while evaluating function silver:langutil:pp:showDoc
(DN.237): Index 3 out of bounds for length 3
```
I looked but was unable to find a message of the form shown in the last line here in Silver or Extensibella, but this is apparently the message Java gives for array out-of-bounds.

If I instead use the function made specific to one type
```
function decorateAndShowP
String ::= a::ProofState ty::Env<TypeEnvItem> rel::Env<RelationEnvItem>
           cons::Env<ConstructorEnvItem> width::Integer
           doFromAbella::Boolean
{
  a.typeEnv = ty;
  a.relationEnv = rel;
  a.constructorEnv = cons;
  return if doFromAbella
         then showDoc(width, a.fromAbella.pp)
         else showDoc(width, a.pp);
}
```
it does not crash.

I am using the original function with two different types in the original code.

krame505:
# Changes
Fixes #846.

Inherited occurs-on constraints should only be satisfied for *undecorated* types, since the constraint means that you can supply an equation for the attribute.  If there are only synthesized constraints, passing a reference is fine, because we will decorate it with no additional inherited attributes.  This adds an error message when attempting to resolve an inherited occurs-on constraint for a reference.

Also fix a bug introduced in #837, where supplying additional inherited attributes to a shared tree in a polymorphic context could crash if the new array of attributes is smaller.

Also make some tweaks to the fetch-jars script to allow grabbing the latest local jars for a branch - I'm on the road right now and can't feasibly download jars.

# Documentation
This is a bug fix, removing an inadvertent undocumented misfeature of satisfying inherited occurs-on constraints for references.  

# Testing
Added a test for the new error message.  Replicating the second problem is non-trivial and nondeterministic, depending on the indices that attributes are dynamically assigned.

### DIFF
--- a/runtime/java/src/common/DecoratedNode.java
+++ b/runtime/java/src/common/DecoratedNode.java
@@ -232,8 +232,10 @@ public class DecoratedNode implements Decorable, Typed {
 	}
 
 	private void copyInhOverrides(final DecoratedNode parent, final Lazy[] inhs, final Lazy[] newInhs) {
-		assert inhs.length == newInhs.length;
-		for(int i = 0; i < inhs.length; i++) {
+		// Arrays can differ in length if a tree is shared in a polymorphic decoration site.
+		// The new inh array should never be larger than the original one.
+		assert inhs.length >= newInhs.length;
+		for(int i = 0; i < newInhs.length; i++) {
 			if(newInhs[i] != null) {
 				Lazy newInh = newInhs[i].withContext(parent);
 				if(inhs[i] == null) {
--- a/fetch-jars
+++ b/fetch-jars
@@ -20,8 +20,15 @@ done
 
 COMMIT_ARTIFACTS="https://foundry.remexre.xyz/commit-artifacts/"
 
+[[ $* == *--local* ]]
+LOCAL_JARS=$?
+
 function has_jars {
-  wget --spider -q "$COMMIT_ARTIFACTS/$1"
+  if [[ $LOCAL_JARS -eq 0 ]]; then
+    [[ -d "JARS-BAK/$1" ]]
+  else
+    wget --spider -q "$COMMIT_ARTIFACTS/$1"
+  fi
 }
 
 rev=""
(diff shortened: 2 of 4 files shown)


runtime/java/src/common/DecoratedNode.java
```diff
@@ -232,8 +232,10 @@ public DecoratedNode decorate(final DecoratedNode parent, final Lazy[] inhs, fin
 	}
 
 	private void copyInhOverrides(final DecoratedNode parent, final Lazy[] inhs, final Lazy[] newInhs) {
-		assert inhs.length == newInhs.length;
-		for(int i = 0; i < inhs.length; i++) {
+		// Arrays can differ in length if a tree is shared in a polymorphic decoration site.
+		// The new inh array should never be larger than the original one.
```
RandomActsOfGrammar, line 236: You should probably comment here *why* the new one should never be larger than the original, unless this is absolutely clear for someone using this.

RandomActsOfGrammar (approved):

Issue #843, opened 2024-05-17
Unary operator for new


krame505:
Previously my thinking has been that tree sharing should be the default over undecoration in most cases, and that most cases where undecoration is needed can be avoided by data nonterminals/nondecorated locals.  Thus, `new()` is a bit more verbose than `@` in part to encourage one to use tree sharing instead.  Additionally, to the flow analysis `new()` is just a function, while `@` has special semantics.

However in practice when doing the #751 refactoring in Silver, there are (so far) 971 uses of `new`, to 657 uses of `@`.  Some of these probably do point to places where further refactoring is possible, but there many cases (e.g. dealing with `Type`, or in a strategy attribute rule) where undecoration is the right choice.  So it might be good to give undecoration its own unary operator for the sake of ergonomics.  Having some symmetry with `@` for the two ways of turning a decorated tree into an undecorated one is also nice.

I'm undecided what the best syntax would be.  The unused operator symbols in silver are ``~ ` # ^ & | \`` - maybe `~` or `#`?  Another option is `*` by analogy with dereferencing in C, etc.  This is also the infix multiplication operator, but that isn't an issue in C (and probably less so in Silver?)

RandomActsOfGrammar:
I don't think that adding a new symbol is a good idea.  Typing `new(t)` isn't that hard, and the name "new" gives some idea of what it is.  Using `~t`, `#t`, `*t`, or any other unary symbol means only the Silver-initiated can understand it.  A related name is also easier for people to remember what it is after reading a definition once.

A symbol could be useful if there was a clear analogy to an existing, common programming construct, which I do not see.  I see your analogy to pointer dereferencing in C, as both sort of unwrap the underlying information, but it is not a close analogy, and trying to use it to explain the syntax connection is going to be more confusing to new users than helpful.

krame505:
> I don't think that adding a new symbol is a good idea. Typing new(t) isn't that hard

That was my initial thinking, too, but I'm second-guessing this after seeing the amount of clutter added when removing implicit undecoration.  And going forward, undecoration will be a very fundamental operation in Silver, like attribute access.

> the name "new" gives some idea of what it is. Using ~t, #t, *t, or any other unary symbol means only the Silver-initiated can understand it. A related name is also easier for people to remember what it is after reading a definition once.

[...]

Issue #842, opened 2024-05-16
Permit accessing the forward parent in implementation productions


krame505:
In the process of refactoring Silver as a part of #751, I have noticed a recurring pattern:
```
aspect production functionInvocation
top::Expr ::= @e::Expr @es::AppExprs @anns::AnnoAppExprs
{
  forward t = application(new(e), '(', new(es), ',', new(anns), ')');

  top.merrors := t.merrors;
  top.mUpSubst = t.mUpSubst;
  top.mtyperep = t.mtyperep;
  top.monadRewritten = t.monadRewritten;
  top.monadicNames = t.monadicNames;
}
```
Here an extension wants to compute some synthesized attributes on the dispatch production (`application` here) and avoid writing the same equations in implementation productions.  As previously discussed offline, it isn't feasible to allow synthesized equations on an implementation production to be omitted and implicitly defaulted to copy from the forward parent.  However perhaps one should be able to write those copy equations explicitly.  These equations should typically never be accessed, however one may also wish to make use of a synthesized equation in the dispatching production in a different attribute equation in an implementation production.  The above approach requires re-decoration of children and re-computation of attributes; sharing isn't permitted in an extension context like this, as the `application` production does not itself implement the `Application` signature.  Some other production may also have forwarded to `functionInvocation`, so this may not always give the desired result.

The fix is just to make the forward parent accessible, like how one can explicitly access e.g. `forward.typerep`.  I'm not sure what the syntax for this should be - `forwardParent`, `backward`?

PR #849, opened 2024-06-19
Require explicit new for undecoration

### LINKED ISSUE #751: Simplifying Silver's possibly-decorated type inference semantics

krame505:
# Changes
Fixes #751. 

This changes the semantics of referring to a (non-data) nonterminal child/local/translation attribute/pattern variable to be always decorated, and require explicit `new` to undecorate the tree.  Let expressions binding decorated trees also no longer implicitly undecorate.

The bulk of the changes here are a largely automated refactor to use tree sharing, and explicit `new` where that isn't possible.  Thus if anyone would actually like to try reviewing this, it is probably easiest to look at individual commits that have the type system changes.

This is a major breaking change - downstream projects will need to be refactored as well before merging this PR.  Issues #842 and #843 should probably also be addressed before attempting this refactoring.

There is a slight regression in the `deep-rebuild` time (maybe 10%?) , which I haven't investigated too closely yet.  I was expecting performance to improve through more use of sharing and type system simpliciations, but there is also a new pass for specializing `InhSet`s, and a lot of calls to `new` added.

# Documentation
The documentation has not yet been updated.  I think the main things to fix are https://melt.cs.umn.edu/silver/concepts/decorated-vs-undecorated/#implicit-decoration and any uses of implicit undecoration in examples.

# Testing
The Silver compiler, test suite and tutorials have been updated to account for these changes.


ericvanwyk (approved):
Approved at MELT Oct 15 meeting.

PR #850, opened 2024-06-22
Support referencing forwardParent

### LINKED ISSUE #842: Permit accessing the forward parent in implementation productions

krame505:
# Changes
Fix #842.

# Documentation
https://github.com/melt-umn/melt-website/pull/63

# Testing
Added a useage to the testsuite

### DIFF
--- a/grammars/silver/compiler/definition/core/Expr.sv
+++ b/grammars/silver/compiler/definition/core/Expr.sv
@@ -266,6 +266,18 @@ top::Expr ::= q::'forward'
   forwards to baseExpr(qName("forward"));
 }
 
+concrete production forwardParentReference
+top::Expr ::= 'forwardParent'
+{
+  top.unparse = "forwardParent";
+
+  top.typerep = top.frame.signature.outputElement.typerep.asNtOrDecType;
+  top.errors <-
+    if !any(map((.elementShared), top.frame.signature.inputElements))
+    then [errFromOrigin(top, "This production has no shared children and is not known to be the target of forwarding.")]
+    else [];
+}
+
 concrete production application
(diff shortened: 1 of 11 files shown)


ericvanwyk (approved):
Accepted in MELT Oct 15 meeting.

PR #851, opened 2024-06-27
Add unary undecoration operator '^'

### LINKED ISSUE #843: Unary operator for new

krame505:
# Changes
Fix #843. 

# Documentation
https://github.com/melt-umn/melt-website/pull/64

# Testing
Added test cases using the operator.

### DIFF
--- a/grammars/silver/compiler/definition/core/Expr.sv
+++ b/grammars/silver/compiler/definition/core/Expr.sv
@@ -667,15 +667,6 @@ top::Expr ::= 'decorate' e::Expr 'with' '{' inh::ExprInhs '}'
   inh.allSuppliedInhs = inh.suppliedInhs;
 }
 
-concrete production decorationSiteExpr
-top::Expr ::= '@' e::Expr
-{
-  top.unparse = s"@${e.unparse}";
-
-  top.typerep = e.typerep.decoratedType;
-  e.isRoot = false;
-}
-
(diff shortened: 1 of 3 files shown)


ericvanwyk (approved):
Reviewed in MELT meeting on Oct 15

PR #852, opened 2024-07-29
Permit building against jars containing compiled grammars

### LINKED ISSUE #152: Removing obsolete grammars, and naming new libraries

### LINKED ISSUE #232: Ability for Silver compiler to find compiled grammars inside jars
The time for this feature may have come.

The added feature would be the ability to:

1. `-I /path/to/a.jar` and search inside it for `Silver.svi` files to find grammars.
2. Maybe `-I /path/` and search inside `*.jar` at that path for `Silver.svi` files.

The benefits of this are starting to compound:

1. We could ship our "standard library" already compiled with Silver. Then `--clean` builds of downstream stuff wouldn't have to rebuild these grammars. And built jars all wouldn't have to include this stuff again.
2. When attempting to re-use artifacts (such as when building AbleC extensions) instead of needing to checkout a copy of AbleC (and instead of needing to copy the `generated` folder from ablec to the extension, to re-use generated files in Jenkins) we would just need to copy a single `jar` file. In particular, this would take a HUGE IO load off coldpress, which could dramatically speed up compile times. (Copying 30MB in one file is a lot less load than copying 30,000 small files. I think coldpress is bogging down during jenkins builds due to a lack of IOPS.)
3. `deep-rebuild` automagically works with changes to translation and runtime at the same time, by timing when these things get built. However, if there are also necessary changes to `core` things get complicated again. With this change, we could have it find `core` via pre-compiled jar at first, and then switch over to using the new core at the appropriate time. This would further simplify builds involving invasive changes to translation.

The process for achieving this would look like:

1. Add necessary IO functions capable of looking inside jar/zip files and searching for files.
2. Change how grammars are searched for in build process
3. Fix up how classpath dependencies get emitted in built jars as a result of depending on other jars. e.g. default mode, relative jar mode, one jar mode.

krame505:
# Changes
Fixes #232, fixes #152. Depends on #849, depends on #842, depends on #851.  

This adds support for including a pre-compiled jar (`-I foo.jar`) instead of needing to include all the source files or resort to hacks.  I finally took a stab at this because dealing with lots of dependent ableC extensions was becoming a pain, and I am going to be revamping ableC and friends next.  Allowing to build against a jar means that we can have multiple parallel silver builds, as long as each has its own `generated/` dir, yet avoid rebuilding the same grammars more than once. This also should permit simplifying our Jenkins flow a bit. And something like @remexre's nix build rules would be a lot more straightforward.

As a test of this I also refactored the Silver compiler and standard libraries into separate artifacts as outlined in #152.  This avoids some of the hacks that were previously needed for building the runtime and accommodating translation changes in `deep-rebuild`.  

Long-term we should probably adopt a proper build system like gradle, or make some Silver-based build tool with a DSL for specifying artifacts.  However for the moment, I just revamped the build scripts to do this.  `deep-rebuild` is now a bit slower, but you shouldn't need to run it as frequently.  Most of the time you can just run `./make-all` after making library changes, or `./make-compiler` after just changing the compiler.  `./make-compiler && ./check-compile --clean` should work as a smoke test when making translation changes.

# Documentation
I added source comments and commented the new build scripts.  I'm not really sure if there is a spot on the website for this sort of discussion.  

# Testing
This feature is now used in building the Silver compiler.  The existing test suite passes, and I also ensured that the language server still works after making a few changes in how that gets built too.


deep-rebuild
```diff
@@ -1,90 +1,71 @@
...
-echo " > Silver Build"
-echo "   Runtime Build"
-echo "   Silver Build"
-echo ""
-echo "Start ..."
-# Also building silver:xml:ast here since the runtime depends on that
-time java $JVM_ARGS --relative-jar --clean silver:compiler:composed:Default silver:xml:ast
 
-# Modifications may have been made to the runtime
+# Doesn't need to be clean, since we are starting with a known good state of generated/.
```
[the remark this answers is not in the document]
krame505: The idea is that in normal development, you are starting with a working version of the Silver compiler, and the current cache of generated files were generated by it.  The scripts endeavor to preserve this invariant (e.g. `fetch-jars` now deletes the generated files cache) and if you make "ABI-breaking" changes you should run `deep-rebuild`.  

In the case when you aren't starting from a good state, you can always run `./deep-rebuild --clean`.

grammars/silver/langutil/reflect/concretesyntax/ConcreteSyntax.sv
```diff
@@ -1,7 +1,10 @@
-grammar silver:reflect:concretesyntax;
+grammar silver:langutil:reflect:concretesyntax;
+
+-- This grammar logically belongs under of silver:reflect, but it needs to live
+-- here due to its dependency on the langutil artifact.
 
-imports silver:langutil;
 imports silver:reflect;
+imports silver:langutil;
```
remexre, line 7: Does the order here have semantic meaning, or did it just change miscellaneously?

grammars/silver/langutil/reflect/Util.sv
```diff
@@ -1,5 +1,7 @@
 grammar silver:langutil:reflect;
 
+import silver:langutil:reflect:serialize;  -- Include this in the silver.langutil artifact.
```
remexre, line 3: i'd kinda rather we had a greppable keyword for "this import only exists for the purpose of having larger artifact JARs" that forwards to `only {}` or something

grammars/silver/langutil/reflect/concretesyntax/ConcreteSyntax.sv
krame505: Yeah it's not important, I think the import got deleted and then re-added.

grammars/silver/langutil/reflect/Util.sv
remexre: (greppable so when/if we get a realer build system, we can find the import edges that should be expressed there instead)

runtime/java/build.xml
```diff
@@ -26,7 +26,7 @@
   </target>
 
   <target name='build' depends='init'>
-    <javac debug='on' classpathref='compile.classpath' srcdir='${src}:${grammars}' destdir='${bin}' includeantruntime='false' source='1.8' target='1.8' release='8'>
+    <javac debug='on' classpathref='compile.classpath' srcdir='${src}:${grammars}' destdir='${bin}' includeantruntime='false' release='11'>
```
remexre, line 29: Are we intentionally bumping our minimum JVM version?

krame505: Yeah, I don't remember exactly but I think I needed some file API for examining jars that was added in Java 11. It was released in 2018 and Java 8 is EOL in a couple of years so I don't think this is going to be an issue for anyone.

remexre: okay, we probably need to update at least the readme then; any other docs that say jdk/jvm/java 8 too

krame505: Updated the README; I didn't see any other references to Java 8 in the docs.

support/bin/silver-custom
```diff
@@ -32,4 +32,16 @@ fi
 # Set flags if not overriden in environment
 SVJVM_FLAGS=${SVJVM_FLAGS:-"-Xmx5000M -Xss20M"}
 
-java $SVJVM_FLAGS -jar "$SILVER" "$@" && ant
+# Save the command line arguments to pass to Silver
+SV_ARGS="$@"
+
+# Tell ant if we named the build file something else
+while [ "$#" -gt 0 ]; do
+    if [ "$1" = --build-xml-location ]; then
```
remexre, line 40: should this be a `case` with a `--build-xml-location=*` clause to catch that form too, or will that blow up for other reasons

krame505: This is parsing a flag that is also passed when invoking the Silver jar, and our `cmdargs` library doesn't currently support that format.

support/bin/silver
```diff
@@ -31,5 +31,17 @@ fi
 # Set flags if not overriden in environment
 SVJVM_FLAGS=${SVJVM_FLAGS:-"-Xmx5000M -Xss20M"}
 
+# Save the command line arguments to pass to Silver
+SV_ARGS="$@"
```
remexre, line 35: shellcheck complains,

```
In support/bin/silver line 35:
SV_ARGS="$@"
        ^--^ SC2124 (warning): Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.

For more information:
  https://www.shellcheck.net/wiki/SC2124 -- Assigning an array to a string! A...
```

make-stdlib
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+# Build the Silver standard libraries.
+# Note that this does not rebuild the silver:core and the runtime.
+
+set -e
+
+export SILVER_HOME=$(pwd)
+JVM_ARGS="-Xss20M -Xmx1G -jar ../jars/silver.compiler.composed.Default.jar --no-stdlib --relative-jar $@"
```
remexre, line 9: shellcheck complains,

```
In make-stdlib line 9:
JVM_ARGS="-Xss20M -Xmx1G -jar ../jars/silver.compiler.composed.Default.jar --no-stdlib --relative-jar $@"
         ^-- SC2124 (warning): Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.
```

it has a bunch of other warnings too, but i _think_ they're all "false positives for our purposes (e.g. idt silver can make a jar file that starts with a `-`)" -- they're all easy to fix, though

make-all
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+# This does a single-pass build of the Silver standard libraries, runtime and compiler.
+# This is normally sufficient if there are no changes to the translation.
+
+set -eu
+
+NOINSTALL=1 ./make-core-runtime $@
+NOINSTALL=1 ./make-stdlib $@
+./make-compiler $@
```
remexre, line 10: these all definitely need `"$@"`

make-stdlib
remexre: lots of scripts are doing this; i think two reasonable things to do might be

- pass `"$@"` (note quotes) after `$JVM_ARGS` (note no quotes) to `java`
- make `JVM_ARGS` an array and do what shellcheck says

deep-rebuild
```diff
@@ -1,90 +1,71 @@
...
+# Only pass along flags like --clean, --mwda, etc. in the first pass.
+NOINSTALL=1 ./make-stdlib $@
+./make-compiler $@
```
remexre, line 38: "$@"

support/vs-code/silverlsp/src/extension.ts
```diff
@@ -25,15 +26,26 @@ export function activate(context: vscode.ExtensionContext) {
...
+			classPath.push(compilerJar);
 		}
-		let classPath = [launcherJar];
-		if (compilerJar) {
-			classPath.unshift(compilerJar);
+		classPath.push(launcherJar);
+
+		for (let jar of classPath) {
+			// Check if the jar file exists, and if not, raise an error.
+			if (!fs.existsSync(jar)) {
```
remexre, line 43: should activate be an async function, and should this be the version from `fs/promises`? https://code.visualstudio.com/api/references/vscode-api makes it sound like `Extension<T>.activate` should return a `Thenable<T>`

grammars/silver/langutil/Message.sv
```diff
@@ -153,6 +153,13 @@ String ::= m::Message
 fun messagesToString String ::= msgs::[Message] =
   implode("\n", map(showMessage, sortBy(messageLte, msgs)));
 
+@{--
+ - Returns a list of strings, without doing the check for an origin in extension-generated code.
+ - This is useful for pretty-printing error productions to avoid a circularity.
```
remexre, line 158: is the problem that this fixes the situation where an extension prod constructs an error prod to forward, or is it more specific? (if it's the former, shouldn't we be hitting this all over the place?)

krame505: This is a corner case where the message created by `showMessage` above can contain the pretty-print of the production that yielded the message (via `showOriginInfoChain`.)  If the error production uses `showMessage`, then you can end up with a weird circularity in pretty-printing it where that requires pretty-printing its own origin chain, which contains its pretty-print.

make-stdlib
```diff
@@ -0,0 +1,34 @@
...
+
+for lib in util langutil rewrite regex; do
+    echo    "=== BUILD LIBRARY $lib ==="
+    time java "${JVM_ARGS[@]}" silver:$lib
+    ant
+    export GRAMMAR_PATH="$GRAMMAR_PATH:silver.$lib.jar"
+done
+
+set +u
+if [ -z "$NOINSTALL" ]; then
```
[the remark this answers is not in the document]
krame505: Yeah, that's a lot nicer, thanks.

support/vs-code/silverlsp/src/extension.ts
krame505: No idea, I copied this from a tutorial a long time ago. Added a TODO comment for now.

grammars/silver/langutil/Attributes.sv
```diff
@@ -1,10 +1,15 @@
...
 grammar silver:langutil;
 
-exports silver:langutil:reflect; -- Contains pp definitions for AST
+exports silver:langutil:reflect with silver:core; -- Contains pp definitions for AST
 
 import silver:langutil:pp;
 
+-- Make sure these get built in the langutil artifact.
+import silver:langutil:lsp;
+import silver:langutil:unparse;
```
remexre:
```suggestion
import silver:langutil:lsp only;
import silver:langutil:unparse only;
```

grammars/silver/langutil/reflect/Util.sv
remexre:
```suggestion
import silver:langutil:reflect:serialize only;  -- Include this in the silver.langutil artifact.
```

grammars/silver/regex/AbstractSyntax.sv
```diff
@@ -5,6 +5,9 @@ imports silver:rewrite;
 imports silver:langutil;
 imports silver:langutil:pp;
 
+-- Not used here, but make sure it gets included in the artifact:
+import silver:regex:concrete_syntax as cnc;
```
remexre:
```suggestion
import silver:regex:concrete_syntax only;
```

remexre (approved):

PR #858, opened 2024-11-11
Fix overly-permissive exemption for cycles involving dispatch prods in inh completeness checks

krame505:
# Changes
Previously, I was making a blanket exemption for any cycle involving a dispatch signature, when resolving the decoration site tree for an inh attribute instance.  This is actually way too permissive, and allows lots of missing inherited equation errors unrelated to dispatching, that don't actually manifest as cycles.  

The only case that actually needs to be exempted is for an implementation prod that forwards to an application of the same dispatch signature.  This actually simplifies the resolution logic a bit, since we don't need separate handling of cycles with dispatch application and other kinds of sharing.

# Documentation
The MWDA support for tree sharing is still under development, docs will be written as a part of thesis writing.

# Testing
Tested on Silver and ableC extensions.


unironically (approved):

PR #859, opened 2024-11-11
Make diagnostic output readable when showing nested messages with generated origins

krame505:
# Changes
Suppress errors about nested messages having generated origins for a message that itself has a generated origin.

# Documentation
Added a comment.

# Testing
Tested with ableC-vector.


unironically (approved):

PR #860, opened 2024-11-11
Fix translation bug with dispatching

krame505:
# Changes
Every node tracks whether it contains shared children, that must be handled specially when the tree is undecorated.  When referencing a dispatch implementation production, we don't know if it will be supplied with shared children, but we need to assume so; unlike regular productions, a dispatch impl production can be indirectly constructed with shared children.

# Documentation
Added a comment; this is an implementation bug fix.

# Testing
Checks exist in the silver runtime that the result of undecorating a tree does not contain shared trees.  This was triggered in testing an ableC extension; this fix resolves the error.


unironically (approved):

PR #862, opened 2024-11-20
Adds uniq and uniqBy.

remexre:
These allow doing the same task as nub and nubBy in O(n log n), if you have an Ord instance. (Adding a warning when using nub on Ord types could be a fun lint :p)

Written to try to reduce allocation, which is why I didn't do more pattern-matching.

# Changes

# Documentation
Please briefly describe the documentation you have written in the following categories, or why you didn't write documentation for a category:
* what general comments you have added
  - none, it's not scarier than other uncommented things in that file
* what doc comments you have added
  - one per function
* what documentation you have added to the website for users of Silver
  - none, not a feature worth advertising
* what documentation you have added to the website for developers of Silver
  - none, it's not scarier than other things in that file

# Testing
Please briefly describe the tests you have written for the changes included in this pull request, or why it is not possible to test the changes.

Added three simple tests. Also about to use it in an ableC change.

*Please remove all the prefilled text other than the headings before submitting your pull request.*

### DIFF
--- a/grammars/silver/core/List.sv
+++ b/grammars/silver/core/List.sv
@@ -236,7 +236,7 @@ fun containsBy Boolean ::= eq::(Boolean ::= a a)  elem::a  lst::[a] =
 fun contains Eq a => Boolean ::= elem::a  lst::[a] = containsBy(eq, elem, lst);
 
 @{--
- - Removes all duplicates from a list.
+ - Removes all duplicates from a list. O(n^2).
  -
  - @param eq  The equality function to use
  - @param xs  The list to remove duplicates from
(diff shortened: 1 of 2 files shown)


grammars/silver/core/List.sv
```diff
@@ -247,13 +247,40 @@ fun nubBy [a] ::= eq::(Boolean ::= a a)  xs::[a] =
...
- - Removes all duplicates from a list.
+ - Removes all duplicates from a list. O(n^2).
  -
  - @param xs  The list to remove duplicates from
  - @return  A list containing no duplicates, according to ==.
  -}
 fun nub Eq a => [a] ::= xs::[a] = nubBy(eq, xs);
 
+@{--
+ - Removes all duplicates from a list. O(n^2).
```
krame505: I'm confused.  Is this supposed to be O(n lg n)?  This doesn't currently use `Ord`?

remexre: Whether to use nub or uniq(sort(...)) depends on whether you have an Ord instance or not -- if you don't, you have no choice but to keep using nub

krame505: Does the list need to be sorted for `uniq` to remove all duplicates?  Or is it just faster on a sorted list?  Either way the comments should indicate this.

remexre: Ah hold up I didn't see what line this was pointing at -- yeah, there's a copy paste bug here, this is O(n) and only removes consecutive duplicates.

I'll add a note about the most common use case of uniq(sort(...)) too

remexre: fixed

krame505 (approved):

PR #865, opened 2024-11-26
Add MWDA check for dispatch depending on inh attributes supplied by the target production

krame505:
# Changes
This adds a check for missing inherited equations on a child of a dispatching production, when those equations are needed to resolve the dispatch to determine what production is supplying the attributes.  For example:
```
production negOp
top::Expr ::= e::Expr
{
  local prod::UnaryProd = e.typerep.negProd;
  forwards to prod(@e);
}

dispatch UnaryProd = Expr ::= e::Expr;
production intNeg implements UnaryProd
top::Expr ::= e::Expr
{
  e.env = top.env;
  ...
}
```
Here `e` in `negOp` has the application of `prod` as its decoration site, and `UnaryOp` is known to supply `env` to its child.  However `prod` depends on `e.typerep`, which depends on `e.env`, which has not yet been supplied when `prod` is being computed - thus this "circularity" manifests as a missing equation.

This change adds a check that the dependencies of any applied dispatch productions do not include any inherited attributes that will be supplied via a child shared in dispatching.  

# Documentation
Added source comments.  Dispatching as a feature is not yet documented.

# Testing
Added test cases in the test suite.  Also tested on Silver and ableC, fixed a few issues discovered in Silver.

### DIFF
--- a/grammars/silver/compiler/analysis/warnings/flow/Inh.sv
+++ b/grammars/silver/compiler/analysis/warnings/flow/Inh.sv
@@ -890,10 +890,94 @@ top::Expr ::= 'decorate' e::Expr 'with' '{' inh::ExprInhs '}'
 aspect production decorationSiteExpr
 top::Expr ::= '@' e::Expr
 {
+  -- Check for an inherited attribute supplied via sharing, that may be needed to determine dispatching.
+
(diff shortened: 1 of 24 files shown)


unironically (approved):
Discussed in MELT meeting 11/26.

PR #878, opened 2025-01-10
No annotations can have the same short name on nonterminal

### LINKED ISSUE #815: Disallow multiple annotations with the same short name occuring on the same nonterminal
@remexre pointed out that if you have e.g. `annotation foo::Integer;` in grammar `a` and `annotation foo::String` in grammar `b`, you can have
```
nonterminal Thing with a:foo, b:foo;
production thing  top::Thing ::= {}
```
However, annotations get reduced to their short names in the function type, meaning you end up with multiple named parameters `foo`.  This actually still sort of works - you can apply it as `thing(foo=42, foo="hello")`.  However this is confusing, one would need to always qualify the annotation name when accessing it as `t.a:foo` or `t.b:foo`, and also the type expression syntax for named params seems to be broken with more than one named parameter.  

Annotations can only occur on a nonterminal in the grammar of the nonterminal anyway - thus I vote we ban having multiple annotations with the same short name occuring on a nonterminal.

allen-liaoo:
# Changes
Add checks so that two annotations with the same short name occurring on the same nonterminal will cause a compilation error. See #815 

# Documentation
Please briefly describe the documentation you have written in the following categories, or why you didn't write documentation for a category:
* what general comments you have added: N/A
* what doc comments you have added: N/A
* what documentation you have added to the website for users of Silver: Should we add this to the website?

# Testing
Added tests for both `occurs on` and `with` syntax.

### DIFF
--- a/grammars/silver/compiler/definition/core/OccursDcl.sv
+++ b/grammars/silver/compiler/definition/core/OccursDcl.sv
@@ -117,6 +117,13 @@ top::AGDcl ::= at::QName attl::BracketedOptTypeExprs nt::QName nttl::BracketedOp
     then [errFromOrigin(at, "Attribute '" ++ at.name ++ "' already occurs on '" ++ nt.name ++ "'.")]
     else [];
 
+  -- Make sure that no two annotations with the same short name (but different full names) can be declared on the same nonterminal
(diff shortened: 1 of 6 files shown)


krame505 (approved):
Looks good!  No need to update docs, I think.

PR #881, opened 2025-01-16
Silver now reports its latest git tag as its version

### LINKED ISSUE #672: Avoid hard-coding the version number in the Silver driver
We currently hard-code the version number in the Silver driver, which is a bit suboptimal as it quickly can get out of date, and it would be more useful to know what the latest commit was anyway.  

One fix would be to have self-compile and deep-rebuild call a script that updates a file in the driver containing only a global with a version string, computed from the output of `git describe --tags`.  This gives much more useful version info of the form `v0.4.4` if the most recent commit has a tag, or else something like `v0.4.4-1772-g99dcddc58`: the most recent tag, the number of commits since that tag, and an appended hash such that one can query `git show v0.4.4-1772-g99dcddc58` to view the latest commit.

allen-liaoo:
# Changes
- Added getJarVersion() to io.common.Util that reads the "Implementation-Version" attribute from the compiler jar's manifest. 
- Added the command `--jar-impl-version` to silver that allows the attribute to be set when building the compiler
- When retrieving silver's version, silver calls getJarVersion(), and if it does not return, defaults to `$TIME ` as the version.

I had to build twice, first without passing `--jar-impl-version` flag inside `make-compiler`, then pass it in again. Otherwise the build fails as `--silver-version` flag is unrecognized. Not sure if this complicates Jenkins process

### DIFF
--- a/grammars/silver/compiler/driver/BuildProcess.sv
+++ b/grammars/silver/compiler/driver/BuildProcess.sv
@@ -54,7 +54,7 @@ fun computeEnv IOErrorable<(Decorated CmdArgs, BuildEnv)> ::= args::[String] =
     -- line decide to go do, but currently it's hard to re-use code if we do that.
     if a.displayVersion then
       throwRunError(127, -- error code so 'ant' isnt run
-        "Silver Version " ++ getSilverVersion() ++ "\n" ++
+        "Silver Version " ++ getJarVersion() ++ "\n" ++
         "SILVER_HOME = " ++ benv.silverHome ++ "\n" ++
         "SILVER_GEN = " ++ benv.silverGen ++ "\n" ++
         "GRAMMAR_PATH:\n" ++ implode("\n", benv.grammarPath))
@@ -185,10 +185,10 @@ type IOErrorable<a> = EitherT<RunError IO a>;
 
 fun throwRunError IOErrorable<a> ::= c::Integer m::String = throwError(runError(code=c, errMsg=m));
 
-function getSilverVersion
+function getJarVersion
 String ::=
 {
   return error("NYI");
 } foreign {
-  "java" : return "common.Util.getSilverVersion()";
+  "java" : return "common.Util.getJarVersion(Init.class)";
 }
--- a/grammars/silver/compiler/driver/BuildProcess.sv
+++ b/grammars/silver/compiler/driver/BuildProcess.sv
@@ -54,7 +54,7 @@ fun computeEnv IOErrorable<(Decorated CmdArgs, BuildEnv)> ::= args::[String] =
     -- line decide to go do, but currently it's hard to re-use code if we do that.
     if a.displayVersion then
       throwRunError(127, -- error code so 'ant' isnt run
-        "Silver Version 0.4.5-dev\n" ++
+        "Silver Version " ++ getSilverVersion() ++ "\n" ++
         "SILVER_HOME = " ++ benv.silverHome ++ "\n" ++
         "SILVER_GEN = " ++ benv.silverGen ++ "\n" ++
         "GRAMMAR_PATH:\n" ++ implode("\n", benv.grammarPath))
@@ -184,3 +184,11 @@ data RunError = runError
 type IOErrorable<a> = EitherT<RunError IO a>;
 
 fun throwRunError IOErrorable<a> ::= c::Integer m::String = throwError(runError(code=c, errMsg=m));
+
+function getSilverVersion
+String ::=
+{
+  return error("NYI");
+} foreign {
+  "java" : return "common.Util.getSilverVersion()";
+}
--- a/grammars/silver/compiler/driver/BuildProcess.sv
+++ b/grammars/silver/compiler/driver/BuildProcess.sv
@@ -48,13 +48,14 @@ fun computeEnv IOErrorable<(Decorated CmdArgs, BuildEnv)> ::= args::[String] =
   | right(a) -> do {
     -- Figure out build env from environment and args
     benv::BuildEnv <- determineBuildEnv(a);
+    silverVersion::String <- lift(stateIO(getJarVersion));
     -- Because we want printing the version to work even if the environment is messed up
     -- we premptively handle that here. This is slightly unfortunate.
     -- Ideally, version printing would be just another thing we could have the command
     -- line decide to go do, but currently it's hard to re-use code if we do that.
     if a.displayVersion then
       throwRunError(127, -- error code so 'ant' isnt run
-        "Silver Version " ++ getJarVersion() ++ "\n" ++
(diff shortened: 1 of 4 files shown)


runtime/java/src/common/Util.java
```diff
@@ -504,6 +506,16 @@ public static StringCatter determineSilverHomePath(Class<?> clazz) {
...
+
+	public static StringCatter getSilverVersion() {
+		String path = determineSilverHomePath(Util.class).toString()+"/jars/silver.compiler.composed.Default.jar";
+        try {
+            JarInputStream file = new JarInputStream(new FileInputStream(path));
+            return new StringCatter(file.getManifest().getMainAttributes().getValue(Name.IMPLEMENTATION_VERSION));
+        } catch (Throwable t) {
+            throw new RuntimeException("Failed to get Silver version.", t);
+        }
+    }
```
remexre:
```suggestion
		try {
			JarInputStream file = new JarInputStream(new FileInputStream(path));
			return new StringCatter(file.getManifest().getMainAttributes().getValue(Name.IMPLEMENTATION_VERSION));
		} catch (Throwable t) {
			throw new RuntimeException("Failed to get Silver version.", t);
		}
	}

```

spaces->tabs, to match rest of file

grammars/silver/compiler/driver/BuildProcess.sv
```diff
@@ -184,3 +184,11 @@ data RunError = runError
 type IOErrorable<a> = EitherT<RunError IO a>;
 
 fun throwRunError IOErrorable<a> ::= c::Integer m::String = throwError(runError(code=c, errMsg=m));
+
+function getSilverVersion
+String ::=
```
remexre: Hm, how would we feel about making this actually be a more general `getJarVersion` that took in the path, then we did the `(_ ++ "/jars/silver.compiler.composed.Default.jar") <$> determineDefaultSilverHome()` or whatever on the Silver side?

That way, ableC or user projects could use it too.

grammars/silver/compiler/translation/java/driver/BuildProcess.sv
```diff
@@ -76,6 +84,9 @@ Either<String  Decorated CmdArgs> ::= args::[String]
            , flagSpec(name="--build-xml-location", paramString=nothing(),
                help="sets the path the Ant build.xml will be saved to",
                flagParser=option(buildXmlFlag))
+           , flagSpec(name="--silver-version", paramString=nothing(),
+               help="set the silver version string displayed to the user in --version",
```
remexre:
```suggestion
               help="set the Silver version string displayed to the user in --version",
```

remexre (approved):
lgtm pending getSilverVersion vs getJarVersion

remexre:
to do the jenkins dance,

- rsync or scp the JARs that you have working locally to foundry
- go to your branch's build, eg https://foundry.remexre.xyz/jenkins/job/melt-umn/job/silver/job/fix%252Fno-harcoding-silver-version/
- hit Build with Parameters on the left
- put the path to the directory your jars are in as `OVERRIDE_JARS` (ignore it mentioning coldpress, it means foundry, oops)

in theory you could avoid this by making one commit that adds the new flag but doesn't use it in the scripts, then another that adjusts the scripts; that way, each commit can be built by the previous one

grammars/silver/compiler/driver/BuildProcess.sv
allen-liaoo: That's a good idea. Should I make getJarVersion take IOToken and return IOVal?

allen-liaoo:
> to do the jenkins dance,
> 
> * rsync or scp the JARs that you have working locally to foundry
> * go to your branch's build, eg https://foundry.remexre.xyz/jenkins/job/melt-umn/job/silver/job/fix%252Fno-harcoding-silver-version/
> * hit Build with Parameters on the left
> * put the path to the directory your jars are in as `OVERRIDE_JARS` (ignore it mentioning coldpress, it means foundry, oops)
> 
> in theory you could avoid this by making one commit that adds the new flag but doesn't use it in the scripts, then another that adjusts the scripts; that way, each commit can be built by the previous one

I see, I'll try to split the commit into two and see if it builds successfully.

grammars/silver/compiler/driver/BuildProcess.sv
krame505: Yeah, that's probably the right way of doing it.

I'm not sure about the best approach to getting the jar path, though.  If someone builds an alternate composed Silver jar, then presumably we want to get the version that they set?  Is there a way to get the path to the JAR file containing the `main()` function that was initially called?

make-compiler
```diff
@@ -6,7 +6,7 @@
 set -euo pipefail
 
 export SILVER_HOME=$PWD
-JVM_ARGS=(-Xss20M -Xmx6500M -jar ../jars/silver.compiler.composed.Default.jar --no-stdlib --include-jar ../jars/CopperCompiler.jar --include-jar ../jars/commonmark-0.17.1.jar --relative-jar "$@")
+JVM_ARGS=(-Xss20M -Xmx6500M -jar ../jars/silver.compiler.composed.Default.jar --no-stdlib --include-jar ../jars/CopperCompiler.jar --include-jar ../jars/commonmark-0.17.1.jar --relative-jar "$@" --silver-version `git describe --tags`)
```
krame505: Maybe make the result of `git describe --tags` a separate var in the script?

grammars/silver/compiler/translation/java/driver/BuildProcess.sv
```diff
@@ -76,6 +84,9 @@ Either<String  Decorated CmdArgs> ::= args::[String]
            , flagSpec(name="--build-xml-location", paramString=nothing(),
                help="sets the path the Ant build.xml will be saved to",
                flagParser=option(buildXmlFlag))
+           , flagSpec(name="--silver-version", paramString=nothing(),
```
krame505: Echoing Nathan's comment, this should probably be `--jar-impl-version`, or something, since this isn't specific to building Silver.

krame505:
For future reference, the `upload-override-jars` script makes doing the rsync to foundry a bit easier.

krame505 (approved):

make-compiler
```diff
@@ -5,8 +5,9 @@
 
 set -euo pipefail
 
+silver_version=`git describe --tags`
```
krame505, line 8:
```suggestion
silver_version=`git describe --tags`
echo "Build version $silver_version"

```

Maybe print the computed version  here, would be helpful for diagnosing builds on Jenkins.

allen-liaoo:
Hmm is this ready to be merged?

krame505:
Yeah, I think so.

PR #882, opened 2025-01-17
Test/fetch unshallow

allen-liaoo:
Just opening this PR here so my PR (#881) can depend on it

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,6 +14,8 @@ melt.trynode('silver') {
 
     checkout scm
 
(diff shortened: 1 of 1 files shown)


krame505:
I don't really want to merge this until we know it actually fixes the problem on Jenkins for your PR.

FYI, if you merge this branch into fix/no-harcoding-silver-version, you can set the base branch of #881 to this branch. 
Then your branch will build with these changes, without having it show up again in your PR.

krame505 (approved):

PR #890, opened 2025-03-14
More refactoring to strategy attributes and flow analysis

krame505:
# Changes
The main change here is in how strategy attributes are translated.  Instead of translating `s1 <* s2` as
```
top.foo = decorate top.s1 with {env = top.env; ...}.s2;
```
with all the inherited attributes known to occur on `top`, we can now use a forward production attribute:
```
forward foo_fst = top.s1;
top.foo = foo_fst.s2;
```
This gets slightly more gross if `s1` isn't total, since we can't use `bind`, or stick the forward prod attribute declaration inside a `case`:
```
forward foo_fst = top.s1.fromJust;
top.foo = if top.s1.isJust then foo_fst.s2 else nothing();
```

## Other changes
* Add `when(s)` strategy utility from Stratego  - alias for `if s then id else fail`.
* MWDA refactor, check if a prod attribute is a forward prod attribute by checking the flow env instead of the regular one.  This is probably how it should have been done in the first place, but this change was needed because production attributes created by propagate don't make it into the production's defs properly.  
* Tweak to `fetch-jars` script to allow keeping the generated files - useful in development if one wishes to reset the jars to the previous working version, but a clean build is not needed.

# Documentation
This is mostly an internal refactor - updated comments as needed.

# Testing
No new test cases are needed, passes the existing strategy attribute tests.

### DIFF
--- a/fetch-jars
+++ b/fetch-jars
@@ -2,12 +2,13 @@
 
 set -eu
 
-# Usage: ./fetch-jars [rev-name] [--unstable] [--local] [--copper]
+# Usage: ./fetch-jars [rev-name] [--unstable] [--local] [--copper] [--keep-generated]
 # If a revision is not specified, the script fetches the latest jars from the current branch,
 # falling back to develop.
 # If --unstable is specified, always fetch from commit artifacts on foundry.
 # If --local is specified, only fetch jars from the local cache.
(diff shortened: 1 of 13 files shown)


unironically (approved):
Reviewed in MELT meeting!

PR #893, opened 2025-04-02
Avoid including extension expressions in interface files

krame505:
# Changes
Currently, deserializing an interface file can fail if it contains a strategy attribute declaration using extended syntax (e.g. `ableC_Expr` in a rule.)  This means that when a grammar that was built with an extended version of Silver, one may also need to use the extended Silver for anything depending on that grammar, even if the strategy wasn't used for anything, such as in an artifact grammar.  This makes build scripts for complex projects (e.g. for ableC) slightly more annoying.

Instead, just put a reflected version of the strategy in the interface file, to avoid needing to directly deserialize syntax that we don't know about.

# Documentation
N/A

# Testing
It builds ableC and extensions.


unironically (approved):
Reviewed in MELT meeting today.

PR #913, opened 2025-12-21
Refactor flow analysis to track deps on syn and eq subtree vertices

### LINKED ISSUE #907: Dependencies of siblings of shared trees are not considered in projection stitch points
Consider the following host language specification fragment, where the `count` attribute counts the number of variables in an expression.  This depends only on the `countIn` attribute.
```
synthesized attribute count::Integer;
inherited attribute countIn::Integer;

nonterminal Expr with env, errors, countIn, count;
flowtype Expr = errors {env}, count {countIn};

production var
top::Expr ::= n::String
{
  top.count = top.countIn + 1;
}

production addOp
{
  e1.countIn = top.countIn;
  e2.countIn = e1.count;
  top.count = e2.count;
}
```

On this extension production, a portion of the forward tree is determined by `top.env`, while `e` is shared under a separate child.
```
production foo
top::Expr ::= e::Expr
{
  forwards to addOp(if lookup("a", top.env).isJust then var("a") else intConst(0), @e);
}

```
The current flow analysis would see that for the `addOp` production, `e2.countIn` depends on `top.countIn`, so in `foo`, `e.countIn` would be projected to depend on just `top.countIn`.  However, there is in fact a dependency on `top.env` that would be missed by the current analysis; in fact, `foo` exceeds the host-language flow type for `count` by depending on `env`.  

The most straightforward fix for this would be to make all the inh vertices for any sharing site depend on the equation vertex of the top-level decoration site vertex (the forward vertex, in this case.)  However, this is quite conservative.  If a portion of the forward tree depends on `env`, for instance, then every projected inherited attribute would be seen to depend on `env`, even where there is otherwise an equation with no dependencies!  I'm not sure yet how significant of a limitation this will be in practice.  I also need to think about what the implications are in the translation pass paradigm, when defining new post-translation analyses.

A more general solution would be to change how projections are done, to project dependencies of all attributes for all production signature vertex types, and not just inherited attributes from RHS to LHS vertex types.  This would replace the nonterminal stitch points for forward and local vertex types; we would only need nonterminal stitch points for "holes" in the static tree that are not a production application or shared tree, e.g. part of the tree constructed by a higher-order attribute.  The main annoyance here is that we don't want the nonterminal stitch points for the children of projected productions any more, since we are filling in the actual dependencies.  Thus there would need to be a second version of every flow graph that does not have these stitch points, not used for flow type inference but only for projections.  We would also need to introduce equation vertices for the children in the projection versions of the productions, in order to project dependencies created by dynamic sub-trees, like in the above example.

The other upside of this approach is that we can easily check for cycles due to a missing inherited equation, if an inherited vertex for a decoration site ever depends on the corresponding shared tree's inherited vertex.

### LINKED ISSUE #909: Cycle in interface comparison due to flow info computation in action code
Comparing if interface files changed involves comparing the syntax ASTs, which contains the translation of action code, which is checked for errors prior to its translation being demanded.  When the MWDA is enabled, this can cause production flow graphs to be demanded, if the action code contains a pattern match.  This may require running flowtype inference while grammars are still being loaded, leading to a cycle.

One workaround might be to put the unparse of the action code in the syntax AST, and compare that instead of the translation.

krame505:
Fix #907. Fix #909.

# Changes
Massive flow analysis refactor.  This fixes a hole in the flow analysis for tree sharing, in which dependencies in constructed productions on their children and children's syn attributes were not being considered in the sharing production.  This yields more specific error messages than the "hidden transitive dependency" one in many cases, since we can now point to where a flow type is actually being exceeded, rather than "there may be a flow type exceeded somewhere".  This change also somewhat improves the precision of the analysis for specifications that do not use sharing, as we can now more precisely determine the dependencies of a synthesized attribute on a tree, when we know by what production that tree was constructed.

## Major changes
* Introduce _tile flow graphs_ for every production, containing additional eq vertices for the production lhs/rhs, but no nonterminal flowtype stitch points for the production RHS elements.
* Introduce "outer equation" vertices, that separately track the dependencies needed to construct the outermost production in a tree (required when accessing an attributes), vs regular equation vertices that track the dependencies of constructing all subtrees of a tree (required when taking a reference.)
* Introduce _tile stitch points_, that introduce edges corresponding to lhs and rhs vertices of a constructed production that has a decoration site vertex type.
* Eliminate nonterminal flowtype stitch points for local/production/translation attributes. Instead, introduce nonterminal stitch points for _holes_, i.e. any expression that has a decoration site vertex type that is not a production application or use of tree sharing.
* Refactor hidden transitive dependencies check.
* Add an additional check, needed for soundness, that override equations (syn or inh) on extension dispatch implementation productions do not depend on any new vertices (*not* lhs inhs!) not depended upon in the host language tile graph for the dispatch.
* Add conservative check for cycles due to missing inherited override equations on shared trees, disabled by default due to common patterns of forwarding violating this. Remove a previous more ad-hoc check for cycles specifically related to dispatching. 
* Some fixes to fully support arbitrary chains of translation attributes in flow types.
* Refactor anonymous vertices, introducing a new vertex type for anon scrutinees.  
* Pattern variables now use subtree vertices instead of anon vertex types, as is done in my thesis.  This means the special case checks for inhs being defined in remote equations are now covered by the existing decoration site tree resolution process.
* Checking of anon decoration site and anon scrutinee inh vertex deps is now also covered in decoration site resolution. 

## Performance optimizations
* Since all anon inh vertex deps are now checked in decoration site tree resolution, the `checkAllEqDeps` is no longer required and can be eliminated, substantially reducing the number of places we check for the presence of inherited equations (and query transitive dependencies.)  
  * This also means we are now finding flow errors in a few places where we should have been using `checkAllEqDeps`, like action block code.
* Reduce overhead of comparing flow vertices:
  * Redefine `Ord`/`Eq` on `FlowVertex` to compare a `vertexName` attribute, which is cached on the `FlowVertex`.  String comparison is much faster than traversing a (now potentially deeply) nested `VertexType` tree.  
  * Further reduce overhead of comparisons in graph operations by assigning a unique integer id to every globally unique `FlowVertex` (using a new `IdCache` library.)  This allows for vertex comparisons to be done using a single integer comparison, instead of the potentially-long `vertexName`.

Note that there is not a huge amount of change in the performance of the analysis; the above-mentioned optimizations more or less offset the overhead of the increase the in size (and number) of flow graphs.  

## Other changes
* Add foil to Jenkins downstream.  It is now passing the flow analysis, with the sharing cycle check enabled.  
* Fix lots of flow errors discovered in the Silver compiler (and downstream) found in the course of this refactor.
* Fix cycle crash due to comparing translation of action code in comparing interface files, by instead comparing the unparse to determine dirtiness.
* Remove obsolete utilities
* Add a few new things in the standard libraries
* Error message improvements
* Lots of other misc cleanup

# Documentation
Updated comments, mostly.  There actually aren't very many user-visible changes here, mostly just better (and more correct) error messages.

I'll try to update the website at some point to discuss the `--warn-sharing-cycles` flag.

# Testing
Tested by finding new flow errors in the Silver compiler and downstream projects.  Also added and updated a number of new tests in the testsuite.


grammars/silver/compiler/definition/core/OccursDcl.sv
```diff
@@ -178,6 +178,10 @@ top::AGDcl ::= at::QName attl::BracketedOptTypeExprs nt::QName nttl::BracketedOp
         " but type variable(s) have kind(s) " ++ implode(", ", map(compose(prettyKind, (.kindrep)), nttl.types)) ++ ".")]
     | _ -> []
     end;
+  
+  -- Seeding flow deps
+  top.errors <- if false then [] else error(hackUnparse(
```
unironically: Why do we have `if false then [] else ...` here instead of the expression in the `else` branch?

krame505: Good catch!  Hmm, this must not actually be getting exercised.

grammars/silver/compiler/extension/autoattr/Propagate.sv
```diff
@@ -148,6 +148,9 @@ dispatch Propagate = ProductionStmt ::= includeShared::Boolean @attr::QName;
 abstract production propagateError implements Propagate
 top::ProductionStmt ::= includeShared::Boolean @attr::QName
 {
+  -- Seeding flow deps
+  top.errors <- if false then [] else error(hackUnparse((attr.lookupAttribute, includeShared)));
```
unironically: Same comment here about `if fale then [] else ...`

grammars/silver/core/Pair.sv
```diff
@@ -65,4 +65,9 @@ Pair<[a] [b]> ::= lst::[Pair<a b>]
   else (head(lst).fst :: recurse.fst, head(lst).snd :: recurse.snd);
 }
 
+@{--
+ - Compute the Cartesian product of two lists.
+ -}
+global cartProd :: ([Pair<a b>] ::= [a] [b]) = lift2(pair(fst=_, snd=_), _, _);
```
unironically, line 71: Just curious why this is a `global` instead of a `fun` - does it matter?

krame505: Mostly just a style thing, I tend to prefer partial application vs defining a function that immediately calls another function with the same parameters.

unironically (approved):

PR #915, opened 2026-02-09
Align Hello.sv with the documentation.

niko-vcc:
Documentation at https://melt.cs.umn.edu/silver/install-guide/#testing-things-out-by-building-the-tutorials claims that the World is upper case (as is expected in most hello-world applications).

# Changes

1. Minor update on a string to align the output of the Hello.sv with the online documentation
2. Simplified the install script. Removed redundant calls to `readlink`, and the code that was required to maintain different executions per platform. Now the code only depends on a `bash`-ism.

# Documentation

- Align website documentation and source code, by updating the source code to align with existing documentation (and common casing in Hello World)

# Testing

Locally, this change affects my silver docker build where I validate that the image is correctly built and things are as expected by utilizing the Hello.sv and expecting the output. 
Same docker build process, also is using the install script.

# Assumptions

That the install script and the scripts to be installed live in the same directory (same as before). The install script is called directly (not symlinked - different than before).

# Quality

There are two `shellcheck` warnings that exist in code I didn't touch, newly added/modified code  is shellcheck warning free

```
$ shellcheck install-silver-bin 

In install-silver-bin line 6:
  echo "~/bin not found. Aborting."
        ^------------------------^ SC2088 (warning): Tilde does not expand in quotes. Use $HOME.


In install-silver-bin line 7:
  case `uname` in
       ^-----^ SC2006 (style): Use $(...) notation instead of legacy backticks `...`.

Did you mean:
  case $(uname) in

For more information:
  https://www.shellcheck.net/wiki/SC2088 -- Tilde does not expand in quotes. ...
  https://www.shellcheck.net/wiki/SC2006 -- Use $(...) notation instead of le...
```

### DIFF
--- a/support/bin/install-silver-bin
+++ b/support/bin/install-silver-bin
@@ -28,51 +28,29 @@ fi
 
 echo "Found ~/bin"
 
-case `uname` in
-*Darwin*)
-  READLINK=greadlink
-  if [ ! -f `which greadlink` ]; then
-    echo "Missing greadlink. Please install coreutils:"
-    echo -e "\tbrew install coreutils"
-    exit 4
-  fi
-  ;;
-*)
-  READLINK=readlink
-  ;;
-esac
+# get REPO location from git
+REPO=$(git rev-parse --show-toplevel)
+echo "Found $REPO"
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
 for SCRIPT_NAME in silver silver-custom
 do
+    SRC_SCRIPT="${SCRIPT_DIR}/${SCRIPT_NAME}"
+    DST_SCRIPT="${HOME}/bin/${SCRIPT_NAME}"
 
-    if [ -f $SCRIPT_NAME ]; then
-        REPO=$("$READLINK" -f ../..)
-        SCRIPT=$SCRIPT_NAME
-    elif [ -f support/bin/$SCRIPT_NAME ]; then
-        REPO=$("$READLINK" -f .)
-        SCRIPT=support/bin/$SCRIPT_NAME
+    if [ -f "${SRC_SCRIPT}" ]; then
+       rm -f "${DST_SCRIPT}"
+       ln -s "${SRC_SCRIPT}" "${DST_SCRIPT}"
     else
-        echo "Couldn't find the Silver jars!"
-        echo "(if this is a fresh checkout, run fetch-jars before this script.)"
-        exit 2
-    fi
-    
-    echo "Found $REPO"
-
-    if [ -e ~/bin/$SCRIPT_NAME -o -h ~/bin/$SCRIPT_NAME ]; then
-        # -h needed in case the link exists but the targeted file does not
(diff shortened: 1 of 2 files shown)


support/bin/install-silver-bin
```diff
@@ -28,51 +28,29 @@ fi
...
-    echo -e "\tbrew install coreutils"
-    exit 4
-  fi
-  ;;
-*)
-  READLINK=readlink
-  ;;
-esac
+# get REPO location from git
+REPO=$(git rev-parse --show-toplevel)
```
krame505: We also distribute Silver as a tarball with `.git` and other items stripped out, so this script should still work if this isn't a git checkout.

krame505 (changes requested):

support/bin/install-silver-bin
```diff
@@ -28,51 +28,29 @@ fi
...
+echo "Found $REPO"
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
krame505, line 31: Does this work if there is a softlink in the path?  I think that is why we were using readlink before.

niko-vcc: When invoked from its symlink (`ln -s`) this would point to the directory containing the symlink. 

What is the use-case that this script is supposed to support? Is the use-case you have a symlink pointing to the install-script, with the additional assumption that the symlink points to the real/correct directory with the right wrappers (silver, ...) living in the same directory?

I could use `realpath` (or revert it to the previous version with the `readlink`) that should be always available which should be available on both Linux/MacOS/BSD that have an installed gnu coreutils package. 

If the old functionality is desirable as-is, I could have an additional install script w/o the `readlink` stuff, and an explicitly stated assumption that it is not to be invoked as a symlink ?  I could also have a check on it to prohibit running if the invocation is from a symlinked location.

support/bin/install-silver-bin
niko-vcc: removed.

support/bin/install-silver-bin
krame505: Honestly, I'm not entirely sure why `readlink` was required.  I think it may have just been used as a platform-independent way of determining an absolute path?  I think that this is probably fine, we can change it back if there are issues.

krame505 (approved):

support/bin/install-silver-bin
krame505: Looking at this history, this used to be used to set an environment variable in the installed script, but that is no longer needed, so removing this is fine.

krame505:
Looks like our CI isn't properly set up to handle pull requests from forks.  Jenkins reports an additional kind of status for pull requests (`continuous-integration/jenkins/pr-merge` instead of `continuous-integration/jenkins/branch`), but we have logic related to matching branch names for fetching jars and downstream builds that only works with `continuous-integration/jenkins/branch`.  There doesn't seem to be a way to tell github to accept either one, so for now we just need to admin override the branch protection rules to merge PRs from forks like this one.

PR #916, opened 2026-02-16
Fix/collection base aspect default

### LINKED ISSUE #290: aspect default productions with collection attributes

unironically:
# Changes
Modified runtime code for collection attributes so that base definitions can be taken from default productions. Essentially the same code as is used to look for synthesized attribute definitions in default productions, seen here:

https://github.com/melt-umn/silver/blob/1184b903f15a865044cabad58cbb17688aebe7d2/runtime/java/src/common/DecoratedNode.java#L521-L532

# Documentation
No new documentation added - small change to runtime.

# Testing
Modified existing collection attribute tests to include cases where base definitions are taken from default productions.

### DIFF
--- a/runtime/java/src/common/CollectionAttribute.java
+++ b/runtime/java/src/common/CollectionAttribute.java
@@ -68,7 +68,7 @@ public abstract class CollectionAttribute implements Lazy {
 				} catch(Throwable t) {
 					throw new TraceException("Error evaluating base of collection attribute " + un.getNameOfSynAttr(index) + " via forward of " + un.getName(),t);
 				}
-			}	else {
+			} else {
 				Lazy l = un.getDefaultSynthesized(index);
 				if(l != null) {
 					try {
--- a/runtime/java/src/common/CollectionAttribute.java
+++ b/runtime/java/src/common/CollectionAttribute.java
@@ -69,15 +69,15 @@ public abstract class CollectionAttribute implements Lazy {
 					throw new TraceException("Error evaluating base of collection attribute " + un.getNameOfSynAttr(index) + " via forward of " + un.getName(),t);
 				}
 			}	else {
-        Lazy l = un.getDefaultSynthesized(index);
-        if(l != null) {
-          try {
-            return l.eval(context);
-          } catch(Throwable t) {
-            throw new TraceException("While evaling default for collection attribute '" + un.getNameOfSynAttr(index) + "' in " + context.getDebugID(), t);
-          }
-        }
-		  }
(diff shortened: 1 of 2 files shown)


runtime/java/src/common/CollectionAttribute.java
```diff
@@ -68,7 +68,16 @@ public Object eval(final DecoratedNode context) {
 				} catch(Throwable t) {
 					throw new TraceException("Error evaluating base of collection attribute " + un.getNameOfSynAttr(index) + " via forward of " + un.getName(),t);
 				}
-			}	
+			}	else {
```
krame505: Nit, spacing is weird here.

krame505:
I don't quite understand what is being fixed here - is this an instance of https://github.com/melt-umn/silver/issues/290 re-surfacing?  There was a (rather clunky) fix for that made by @remexre in https://github.com/melt-umn/silver/pull/510, is that check still needed with this change?

runtime/java/src/common/CollectionAttribute.java
unironically: Oopsie - will fix this.

unironically:
This seems to be the same as #290. For instance, running the jar produced by compiling this simple grammar will give `No base defined for collection attribute hello:things in hello:foo`. After fix, program exits with 0 normally.

```
grammar hello;

fun main IO<Integer> ::= largs::[String] =
  do {
    print("Hello, World!\n");
    return if foo().ok then 0 else 1;
  };

synthesized attribute ok::Boolean;
synthesized attribute things::[String] with ++;

nonterminal Foo with ok, things;

production foo
top::Foo ::=
{ top.ok = length(top.things) == 1;
  top.things <- ["thing"]; }

aspect default production top::Foo ::=
{ top.things := []; }
```

Not sure about #510, but I can experiment with removing that check before this is merged. In any case it seems from my testing that #290 was not fixed before this PR.

krame505:
I took another look and this actually seems to be a different issue than #290 - that was collection contribution equations in default prods being ignored when there is a base equation, but this PR deals with the opposite issue: a contribution in a production where the default is defined in the base.  So the fix from #510 should stay.

krame505 (approved):

